PR #7, opened 2020-05-05
changed GetCurrentDirectoryA to GetCurrentDirectory

Sebanisu:
sithlord48 was testing on linux under wine and for some reason GetCurrentDirectoryA didn't work but we changed to GetCurrentDirectory and it is working now.

### DIFF
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -2020,7 +2020,7 @@ BOOL WINAPI DllMain(HINSTANCE hinstDLL, DWORD fdwReason, LPVOID lpvReserved)
 		__debugbreak();
 		*/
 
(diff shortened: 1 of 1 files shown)


sithlord48:
After the change I am now able to load the external music and my mods from my mod patch

julianxhokaxhiu:
Oh I see! I think this PR is not required at all. You may probably just need an [hard link](https://en.wikipedia.org/wiki/Hard_link) and you should be good to go :)

Sebanisu:
Standard library to maybe get full Unicode and text transcoding in cpp23. There was a talk at cppcon 2019 about it

julianxhokaxhiu:
That's too far away but I think it makes sense. Thanks for letting me know!

PR #10, opened 2020-05-06
FF8 External Music

myst6re:
Implements external music plugin for FF8.

### DIFF
--- a/src/ff8_data.h
+++ b/src/ff8_data.h
@@ -210,6 +210,9 @@ void ff8_find_externals()
 	ff8_externals.sub_469640 = get_relative_call(ff8_externals.pubintro_init, 0xD8);
 	ff8_externals.sub_46DBF0 = get_relative_call(ff8_externals.sub_469640, 0x5D);
 	common_externals.directsound = (IDirectSound**)get_absolute_value(ff8_externals.sub_46DBF0, 0x26);
+	common_externals.play_midi = 0x46C290;
+	common_externals.stop_midi = 0x46C6A0;
+	common_externals.get_midi_name = (char* (*)(uint))0x46C840;
 
 	ff8_externals.sub_5304B0 = (void (*)())get_relative_call(common_externals.update_movie_sample, 0x3D9);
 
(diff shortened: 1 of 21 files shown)


src/ff8_data.h
```diff
@@ -210,6 +210,14 @@ void ff8_find_externals()
 	ff8_externals.sub_469640 = get_relative_call(ff8_externals.pubintro_init, 0xD8);
 	ff8_externals.sub_46DBF0 = get_relative_call(ff8_externals.sub_469640, 0x5D);
 	common_externals.directsound = (IDirectSound**)get_absolute_value(ff8_externals.sub_46DBF0, 0x26);
+	common_externals.play_midi = 0x46C290;
+	common_externals.stop_midi = 0x46C6A0;
+	common_externals.midi_status = 0x46B960;
+	//common_externals.pause_midi = 0x46F2F0;
+	//common_externals.restart_midi = 0x46C670;
```
julianxhokaxhiu, line 217: Are these values safe enough for all the languages? Can we use a relative logic to get them otherwise?

myst6re: Safe for french only! And yes :)

julianxhokaxhiu: I would appreciate if we can get a relative log. Let me know if I can help you there in case! Thanks :D

src/ff8_data.h
```diff
@@ -105,11 +105,7 @@ void ff8_find_externals()
 	common_externals.get_movie_frame = get_relative_call(ff8_externals.sub_529FF0, 0x26);
 
 	ff8_externals.movie_object = (ff8_movie_obj *)get_absolute_value(common_externals.prepare_movie, 0xDB);
-
-	ff8_externals.mode1_main_loop = get_absolute_value(ff8_externals.main_loop, 0x144);
```
myst6re, line 109: I remove this, because this is already done before, dunno why it is repeated

src/ff8_data.h
```diff
@@ -172,8 +168,7 @@ void ff8_find_externals()
 	}
 
 	ff8_externals.menu_viewport = (sprite_viewport *)(get_absolute_value(ff8_externals.sub_4972A0, 0x12) - 0x20);
-
-	ff8_externals.sub_4A24B0 = get_absolute_value(ff8_externals.sub_470520, 0x2B);
```
myst6re, line 176: Same here, we already set it

src/ff8_data.h
myst6re: This was actually really hard to do, but here we go :)

julianxhokaxhiu:
As always fantastic work over here. You've just revolutionized not only the music layer but also introduced a cool new feature for FF8!

PR #15, opened 2020-05-10
Improving resuming time

myst6re:
That's a naive implementation for now: I duplicate the vgmstream instance + the state and restore it later.

Maybe I should look how vgmstream can handle multiple buffers.
Winamp plugins haven't this kind of logic, but I written my code like a possible extension of the Winamp's in plugin. Currently external Winamp plugins will behave like before, ie with GetOutputTime + SetOutputTime


julianxhokaxhiu:
Amazing as always :) thank you!

PR #19, opened 2020-05-13
Set plugin if music path is set

sithlord48:
If the user has set a path check if we can open that path and we if can ensure our plugin is set correctly .. This happens before our other checks so our defaults will now acts as a fallback as an added bonus.

### DIFF
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -2122,6 +2122,19 @@ BOOL WINAPI DllMain(HINSTANCE hinstDLL, DWORD fdwReason, LPVOID lpvReserved)
 
 		_strlwr(parentName);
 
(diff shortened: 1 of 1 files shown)


src/common.cpp
```diff
@@ -2172,13 +2185,13 @@ BOOL WINAPI DllMain(HINSTANCE hinstDLL, DWORD fdwReason, LPVOID lpvReserved)
 				if (external_music_path == nullptr) external_music_path = "data/music_ogg";
 
 			}
+
```
julianxhokaxhiu, line 2188: This space should not be there, as there's an `else if` following :D

julianxhokaxhiu:
Thank you :)

PR #21, opened 2020-05-13
Remove FF7Music Support: Superseed by internal winamp plugin support

sithlord48:


julianxhokaxhiu:
@myst6re can I have your opinion on this too please? I would be happy to deprecate but only if the Winamp layer is fully capable of replacing it. Let me know your thoughts :) Thank you!

myst6re:
In addition of being fully capable of replacing it, it is much better

julianxhokaxhiu:
Allright, I trust you both. Let's merge this :) Thank you!

PR #25, opened 2020-05-14
Feature/wav in winamp plugin + music lock patch

myst6re:
Adding ability to replace the wav files (FF7 only)

Fix a music bug in the game: the "lock music" instruction is cleared on field changes, but "lock music" should be persistent across fields. This fixes:
- music stop bug in the first scene with Reno and Aerith
- Battle music restarted after the fight against Rufus
- Underneath the rotting pizza intervention after the movie with Sephiroth in fire
- more scenes I haven't tested involving MUSCLCK opcode (in space for example)

Possible consequence: if fields scripts does not unlock music correctly, a music could be played for too long in wrong scenes.

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -182,6 +182,7 @@ struct common_externals
 	uint get_time;
 	uint midi_init;
 	char *(*get_midi_name)(uint);
+	uint use_midi;
 	uint play_midi;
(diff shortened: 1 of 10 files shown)


julianxhokaxhiu:
Thank you!

PR #47, opened 2020-05-23
Replace open_flevel_siz method with a more flexible implementation

myst6re:
flevel.lgp/flevel.siz contains the uncompressed sizes of the LZSed field map files.

The current implementation uses this file to allocate enough memory to uncompress the field. It adds 2000000 to this value, just to be safe.

But not matters how long is flevel.siz, the parser always read 387 entries (aka the field map count). The game initialize 1200 (see sub_60EEB2) entries of the field info structure (where the uncompressed size is stored), so we should be safe to increase this limitation.

### DIFF
--- a/src/ff7/field.cpp
+++ b/src/ff7/field.cpp
@@ -20,6 +20,7 @@
 
 #include "../ff7.h"
 #include "../log.h"
+#include "defs.h"
 
 /*
  * This file contains the changes necessary to support subtractive and 25%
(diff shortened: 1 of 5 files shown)


src/ff7/field.cpp
```diff
@@ -129,3 +130,40 @@ void field_layer2_pick_tiles(short x_offset, short y_offset)
...
+	uint* uncompressed_sizes = reinterpret_cast<uint*>(buffer);
+	uint count = size / sizeof(uint);
+	uint* flevel_sizes = reinterpret_cast<uint*>(ff7_externals.field_map_infos + 0xBC);
+
+	if (count > max_map_count) {
+		count = max_map_count;
+	}
+
+	for (uint i = 0; i < count; ++i) {
+		flevel_sizes[i * 0x34] = uncompressed_sizes[i] + 4000000; // +2 MB compared to the original implementation
```
myst6re, line 158: Uncompressed size is never > 1MB in real life. I feel like I could remove the dependency to flevel.siz completely and force a fixed buffer size.

myst6re: Okay no I keep it like this, if one day someone make a huge field map file, he can increase the limitation by modifying flevel.siz.

PR #48, opened 2020-05-24
Fixing crash when using external music with FF8

myst6re:


julianxhokaxhiu:
Good catch! Thank you :)

PR #63, opened 2020-06-07
SFX fixes

myst6re:
This PR mainly fixes game SFX bugs.

Changelog:
 - When you modify the SFX volume in the game (config menu), now it is updated dynamically without restart
 - Ambient sounds are resumed after a battle (for example the alarm in the first reactor)
 - Escape sound is now played everytime, not only the first time
 - Fixing music volume fade timing (my fault)
 - Apply theses fixes and other SFX fixes even if external_music is not enabled

### DIFF
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -35,6 +35,7 @@
 #include "gl.h"
 #include "movies.h"
 #include "music.h"
+#include "sfx.h"
 #include "saveload.h"
 #include "gamepad.h"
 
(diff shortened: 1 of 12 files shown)


myst6re:
Fixed! I also noticed that channel 5 is not restored on the PSX version.

src/common.cpp
```diff
@@ -2065,7 +2066,8 @@ __declspec(dllexport) void *new_dll_graphics_driver(void *game_object)
 	info("Original resolution %ix%i, New resolution %ix%i\n", game_width, game_height, window_size_x, window_size_y);
 
 	// perform any additional initialization that requires the rendering environment to be set up
-	if (use_external_music) music_init();
+	music_init();
```
julianxhokaxhiu, line 2069: Is there a reason why the flag is now gone?

myst6re: There is a music bug patch in music_init, this bug is not related to external music plugin. If you want to reorganize this, it is possible :)

julianxhokaxhiu: Looks good to me in general, but I would be happy to ensure users can still use the original music engine when they do disable the flag.

Do you think is it feasible?

myst6re: This is already the case, I wrote the test in music.cpp. My removal of the if (use_external_music) is something I forgot to do in a previous PR of mine

julianxhokaxhiu:
Thank you! Really appreciated :)

PR #68, opened 2020-06-12
Fix inconsistent sfx volume

myst6re:
There is two volume levels: the one (per channel) from the original game (from 0 to 127) and a master level (only in the PC version) which can be modified via the configuration menu.

The final volume is computed like this: master * volume[channel] / 100.

The set_sfx_volume (on channel) method has a bad behavior: it set the sfx state with the result of this volume, but everytime the state is used, the computation above is computed... again!

Instead of removing computations when the state is used, I chose to force the value of the state in set_sfx_volume directly.

### DIFF
--- a/src/sfx.cpp
+++ b/src/sfx.cpp
@@ -27,6 +27,7 @@
 
 uint sfx_volumes[5];
 ff7_field_sfx_state sfx_buffers[4];
+uint real_volume;
 
 void sfx_init()
 {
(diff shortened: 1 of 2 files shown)


src/sfx.cpp
```diff
@@ -116,7 +131,6 @@ void sfx_remember_volumes()
 
 	for (int i = 0; i < 5; ++i) {
 		sfx_volumes[i] = sfx_state[i].buffer1 != nullptr ? sfx_state[i].volume1 : sfx_state[i].volume2;
-		sfx_volumes[i] = floor(sfx_volumes[i] * 100 / float(*common_externals.master_sfx_volume) + 0.9f);
```
myst6re, line 119: No more approximation, yay!

julianxhokaxhiu:
Thank you!

PR #73, opened 2020-06-24
Added, ImGui, input listeners, WindowProc proxy and keyboard blocking

Extapathy:
- Merged some keyboard externals into common
- Removed a DInput patch as that function is now replaced
- Added event listener functionality
- Added WindowProc proxy
- Added ability to block game from receiving keyboard input
- Added ImGui for debug windows

### DIFF
--- a/misc/FFNX.overlay.frag
+++ b/misc/FFNX.overlay.frag
@@ -1,0 +1,11 @@
+$input v_color0, v_texcoord0
+
+#include <bgfx/bgfx_shader.sh>
+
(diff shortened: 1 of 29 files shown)


Extapathy:
Yeah, the input state override isn't used at the moment because it was intended to be used by imgui to block the input.
It's fine by me if you want to wait for the full commit, I've still got a few rendering issues with imgui to fix and a bunch of clutter to trim.

julianxhokaxhiu:
Thank you!

Issue #35, opened 2020-05-15
[ FF7 ] SFX sound bug


satsukiyatoshi:
In the game SFX sounds are randomly played
The SFX sound volume is not applyied too

myst6re:
Is it related to the "escape sound only played once" bug?
The original PC game always had this bug?

satsukiyatoshi:
Not this bug but sfx sometime play, sometime not, seems realy random.

myst6re:
I heard something like this once, it was like two sounds cannot be played at the same time.

Rumbah2:
For me it's in the original game, too and dependent on the SFX volume I set in game.
For different setting different sounds disappear. With set to 100 I have all sounds.

Reproducible for me:
Set SFX to 30
completely exit the menu
enter the menu
menu cursor sounds are gone

when setting the sfx back to 100 you'll notice how the cursor sound slowly comes back but stays quiet
completely exit the menu again
enter the menu
sounds back to normal expected volume at 100

myst6re:
Dunno if it is related, but on the original game (without music plugin) about volume changes when there is a music fade out, the volume is globally modified, and can be muted completely. For example the prelude in credits, after the fade out you cannot hear anymore the sound of the start / load menu, because the application volume is globally muted. Since we moved music to directsound with the external_music plugin, this problem is fixed.

myst6re:
> 
> 
> For me it's in the original game, too, and dependent on the SFX volume I set in game.
> For different setting different sounds disappear. With set to 100 I have all sounds.
> 
> Reproducible for me:
> Set SFX to 30
> completely exit the menu
> enter the menu
> menu cursor sounds are gone
> 
> when setting the sfx back to 100 you'll notice how the cursor sound slowly comes back but stays quiet
> completely exit the menu again
> enter the menu
> sounds back to normal expected volume at 100

This bug is not related to the SFX bug in this issue. The game doesn't change the SFX volume dynamically when you set it in menus. But it does save it in the register when you quit the main menu. If you close and restart the game after your change, you will hear the modification.

I can fix it, in fact I already made a patch. We'll see if the PR is relevant or not :)

> Is it related to the "escape sound only played once" bug?

This bug is also not related to the current issue. Bug I found how to fix it aswell. The sound is started in a loop at the end of the battle, so a flag is set to prevent the sound to be played several times at once. The problem is that this flag is **never** reset.

myst6re:
[...]

PR #76, opened 2020-06-30
Fix lots of SFX bugs (FF7)

myst6re:
- Play music in Game Over Screen when the timer reached zero in the first mission
- Fix summon sfx volume (user SFX volume wasn't applied)
- Fix Leviathan sound stopped early
- Fix Omnislash sound not loaded


julianxhokaxhiu:
As Always, thank you!

myst6re:
#35

PR #85, opened 2020-08-17
speedhack stays at max when trying to increment further, same for decrementing

berserkingyadis:
before:
incrementing speedhack speed goes very fast and ticks through multiplier very quickly. min->max->min->max

after
it stays at max when holding the "up the speed pls" button

### DIFF
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -398,16 +398,16 @@ LRESULT CALLBACK WindowProc(HWND hwnd, UINT uMsg, WPARAM wParam, LPARAM lParam)
 // SPEEDHACK
 void speedhack_incr()
 {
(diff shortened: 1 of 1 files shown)


berserkingyadis:
if https://github.com/julianxhokaxhiu/FFNx/pull/86 gets accepted, this PR here is not that necessary anymore

julianxhokaxhiu:
The idea of being "rolling" whichever direction you go it's totally fine and I don't feel the need to remove that feature.

The fact that your symphtom is on the gamepad doesn't mean a feature has to be completely removed. As said in the other PR, this will be fixed with a proper patch.

Because of this, I'll close this PR too.

Cheers!

PR #86, opened 2020-08-17
speedhack changes only once per actual gamepad button press

berserkingyadis:
Problem for me was that when I pressed the "speedhack speed up please" combo on gamepad for a bit, the speed already went from 1x to like 5x. If I pressed it abit longer it went from 1x->8x->1x again. 

It triggered too often IMO:

With this PR it only changes once per actual button press. That means if you press the gamepad buttons and hold them it will only add/decrement the increment value from the cfg once.

I only needed to change this for the gamepad input management. On keyboard it works fine without change needed. You press keyboard once and hold for a bit but it doesnt increment further unless you hold buttons for more than a sec or so.

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -282,6 +282,7 @@ void get_userdata_path(PCHAR buffer, size_t bufSize, bool isSavegameFile);
 void speedhack_incr();
 void speedhack_decr();
 void speedhack_reset();
+bool speedhack_pressed;
 
 
 #if defined(__cplusplus)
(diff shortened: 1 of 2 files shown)


berserkingyadis:
with this, the default "speedhack_step" cfg value should maybe a bit higher so people do not have to press the gamepad combo so often. maybe change to 1 and not 0.5

julianxhokaxhiu:
Hi there :) Thank you very much for this PR. Although I do get the issue, I don't think this is the right solution. I've got a better idea which fits better with the engine flow and I'll implement this ASAP. Meanwhile I'll close this if you don't mind!

Cheers!

berserkingyadis:
Thats ok, quickest solutions arent always the best. That applies even more when the dev doing them hasnt gotten the full picture of the engine/project yet ;)

PR #96, opened 2020-09-12
Winamp layer fixes

myst6re:


src/audio.cpp
```diff
@@ -215,6 +215,11 @@ void NxAudioEngine::setMusicSpeed(float speed)
 	_engine.setRelativePlaySpeed(_musicHandle, speed);
 }
 
+void NxAudioEngine::setMusicLooping(bool looping)
+{
+	_engine.setLooping(_musicHandle, false);
```
julianxhokaxhiu, line 220: Was this supposed to be `_engine.setLooping(_musicHandle, looping);`?

src/audio.h
```diff
@@ -73,6 +73,7 @@ class NxAudioEngine
 	float getMusicVolume();
 	void setMusicVolume(float _volume, size_t time = 0);
 	void setMusicSpeed(float speed);
+	void setMusicLooping(bool looping);
```
julianxhokaxhiu, line 76: Maybe we can set this parameter to `false` by default, in order to match your previous line :)

julianxhokaxhiu (changes requested):

src/audio/winamp/winamp.cpp
```diff
@@ -73,14 +73,32 @@ namespace SoLoud
...
+	{
+		// Can resume music
+		if (aInfoKey == 7777) {
+			// Needs to Duplicate
+			return 1.0f;
+		}
+
+		return 0.0f;
+	}
+
```
julianxhokaxhiu, line 101: This section needs to be added to VGMStream too I guess, as the whole flow of "can I resume" is called in `music.cpp` therefore, also VGMStream is impacted by this change. Would you be able to add the required logic over there too?

julianxhokaxhiu (changes requested):

src/audio/winamp/winamp.cpp
myst6re: Yes I can do that
Just know that vgmstream is currently not impacted. getInfo returns 0.0 in AudioSourceInstance parent implementation.

julianxhokaxhiu: Ah ok so if there's no impact then it's not needed :) Thanks! Fine by you if I merge this then?

myst6re: No don't merge it, I have more changes to push :)

myst6re:
TODO: edit features in README, change ps2_bios_path description in configuration file.

PR #111, opened 2020-10-08
Feature/field music resuming

myst6re:
Music behavior in FF7:
 - [FIELD] A music can be played in channel 0 or channel 1 (not both, except during the transition). Only channel 0 is resumable, and it must be when channel 1 is played after channel 0.
 - [WM] The main theme and the other WM them are always resumable
 - [OTHER MODES] No resuming

Stopping a music (or set the volume to 0 => stop) will never save the last playing music. So resuming is not possible after a stop.

Music behavior in FF8:
TODO


julianxhokaxhiu (approved):

julianxhokaxhiu:
As always, thank you!

PR #140, opened 2020-12-13
Fix/volume transition time

myst6re:
- Improving the precision of fade time
- stop the midi when cross fade with midi = 0 is called
- fix cross fade after battle (for the battle/victory theme fade out)
- fix cross fade time to the ~same value from the original game (PlayStation, not PC!)
- fix music volume transition time to the same value from the original game (PlayStation, not PC!)

Checked in the PlayStation version:
 - step = 64 => 1 second, step = 128 => 2 seconds...
 - cross music is around 1 to 2 seconds, never more


myst6re:
I also disabled music speed changes for FF7 (with external music), as we don't handle it well.
And I fixed an issue with music resuming (FF7 too).

julianxhokaxhiu (approved):

PR #153, opened 2021-01-05
FF8: Pause/resume music and SFX when pausing the game in field

myst6re:
Tiny fix to make this feature also available for native midis


julianxhokaxhiu:
Thanks!

Issue #202, opened 2021-02-21
[ FF7 ] No music in the world map after a game over.


daedrixxx2:
After a Game Over, if "external_music_resume" is set to "true", when reloading the game there'll be no music in the world map (didn't try if it happens in fields too). After a battle, or entering a town, music restarts normally.
This bug doesn't happen if "external_music_resume" is set to "false".

julianxhokaxhiu:
@myst6re Would you be able to look at this one? Thank you!

myst6re:
Yes assign this issue to me please

julianxhokaxhiu:
@daedrixxx2 can you please try the latest canary and report back? Thank you

daedrixxx2:
The bug has been resolved :)

And I'm pleased to learn that the issue with the music resume feature on known areas has been resolved as well :D

Good work everyone. Thank you 👍

julianxhokaxhiu:
Thanks for testing, glad everything was sorted out :) Enjoy!

PR #206, opened 2021-02-25
FF7: resume any music after battle in WM

myst6re:


src/music.cpp
```diff
@@ -531,17 +472,34 @@ uint32_t ff7_music_sound_operation_fix(uint32_t type, uint32_t param1, uint32_t
...
 	if (*unk1)
 	{
-		ff7_next_battle_world_id = *unk3;
+		ff7_next_battle_world_id = *battle_id;
 
 		// Now we know the battle scene ID, so we can try to customize battle music in Worldmap too
-		stop_music();
-		ff7_play_midi(7);
+		// Play music
+		((uint32_t(*)(uint32_t, uint32_t, uint32_t, uint32_t, uint32_t, uint32_t))ff7_externals.sound_operation)(
```
myst6re, line 500: Oops, removing stop_music() is a bad idea, at least the current music id must be discarded

julianxhokaxhiu: Yes, that's why I used it, in order to flush the current id actually.

myst6re: Done

myst6re:
Should fix #202

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #227, opened 2021-04-11
Implemented real-time lighting and shadow option

CosmosXIII:
- Implemented vertex normal calculation by averaging adjacent triangle normals
- Implemented physically-based lighting with diffuse and Cook-Torrance specular BRDF
- Implemented real-time shadows through shadow maps
- Implemented OptimizedPCF method to remove shadow aliasing / create soft shadows
- Implemented deferred draw calls for most scene models
- Improved shadow map setup by setting frustum target to the scene AABB center
- Implemented character field shadows by projecting 3D models into the walkmesh
- Implemented method that extrudes the walkmesh to fade out sharp shadow discontinuities
- Implemented field shadow fade out with distance from occluder to the walkmesh floor
- Added ImGui window for adjusting parameters and debugging
- Improved field shadows using the stencil buffer to assure that they are only drawn over the background

More details here: http://forums.qhimm.com/index.php?topic=20612.0

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -6,6 +6,7 @@
 #    Copyright (C) 2020 Chris Rizzitello                                      #
 #    Copyright (C) 2020 John Pritchard                                        #
 #    Copyright (C) 2021 Julian Xhokaxhiu                                      #
+#    Copyright (C) 2021 Cosmos                                                #
 #                                                                             #
 #    This file is part of FFNx                                                #
 #                                                                             #
@@ -119,6 +120,77 @@ foreach(BGFX_VARYING flat smooth)
         COMMAND ${CMAKE_INSTALL_PREFIX}/vendor/bin/shaderc -i ${CMAKE_INSTALL_PREFIX}/vendor/include -f ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.vert -o ${CMAKE_INSTALL_PREFIX}/bin/shaders/FFNx.${BGFX_VARYING}.d3d12.vert --type v --varyingdef ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.varying.${BGFX_VARYING}.def.sc --platform windows -p vs_5_0 -O 3
     )
 endforeach()
+
+# LIGHTING SHADERS
+foreach(BGFX_VARYING flat smooth)
+    add_custom_command(
+        TARGET ${RELEASE_NAME}
+        POST_BUILD
+        # ensure bin directory exists
+        COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_INSTALL_PREFIX}/bin/shaders
+        # OpenGL
+        COMMAND ${CMAKE_INSTALL_PREFIX}/vendor/bin/shaderc -i ${CMAKE_INSTALL_PREFIX}/vendor/include -f ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.lighting.frag -o ${CMAKE_INSTALL_PREFIX}/bin/shaders/FFNx.lighting.${BGFX_VARYING}.gl.frag --type f --varyingdef ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.varying.${BGFX_VARYING}.def.sc --profile 120
+        COMMAND ${CMAKE_INSTALL_PREFIX}/vendor/bin/shaderc -i ${CMAKE_INSTALL_PREFIX}/vendor/include -f ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.lighting.vert -o ${CMAKE_INSTALL_PREFIX}/bin/shaders/FFNx.lighting.${BGFX_VARYING}.gl.vert --type v --varyingdef ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.varying.${BGFX_VARYING}.def.sc --profile 120
+	    # Vulkan
+        COMMAND ${CMAKE_INSTALL_PREFIX}/vendor/bin/shaderc -i ${CMAKE_INSTALL_PREFIX}/vendor/include -f ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.lighting.frag -o ${CMAKE_INSTALL_PREFIX}/bin/shaders/FFNx.lighting.${BGFX_VARYING}.vk.frag --type f --varyingdef ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.varying.${BGFX_VARYING}.def.sc --platform windows --profile spirv
+        COMMAND ${CMAKE_INSTALL_PREFIX}/vendor/bin/shaderc -i ${CMAKE_INSTALL_PREFIX}/vendor/include -f ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.lighting.vert -o ${CMAKE_INSTALL_PREFIX}/bin/shaders/FFNx.lighting.${BGFX_VARYING}.vk.vert --type v --varyingdef ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.varying.${BGFX_VARYING}.def.sc --platform windows --profile spirv
+        # Direct3D 9
+        COMMAND ${CMAKE_INSTALL_PREFIX}/vendor/bin/shaderc -i ${CMAKE_INSTALL_PREFIX}/vendor/include -f ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.lighting.frag -o ${CMAKE_INSTALL_PREFIX}/bin/shaders/FFNx.lighting.${BGFX_VARYING}.d3d9.frag --type f --varyingdef ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.varying.${BGFX_VARYING}.def.sc --platform windows -p ps_3_0 -O 3
+        COMMAND ${CMAKE_INSTALL_PREFIX}/vendor/bin/shaderc -i ${CMAKE_INSTALL_PREFIX}/vendor/include -f ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.lighting.vert -o ${CMAKE_INSTALL_PREFIX}/bin/shaders/FFNx.lighting.${BGFX_VARYING}.d3d9.vert --type v --varyingdef ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.varying.${BGFX_VARYING}.def.sc --platform windows -p vs_3_0 -O 3
+        # Direct3D 11
+        COMMAND ${CMAKE_INSTALL_PREFIX}/vendor/bin/shaderc -i ${CMAKE_INSTALL_PREFIX}/vendor/include -f ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.lighting.frag -o ${CMAKE_INSTALL_PREFIX}/bin/shaders/FFNx.lighting.${BGFX_VARYING}.d3d11.frag --type f --varyingdef ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.varying.${BGFX_VARYING}.def.sc --platform windows -p ps_5_0 -O 3
+        COMMAND ${CMAKE_INSTALL_PREFIX}/vendor/bin/shaderc -i ${CMAKE_INSTALL_PREFIX}/vendor/include -f ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.lighting.vert -o ${CMAKE_INSTALL_PREFIX}/bin/shaders/FFNx.lighting.${BGFX_VARYING}.d3d11.vert --type v --varyingdef ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.varying.${BGFX_VARYING}.def.sc --platform windows -p vs_5_0 -O 3
+        # Direct3D 12
+        COMMAND ${CMAKE_INSTALL_PREFIX}/vendor/bin/shaderc -i ${CMAKE_INSTALL_PREFIX}/vendor/include -f ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.lighting.frag -o ${CMAKE_INSTALL_PREFIX}/bin/shaders/FFNx.lighting.${BGFX_VARYING}.d3d12.frag --type f --varyingdef ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.varying.${BGFX_VARYING}.def.sc --platform windows -p ps_5_0 -O 3
+        COMMAND ${CMAKE_INSTALL_PREFIX}/vendor/bin/shaderc -i ${CMAKE_INSTALL_PREFIX}/vendor/include -f ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.lighting.vert -o ${CMAKE_INSTALL_PREFIX}/bin/shaders/FFNx.lighting.${BGFX_VARYING}.d3d12.vert --type v --varyingdef ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.varying.${BGFX_VARYING}.def.sc --platform windows -p vs_5_0 -O 3
+    )
+endforeach()
+
+# SHADOWMAP SHADERS
+add_custom_command(
+        TARGET ${RELEASE_NAME}
+        POST_BUILD
+        # ensure bin directory exists
+        COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_INSTALL_PREFIX}/bin/shaders
+        # OpenGL
+        COMMAND ${CMAKE_INSTALL_PREFIX}/vendor/bin/shaderc -i ${CMAKE_INSTALL_PREFIX}/vendor/include -f ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.shadowmap.frag -o ${CMAKE_INSTALL_PREFIX}/bin/shaders/FFNx.shadowmap.gl.frag --type f --varyingdef ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.varying.smooth.def.sc --profile 120
+        COMMAND ${CMAKE_INSTALL_PREFIX}/vendor/bin/shaderc -i ${CMAKE_INSTALL_PREFIX}/vendor/include -f ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.shadowmap.vert -o ${CMAKE_INSTALL_PREFIX}/bin/shaders/FFNx.shadowmap.gl.vert --type v --varyingdef ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.varying.smooth.def.sc --profile 120
+	    # Vulkan
+        COMMAND ${CMAKE_INSTALL_PREFIX}/vendor/bin/shaderc -i ${CMAKE_INSTALL_PREFIX}/vendor/include -f ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.shadowmap.frag -o ${CMAKE_INSTALL_PREFIX}/bin/shaders/FFNx.shadowmap.vk.frag --type f --varyingdef ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.varying.smooth.def.sc --platform windows --profile spirv
+        COMMAND ${CMAKE_INSTALL_PREFIX}/vendor/bin/shaderc -i ${CMAKE_INSTALL_PREFIX}/vendor/include -f ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.shadowmap.vert -o ${CMAKE_INSTALL_PREFIX}/bin/shaders/FFNx.shadowmap.vk.vert --type v --varyingdef ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.varying.smooth.def.sc --platform windows --profile spirv
+        # Direct3D 9
+        COMMAND ${CMAKE_INSTALL_PREFIX}/vendor/bin/shaderc -i ${CMAKE_INSTALL_PREFIX}/vendor/include -f ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.shadowmap.frag -o ${CMAKE_INSTALL_PREFIX}/bin/shaders/FFNx.shadowmap.d3d9.frag --type f --varyingdef ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.varying.smooth.def.sc --platform windows -p ps_3_0 -O 3
+        COMMAND ${CMAKE_INSTALL_PREFIX}/vendor/bin/shaderc -i ${CMAKE_INSTALL_PREFIX}/vendor/include -f ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.shadowmap.vert -o ${CMAKE_INSTALL_PREFIX}/bin/shaders/FFNx.shadowmap.d3d9.vert --type v --varyingdef ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.varying.smooth.def.sc --platform windows -p vs_3_0 -O 3
+        # Direct3D 11
+        COMMAND ${CMAKE_INSTALL_PREFIX}/vendor/bin/shaderc -i ${CMAKE_INSTALL_PREFIX}/vendor/include -f ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.shadowmap.frag -o ${CMAKE_INSTALL_PREFIX}/bin/shaders/FFNx.shadowmap.d3d11.frag --type f --varyingdef ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.varying.smooth.def.sc --platform windows -p ps_5_0 -O 3
+        COMMAND ${CMAKE_INSTALL_PREFIX}/vendor/bin/shaderc -i ${CMAKE_INSTALL_PREFIX}/vendor/include -f ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.shadowmap.vert -o ${CMAKE_INSTALL_PREFIX}/bin/shaders/FFNx.shadowmap.d3d11.vert --type v --varyingdef ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.varying.smooth.def.sc --platform windows -p vs_5_0 -O 3
+        # Direct3D 12
+        COMMAND ${CMAKE_INSTALL_PREFIX}/vendor/bin/shaderc -i ${CMAKE_INSTALL_PREFIX}/vendor/include -f ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.shadowmap.frag -o ${CMAKE_INSTALL_PREFIX}/bin/shaders/FFNx.shadowmap.d3d12.frag --type f --varyingdef ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.varying.smooth.def.sc --platform windows -p ps_5_0 -O 3
+        COMMAND ${CMAKE_INSTALL_PREFIX}/vendor/bin/shaderc -i ${CMAKE_INSTALL_PREFIX}/vendor/include -f ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.shadowmap.vert -o ${CMAKE_INSTALL_PREFIX}/bin/shaders/FFNx.shadowmap.d3d12.vert --type v --varyingdef ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.varying.smooth.def.sc --platform windows -p vs_5_0 -O 3
+    )
+
+# FIELD SHADOW SHADERS
+add_custom_command(
+    TARGET ${RELEASE_NAME}
+    POST_BUILD
+    # ensure bin directory exists
+    COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_INSTALL_PREFIX}/bin/shaders
+    # OpenGL
+    COMMAND ${CMAKE_INSTALL_PREFIX}/vendor/bin/shaderc -i ${CMAKE_INSTALL_PREFIX}/vendor/include -f ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.field.shadow.frag -o ${CMAKE_INSTALL_PREFIX}/bin/shaders/FFNx.field.shadow.gl.frag --type f --varyingdef ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.varying.smooth.def.sc --profile 120
+    COMMAND ${CMAKE_INSTALL_PREFIX}/vendor/bin/shaderc -i ${CMAKE_INSTALL_PREFIX}/vendor/include -f ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.field.shadow.vert -o ${CMAKE_INSTALL_PREFIX}/bin/shaders/FFNx.field.shadow.gl.vert --type v --varyingdef ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.varying.smooth.def.sc --profile 120
+    # Vulkan
+    COMMAND ${CMAKE_INSTALL_PREFIX}/vendor/bin/shaderc -i ${CMAKE_INSTALL_PREFIX}/vendor/include -f ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.field.shadow.frag -o ${CMAKE_INSTALL_PREFIX}/bin/shaders/FFNx.field.shadow.vk.frag --type f --varyingdef ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.varying.smooth.def.sc --platform windows --profile spirv
+    COMMAND ${CMAKE_INSTALL_PREFIX}/vendor/bin/shaderc -i ${CMAKE_INSTALL_PREFIX}/vendor/include -f ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.field.shadow.vert -o ${CMAKE_INSTALL_PREFIX}/bin/shaders/FFNx.field.shadow.vk.vert --type v --varyingdef ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.varying.smooth.def.sc --platform windows --profile spirv
+    # Direct3D 9
+    COMMAND ${CMAKE_INSTALL_PREFIX}/vendor/bin/shaderc -i ${CMAKE_INSTALL_PREFIX}/vendor/include -f ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.field.shadow.frag -o ${CMAKE_INSTALL_PREFIX}/bin/shaders/FFNx.field.shadow.d3d9.frag --type f --varyingdef ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.varying.smooth.def.sc --platform windows -p ps_3_0 -O 3
+    COMMAND ${CMAKE_INSTALL_PREFIX}/vendor/bin/shaderc -i ${CMAKE_INSTALL_PREFIX}/vendor/include -f ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.field.shadow.vert -o ${CMAKE_INSTALL_PREFIX}/bin/shaders/FFNx.field.shadow.d3d9.vert --type v --varyingdef ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.varying.smooth.def.sc --platform windows -p vs_3_0 -O 3
+    # Direct3D 11
+    COMMAND ${CMAKE_INSTALL_PREFIX}/vendor/bin/shaderc -i ${CMAKE_INSTALL_PREFIX}/vendor/include -f ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.field.shadow.frag -o ${CMAKE_INSTALL_PREFIX}/bin/shaders/FFNx.field.shadow.d3d11.frag --type f --varyingdef ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.varying.smooth.def.sc --platform windows -p ps_5_0 -O 3
+    COMMAND ${CMAKE_INSTALL_PREFIX}/vendor/bin/shaderc -i ${CMAKE_INSTALL_PREFIX}/vendor/include -f ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.field.shadow.vert -o ${CMAKE_INSTALL_PREFIX}/bin/shaders/FFNx.field.shadow.d3d11.vert --type v --varyingdef ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.varying.smooth.def.sc --platform windows -p vs_5_0 -O 3
+    # Direct3D 12
+    COMMAND ${CMAKE_INSTALL_PREFIX}/vendor/bin/shaderc -i ${CMAKE_INSTALL_PREFIX}/vendor/include -f ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.field.shadow.frag -o ${CMAKE_INSTALL_PREFIX}/bin/shaders/FFNx.field.shadow.d3d12.frag --type f --varyingdef ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.varying.smooth.def.sc --platform windows -p ps_5_0 -O 3
+    COMMAND ${CMAKE_INSTALL_PREFIX}/vendor/bin/shaderc -i ${CMAKE_INSTALL_PREFIX}/vendor/include -f ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.field.shadow.vert -o ${CMAKE_INSTALL_PREFIX}/bin/shaders/FFNx.field.shadow.d3d12.vert --type v --varyingdef ${CMAKE_SOURCE_DIR}/misc/${RELEASE_NAME}.varying.smooth.def.sc --platform windows -p vs_5_0 -O 3
+)
 
 # OVERLAY SHADERS
 add_custom_command(
--- a/src/renderer.cpp
+++ b/src/renderer.cpp
@@ -6,6 +6,7 @@
 //    Copyright (C) 2020 Chris Rizzitello                                   //
 //    Copyright (C) 2020 John Pritchard                                     //
 //    Copyright (C) 2021 Julian Xhokaxhiu                                   //
+//    Copyright (C) 2021 Cosmos                                             //
 //                                                                          //
 //    This file is part of FFNx                                             //
 //                                                                          //
@@ -20,6 +21,7 @@
 /****************************************************************************/
 
 #include "renderer.h"
+#include "lighting.h"
 
 Renderer newRenderer;
 RendererCallbacks bgfxCallbacks;
@@ -112,6 +114,26 @@ void Renderer::setCommonUniforms()
     setUniform("d3dViewport", bgfx::UniformType::Mat4, internalState.d3dViewMatrix);
     setUniform("d3dProjection", bgfx::UniformType::Mat4, internalState.d3dProjectionMatrix);
     setUniform("worldView", bgfx::UniformType::Mat4, internalState.worldViewMatrix);
+    setUniform("normalMatrix", bgfx::UniformType::Mat4, internalState.normalMatrix);
+    setUniform("viewMatrix", bgfx::UniformType::Mat4, internalState.viewMatrix);
+    setUniform("invViewMatrix", bgfx::UniformType::Mat4, internalState.invViewMatrix);
+}
+
+void Renderer::setLightingUniforms()
+{
+    auto lightingState = lighting.getLightingState();
+
+    setUniform("lightDirData", bgfx::UniformType::Vec4, lightingState.lightDirData);
+    setUniform("lightData", bgfx::UniformType::Vec4, lightingState.lightData);
+    setUniform("ambientLightData", bgfx::UniformType::Vec4, lightingState.ambientLightData);
+    setUniform("shadowData", bgfx::UniformType::Vec4, lightingState.shadowData);
+    setUniform("fieldShadowData", bgfx::UniformType::Vec4, lightingState.fieldShadowData);
+    setUniform("materialData", bgfx::UniformType::Vec4, lightingState.materialData);
+    setUniform("lightingDebugData", bgfx::UniformType::Vec4, lightingState.lightingDebugData);
+
+    setUniform("lightViewProjMatrix", bgfx::UniformType::Mat4, lightingState.lightViewProjMatrix);
+    setUniform("lightViewProjTexMatrix", bgfx::UniformType::Mat4, lightingState.lightViewProjTexMatrix);
+    setUniform("lightInvViewProjTexMatrix", bgfx::UniformType::Mat4, lightingState.lightInvViewProjTexMatrix);
 }
 
 bgfx::RendererType::Enum Renderer::getUserChosenRenderer() {
@@ -181,6 +203,14 @@ void Renderer::updateRendererShaderPaths()
     fragmentPostPath += shaderSuffix + ".frag";
     vertexOverlayPath += shaderSuffix + ".vert";
     fragmentOverlayPath += shaderSuffix + ".frag";
+    vertexLightingPathFlat += ".flat" + shaderSuffix + ".vert";
+    fragmentLightingPathFlat += ".flat" + shaderSuffix + ".frag";
+    vertexLightingPathSmooth += ".smooth" + shaderSuffix + ".vert";
+    fragmentLightingPathSmooth += ".smooth" + shaderSuffix + ".frag";
+    vertexShadowMapPath += shaderSuffix + ".vert";
+    fragmentShadowMapPath += shaderSuffix + ".frag";
+    vertexFieldShadowPath += shaderSuffix + ".vert";
+    fragmentFieldShadowPath += shaderSuffix + ".frag";
 
     // Update pipeline cache path for the callbacks
     std::filesystem::create_directories(bgfxCallbacks.cachePath);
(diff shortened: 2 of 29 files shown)


src/gl/texture.cpp
```diff
@@ -94,7 +95,7 @@ void gl_bind_texture_set(struct texture_set *_texture_set)
 {
 	VOBJ(texture_set, texture_set, _texture_set);
 
-	if(VPTR(texture_set))
+	if(VPTR(texture_set) && VREF(texture_set, texturehandle))
```
julianxhokaxhiu, line 98: What was the reason for this change? Be aware that this code must work on both FF7 and FF8 which not always share the same structure, although being very similar.

src/renderer.cpp
```diff
@@ -597,24 +645,140 @@ void Renderer::shutdown()
...
+    bgfx::setViewFrameBuffer(VIEW_ID_SCENE, backendFrameBuffer);
+
+    // Set current view rect
+    bgfx::setViewRect(VIEW_ID_SCENE, 0, 0, framebufferWidth, framebufferHeight);
+
+    if (internalState.bDoScissorTest) bgfx::setScissor(scissorOffsetX, scissorOffsetY, scissorWidth, scissorHeight);    
 
     // Set current view transform
-    bgfx::setViewTransform(backendViewId, NULL, internalState.backendProjMatrix);
+    bgfx::setViewTransform(VIEW_ID_SCENE, internalState.viewMatrix, internalState.backendProjMatrix);
```
julianxhokaxhiu, line 781: The idea of having a variable named `backendViewId` is also to highlight it's actually changing during runtime, while the notation `VIEW_ID_SCENE` seems to be a constant which is definitely not the case, as depending on which game we run (FF7 or FF8) and how many blit commands are required per-frame, the ID might change.

May I ask why this change?

src/renderer.cpp
```diff
@@ -701,7 +958,7 @@ void Renderer::draw()
     }
     bgfx::setState(internalState.state);
 
-    bgfx::submit(backendViewId, backendProgramHandles[backendProgram]);
+    bgfx::submit(VIEW_ID_POSTEFFECT, backendProgramHandles[backendProgram]);
```
julianxhokaxhiu, line 961: Also here, the idea of having the same var in each stage means we have a progressive view when dumping using RenderDoc. Maybe I'm missing something here, but I'm open for clarifications :) Thanks

src/renderer.cpp
```diff
@@ -826,7 +1088,7 @@ void Renderer::setScissor(uint16_t x, uint16_t y, uint16_t width, uint16_t heigh
     scissorHeight = getInternalCoordY(height);
 }
 
-void Renderer::setClearFlags(bool doClearColor, bool doClearDepth)
+void Renderer::setClearFlags(bgfx::ViewId viewId, bool doClearColor, bool doClearDepth)
```
julianxhokaxhiu, line 1091: Ideally there's no need to know which view to clear. This class was created with a complexity abstraction in mind, introducing this parameter it means killing this logic. Personally I'd prefer finding another way and keep the API call as it was meant to be.

src/renderer.cpp
```diff
@@ -1216,14 +1478,9 @@ uint32_t Renderer::blitTexture(uint32_t x, uint32_t y, uint32_t width, uint32_t
...
 
-    backendViewId++;
-
-    bgfx::blit(backendViewId, ret, 0, dstY, bgfx::getTexture(backendFrameBuffer), newX, newY, newWidth, newHeight);
-    bgfx::touch(backendViewId);
-    setClearFlags(false, false);
-
-    backendViewId++;
-    setClearFlags(false, false);
+    bgfx::blit(VIEW_ID_BLIT, ret, 0, dstY, bgfx::getTexture(backendFrameBuffer), newX, newY, newWidth, newHeight);
```
julianxhokaxhiu, line 1481: Same here, as said previous, the idea of a progressive ID is exactly to sort blits where they are supposed to be. In FF7 you'll not notice it, but on FF8 this code would kill a lot of scenes ( you will notice it the most during the final battle )

src/renderer.h
```diff
@@ -93,6 +94,14 @@ enum RendererTextureType
     YUV
 };
 
+enum ViewId
```
julianxhokaxhiu, line 97: This is not even supposed to exist, because of the aforementioned reasons.

julianxhokaxhiu (changes requested):
I have mixed feelings about this PR and the way code is being implemented.

On one side, I see that you put ton of efforts and I definitely understand the complexity behind this PR, as introducing light shadows support means also bringing a lot of wiring in the code.

On the other hand though I think it introduces too much changes as I had in mind:
1) The way the Renderer class is being reimplemented basically looks much more confusing to me and overall it will break FF8 support, so at the moment this point only is a showstopper for a merge
2) I don't understand why we do have new shaders just for the light part. The idea I had in mind was to implement it on the FFNx vert and frag core shaders, where for each draw call we compute the light too, which would come as a paramter from the engine ( what is basically computed on the Software side, will be now done on the GPU side ). If you do although implement them in a post-processing logic, then I'd prefer deferring this to ReShade.
3) The fact we have two shaders one for the World and one for fields is a no-go. We can distringuish which call is for the "2D" rendered calls and which one for 3D through `isTLVertex`, and I think we should base the entire logic around it, again on the core shaders.

Because of this I'd need to confront you on all those arguments and maybe let me understand if I'm missing something in here, or I'm not seeing some more complexity behind this overall goal.

I do appreciate although the ability to "play with this code" using the DevTools, it makes it easier to test and understand it.

Aside of the way you wired up things, I just want to let you know that I'd like to see this moving forward and not just stopping this. Reviewing means also learning something from each other. so as much as I'm opened to learn from you, I hope you'll see these points as constructive towards you.

Thank you in advance and looking forward to your own reply.

src/gl/texture.cpp
CosmosXIII: I added this to check that the texture handle is valid before accessing the array.
The reason being that when rendering the deferred models as implemented in deferred.cpp I was getting an access violation here. You can test it if you remove this check, load the game, access the menu, then close.
I suppose that this is caused because when saving the render state with gl_save_state it does not make a copy of the contents of ff7_texture_set::texturehandle and under certain conditions the memory is deleted before use.

There may be a better way to prevent this problem so please let me know if you have some idea.

src/renderer.cpp
CosmosXIII: The reason for this change is that when using `backendViewId` variable to specify the view id it was very confusing for me to know what view id was used at any time.
For the changes in this PR I needed to draw the shadow map in a view id lower than normal scene rendering because the shadow map is being referred as a texture during the normal scene rendering.
I think that making the view ids constant in an enum makes it clear to which view id is being drawn. By making view ids we can assure that the shadow map pass is always view id 0, normal scene rendering is 1, and posteffect is 2. I think this also makes it clear when debugging with RenderDoc.

That said, I did not know that multiple blit commands could happen in a single frame. I thought that was only used for the swirl effect in ff7 but I do not currently own a copy of ff8 so I cannot currently check if something is broken there (maybe I'll buy a copy this weekend if necessary).

I was wondering, do blit commands need to happen in a lower view id than normal scene rendering?
If not necessary maybe blit commands could start at a constant view id and have a `blitViewId` variable that increases after each blit command. Something like this:
enum VIEW_ID
{
    VIEW_ID_SHADOWMAP,
    VIEW_ID_SCENE,
    VIEW_ID_POSTEFFECT,
    VIEW_ID_BLIT_START,
};

// At start of each frame
blitViewId = VIEW_ID_BLIT_START;

void blitTexture(...)
{
     ...
     bgfx::blit(blitViewId, ...);
     ++blitViewId;
}

This would enable to have constant view ids that make it clear the purpose of each draw call as well as allowing variable view ids for the blit commands where it is necessary. What do you think?

If the above does not look good to you I am ok reverting to using `backendViewId`.

src/renderer.cpp
CosmosXIII: Please see the comment above but other than the blit command problem, the other view ids are still progressive.
The shadow map pass rendering is view id  0, normal scene rendering is 1, and posteffect is 2.

src/renderer.cpp
CosmosXIII: Please see comment above. If you decide its best to revert to using `backendViewId` then I will revert this API to its previous state.

src/renderer.cpp
CosmosXIII: Please see comment above. Unfortunately I do not own a copy of FF8 right now so I was not able to check if something is broken.

src/renderer.h
CosmosXIII: Please see comment above. If you decide its best to revert to using backendViewId then I will delete this enum.

CosmosXIII:
Thank you for your code review!

I appreciate the time you dedicated to review the changes in this PR. Any criticism is welcome so I hope that with your feedback we can make it to an acceptable state for merging.

I will try explaining my reasoning for the changes you listed.

1. About the Renderer class, I made a comment above so please look a that one for a more detailed answer.

2. First, I want to ask, by "light part" you mean the shader that renders to the shadow map from the light point of view?
(FFNx.shadowmap.vert/frag)
If that is the case, I separated the implementation into two shaders for the following reasons:
- For code readability. Having a lot of if branches makes it a bit confusing understanding the flow of the shader program.
- For performance. I may be wrong on this one, but my understanding is that having too many if branches in shaders is not a good idea for performance. Please take a look at the following link for more info: [https://shader-tutorial.dev/advanced/branching/](url)
That said, it seems that moderns GPUs are good at optimizing when the if condition only depends on uniform variables.
Code readability aside, a good compromise may be use the --define option in the shaderc.exe compiler to make pre-compiler bifurcations inside the FFNx core shader. If performance here is not something to worry too much about I could changed it to uniform bifurcation but please let me know what you think.
Also, just for clarification, shadows are not implemented as post-processing. It is typical shadow mapping where the scene is rendered first from the point o view of the light into a texture (the shadowmap), then fetched during normal rendering.

3. By two shader, do you mean the core FFNx shader and the shader used to draw the walkmesh for field shadow?(FFNx.fieldShadow.vert/frag)
If that's the case, one reason for splitting the implementation, is that I though that it would be better separating what is originally rendered by the game and what is added as new functionality. That said, I think it would be possible implementing it in the same shader files by using the pre-compiled bifurcations explained in 2 ( or just uniforms if performance is not something to worry about).

Thank you again!

CosmosXIII:
I removed the following unused file: FFNx.shadow.sh
Sorry for the mistake.

src/gl/texture.cpp
julianxhokaxhiu: Interesting finding, not sure why no one was able to trigger this before. I'll try to reproduce and let you know, thanks!

src/renderer.cpp
julianxhokaxhiu: Thanks for clarifying this, I get the complexity behind but you can easily workaround that but simply injecting your own code in the `::draw()` call without needing to revamp the overall view ID logic. Just create your own `::drawShadow()` method and call it in the first function body of the `::draw()` method.

Regarding mixing constant view ids with variable ones I think it will be bad because everytime you shift the viewID you need to shift also the constants value. In FF7 blits are used only for the swirl, but in FF8 they're used to optimize some draw calls because in the PSX era it was nearly impossible to render that 3D effects everytime from scratch ( the game would have run half the FPS speed if not less ), so they came up with the concept of blitting the last rendered frame and reusing it, if the next draw call allows it.

Because of this, I'd prefer keeping the current renderer code the way it is, and not touching this logic as it has been battle tested on FF8, and I think it's one of the most delicate pillars of the rendering class itself.

Thanks and sorry :)

julianxhokaxhiu:
> Thank you for your code review!
> 
> I appreciate the time you dedicated to review the changes in this PR. Any criticism is welcome so I hope that with your feedback we can make it to an acceptable state for merging.

Thank you actually, I really appreciate this! I'm sure we will find the best way, I'm all about merging but also fan of KISS :) Let's rock and roll!

> I will try explaining my reasoning for the changes you listed.
> 
> 1. About the Renderer class, I made a comment above so please look at that one for a more detailed answer.
Yes, see also my relative reply on this point. All clear.

> 2. First, I want to ask, by "light part" you mean the shader that renders to the shadow map from the light point of view?
>    (FFNx.shadowmap.vert/frag)
>    If that is the case, I separated the implementation into two shaders for the following reasons:
> 
> * For code readability. Having a lot of if branches makes it a bit confusing understanding the flow of the shader program.
> * For performance. I may be wrong on this one, but my understanding is that having too many if branches in shaders is not a good idea for performance. Please take a look at the following link for more info: [https://shader-tutorial.dev/advanced/branching/](url)
>   That said, it seems that moderns GPUs are good at optimizing when the if condition only depends on uniform variables.
>   Code readability aside, a good compromise may be use the --define option in the shaderc.exe compiler to make pre-compiler bifurcations inside the FFNx core shader. If performance here is not something to worry too much about I could change it to uniform bifurcation but please let me know what you think.
>   Also, just for clarification, shadows are not implemented as post-processing. It is typical shadow mapping where the scene is rendered first from the point of view of the light into a texture (the shadowmap), then fetched during normal rendering.

You're right that was the case and indeed most of GPU devs suggest to keep shaders as simple as possible avoiding bifurcation, although I think the current shader is not that heavy that needs so much optimization. Let's KISS this part too, and if you're fine with it, implement everything you need into a helper include file, and feel free to call the shadow methods inside the core FFNx frag and vert files!

If you need to trigger only the shadow code, feel free to attach a uniform with a flag like it is being done now ( and actually like you did already ) and if that code through your own flags.

> 3. By two shader, do you mean the core FFNx shader and the shader used to draw the walkmesh for field shadow?(FFNx.fieldShadow.vert/frag)
>    If that's the case, one reason for splitting the implementation, is that I though that it would be better separating what is originally rendered by the game and what is added as new functionality. That said, if mixing the original and new  rendering functionality is not a problem, I think it would be possible implementing it in the same shader files by using the pre-compiled bifurcations explained in 2 ( or just uniforms if performance is not something to worry about).

Same as before, let's KISS this part too. If there will be a need for optimization, we will think about this later :)

> Thank you again!

Thank you actually, looking forward for the next batch of changes! Great job so far, I'm really impressed!

src/gl/texture.cpp
CosmosXIII: I have investigated more about this problem and was able to fix it without modifying texture.cpp.
As said before, when closing the menu there seems to be some texture unloading and the gl_check_deferred function in deferred.cpp is called. This function deletes the memory allocated and invalidates the draw calls. I just added a check in the deferred draw loop to omit invalidated draw calls.
Before the changes of this PR only a few special case draw calls needed to be deferred so in those cases the problem described above simply does not occur.

src/renderer.cpp
CosmosXIII: Thank you for the explanation.

> so they came up with the concept of blitting the last rendered frame and reusing it, if the next draw call allows it.
This seems to imply that the result of blit operations is only used in the next frame. Would this not mean that it does not matter if blit operations are done last and allow the first view ids to be constant?

Anyway, I have reverted the view ID logic to use `backendViewId` variable. It is now exactly as before with just one difference. The `backendViewId` variable now starts at view ID 1 instead of 0. This is because as explained before, I need the shadow map to be drawn in a view ID with smaller value to be able to reference it as a texture during normal rendering. Please let me know what you think.

CosmosXIII: By the way, I said I may buy FF8 to check that nothing is broken but unfortunately it is region locked, meaning that it is not available for purchase or activation in my country :(

src/renderer.cpp
CosmosXIII: As explained in the previous comment, I reverted to using the `backendViewId` variable.

src/renderer.cpp
CosmosXIII: As explained in the previous comment, I reverted to using the `backendViewId` variable so this API was also reverted.

src/renderer.cpp
CosmosXIII: As explained in the previous comment, I reverted to using the `backendViewId` variable.

src/renderer.h
CosmosXIII: As explained in the previous comment, I reverted to using the `backendViewId` variable so I deleted this enum.

CosmosXIII:
Thanks again for your feedback.

I made a few commits to fix the problems raised in your review.
1.  I reverted the view ID logi to use the `backendViewId` variable so please take a look at the comment above for more details.
2.  I integrated the shadow map rendering into the core FFNx shaders and removed the previous shader files.  Bifurcation is done using uniforms as previously discussed.
3. I also integrated  the field shadow rendering into the core FFNx shaders and removed the previous shader files. 

Let me know what you think!

src/CMakeLists.txt
```diff
@@ -6,6 +6,7 @@
 #    Copyright (C) 2020 Chris Rizzitello                                      #
 #    Copyright (C) 2020 John Pritchard                                        #
 #    Copyright (C) 2021 Julian Xhokaxhiu                                      #
+#    Copyright (C) 2021 Cosmos                                                #
```
julianxhokaxhiu, line 9: If you didn't change anything here, please consider removing this line, thanks :)

src/gl/texture.cpp
```diff
@@ -6,6 +6,7 @@
 //    Copyright (C) 2020 Chris Rizzitello                                   //
 //    Copyright (C) 2020 John Pritchard                                     //
 //    Copyright (C) 2021 Julian Xhokaxhiu                                   //
+//    Copyright (C) 2021 Cosmos                                             //
```
julianxhokaxhiu, line 9: Same as before :)

src/renderer.cpp
```diff
@@ -596,25 +625,173 @@ void Renderer::shutdown()
 
     bgfx::shutdown();
 }
+void Renderer::clearShadowMap()
+{
+    // Save current view id
+    int tmpBackendViewId = backendViewId;
```
julianxhokaxhiu, line 631: Is this code really needed?

src/renderer.cpp
```diff
@@ -596,25 +625,173 @@ void Renderer::shutdown()
...
+{
+    // Save current view id
+    int tmpBackendViewId = backendViewId;
 
-void Renderer::draw()
+    backendViewId = 0;
+    setClearFlags(false, true);
+
+    // Revert view id to its original value
+    backendViewId = tmpBackendViewId;
```
julianxhokaxhiu, line 637: Same for this one

src/renderer.cpp
```diff
@@ -596,25 +625,173 @@ void Renderer::shutdown()
...
+
+void Renderer::draw()
+{
+    if (enable_lighting && internalState.isCastShadow)
+    {
+        // Save current view id
+        int tmpBackendViewId = backendViewId;
+
+        backendViewId = 0;
+        drawShadowMap();
```
julianxhokaxhiu, line 774: If you need to ensure that your code runs before the current draw call, just fire `drawShadowMap` as the first line after
```cpp
::draw()
{
 // your code here
 
 // rest of the code here
}
```
without mangling the current `backendViewId`. Or am I missing something?

src/renderer.cpp
```diff
@@ -1228,6 +1506,11 @@ uint32_t Renderer::blitTexture(uint32_t x, uint32_t y, uint32_t width, uint32_t
     return ret.idx;
 };
 
+void Renderer::setDrawType(DrawType drawType)
```
julianxhokaxhiu, line 1509: If this function is about setting the Light draw type, it should be named accordingly `setLightDrawType` no?

src/renderer.cpp
```diff
@@ -1357,6 +1750,248 @@ void Renderer::setD3DProjection(struct matrix* matrix)
...
+}
+
+float Renderer::getWalkmeshExtrudeSize()
+{
+    return walkMeshExtrudeSize;
+}
+
+void Renderer::setWalkmeshPosOffset(float offset)
+{
+    walkMeshPosOffset = offset;
```
julianxhokaxhiu, line 1967: Shouldn't this be part of the `internalState`?

src/renderer.cpp
```diff
@@ -1357,6 +1750,248 @@ void Renderer::setD3DProjection(struct matrix* matrix)
...
+}
+
+float Renderer::getFieldShadowFadeRange()
+{
+    return internalState.fieldShadowData[2];
+}
+
+void Renderer::setWalkmeshExtrudeSize(float size)
+{
+    walkMeshExtrudeSize = size;
```
julianxhokaxhiu, line 1957: Shouldn't this be part of the `internalState`?

src/renderer.cpp
```diff
@@ -1357,6 +1750,248 @@ void Renderer::setD3DProjection(struct matrix* matrix)
...
+}
+
+float Renderer::getFieldShadowMapArea()
+{
+    return fieldShadowMapArea;
+}
+
+void Renderer::setFieldShadowMapNearFarSize(float size)
+{
+    fieldShadowMapNearFarSize = size;
```
julianxhokaxhiu, line 1917: Shouldn't this be part of the `internalState`?

src/renderer.cpp
```diff
@@ -1357,6 +1750,248 @@ void Renderer::setD3DProjection(struct matrix* matrix)
...
+}
+
+float Renderer::getShadowMapNearFarSize()
+{
+    return shadowMapNearFarSize;
+}
+
+void Renderer::setFieldShadowMapArea(float area)
+{
+    fieldShadowMapArea = area;
```
julianxhokaxhiu, line 1907: Shouldn't this be part of the `internalState`?

src/renderer.cpp
```diff
@@ -1357,6 +1750,248 @@ void Renderer::setD3DProjection(struct matrix* matrix)
...
+}
+
+float Renderer::getShadowMapArea()
+{
+    return shadowMapArea;
+}
+
+void Renderer::setShadowMapNearFarSize(float size)
+{
+    shadowMapNearFarSize = size;
```
julianxhokaxhiu, line 1897: Shouldn't this be part of the `internalState`?

src/renderer.cpp
```diff
@@ -1357,6 +1750,248 @@ void Renderer::setD3DProjection(struct matrix* matrix)
...
+}
+
+float Renderer::getShadowConstantBias()
+{
+    return internalState.shadowData[0];
+}
+
+void Renderer::setShadowMapArea(float area)
+{
+    shadowMapArea = area;
```
julianxhokaxhiu, line 1887: Shouldn't this be part of the `internalState`?

src/renderer.cpp
```diff
@@ -1357,6 +1750,248 @@ void Renderer::setD3DProjection(struct matrix* matrix)
...
+}
+
+float Renderer::getMetalness()
+{
+    return internalState.materialData[1];
+}
+
+void Renderer::setShadowFaceCullingEnabled(bool isEnabled)
+{
+    isShadowMapFaceCullingEnabled = isEnabled;
```
julianxhokaxhiu, line 1837: Shouldn't this be part of the `internalState`?

src/renderer.cpp
```diff
@@ -1366,3 +2001,151 @@ uint16_t Renderer::getInternalCoordY(uint16_t inY)
 {
     return (inY * framebufferHeight) / game_height;
 }
+
+void lighting_debug(bool* isOpen)
```
julianxhokaxhiu, line 2005: This code shhould not be part of this class, as you notice for DevTools we have separate files containing this code. Cosinder moving this to a dedicated `.cpp` file and separate class named as you prefer ( `light.cpp` maybe? )

src/renderer.h
```diff
@@ -250,6 +287,24 @@ class Renderer {
...
+    float shadowMapNearFarSize = 20000.0f;
+
+    // Field shadowmap frustum parameters
+    float fieldShadowMapArea = 3000.0f;
+    float fieldShadowMapNearFarSize = 3000.0f;
+    
+    // Field Shadow walkmesh parameters
+    float walkMeshExtrudeSize = 20.0;
+    float walkMeshPosOffset = -11.0;
+
```
julianxhokaxhiu, line 307: As mentioned before, this should be part of the `internalState` structure, as the global variables here are only related to the `bgfx` class, whatever we do with it, and mutates ( get/set ) it must be in the `internalState`.

src/renderer.h
```diff
@@ -343,13 +404,71 @@ class Renderer {
...
+    bool isShowWalkmeshEnabled();
+
     // Internal coord calculation
     uint16_t getInternalCoordX(uint16_t inX);
     uint16_t getInternalCoordY(uint16_t inY);
 };
 
 extern Renderer newRenderer;
+
+void lighting_debug(bool* isOpen);
```
julianxhokaxhiu, line 474: This one must be moved in a relative header file ( `light.h` maybe? )

julianxhokaxhiu:
Thank you for the commits. The PR looks much better although there are still some rough corners to be adjusted.

Please consider squashing all your commits into one and force push your branch, so I do see only your latest status, when asking for a new review.

Additionally I still see a lot of changes in the `renderer.cpp` class which I think we can do better ( like we have too many `draw` commands, while there should be only one, then what we draw, how and why, it's up to the class to be managed ). I'm not sure if those new `draw` APIs are public or private ( it's hard to spot it from GH ) but I'll double check this later when checking out the code from my own PC.

Nevertheless, I think we're on the right track now and it's only a matter of hammering some edges. We'll get there, great job so far, and thank you again.

src/CMakeLists.txt
CosmosXIII: Sorry I forgot to remove this line when I deleted the other changes, it has been deleted now.

src/gl/texture.cpp
CosmosXIII: Sorry I forgot to remove this line when I deleted the other changes, it has been deleted now.

src/renderer.cpp
CosmosXIII: I did this because the SetClearFlags API uses the `backendViewId` and I need to have the viewId 0  reserved for rendering the shadowmap (more on why this is needed in the comment below).

Anyway, I removed the need to temporally store the `backendViewId` variable by just calling bgfx::setViewClear directly.

src/renderer.cpp
CosmosXIII: Removed the need for this change, see comment above.

src/renderer.cpp
CosmosXIII: That would not work because I need the whole scene to be drawn to the shadowmap before any normal rendering is done, not just the model currently being rendered.
If I just called `drawShadowMap` without changing the view Id to a lower value (0 for the shadowmap), only the currently model being drawn and previously drawn models would exist in the shadow map referenced during normal rendering.
For example if there is box placed on the floor, and the floor is drawn first, the box would not have been drawn to the shadowmap yet and therefore no shadow would be cast on the floor. 
Hope this clears out the reason why there is a need to reserve view id 0 for the shadowmap rendering.

src/renderer.cpp
CosmosXIII: I intentionally set the name somehow ambiguous because this may be used for other purposes than lighting in the future.

src/renderer.cpp
CosmosXIII: I moved this variable to the `internalState`.

src/renderer.cpp
CosmosXIII: I moved this variable to the `internalState`.

src/renderer.cpp
CosmosXIII: I moved this variable to the `internalState`.

src/renderer.cpp
CosmosXIII: I moved this variable to the `internalState`.

src/renderer.cpp
CosmosXIII: I moved this variable to the `internalState`.

src/renderer.cpp
CosmosXIII: I moved this variable to the `internalState`.

src/renderer.cpp
CosmosXIII: I moved this variable to the `internalState`.

src/renderer.cpp
CosmosXIII: I added `lighting_debug.h` and  `lighting_debug.cpp` and moved this code there.
I also modified the code so that it is only called for FF7.

src/renderer.h
CosmosXIII: I moved these variables to the internalState.

src/renderer.h
CosmosXIII: I moved it to the newly added `lighting_debug.h` file.

CosmosXIII:
Thanks again for you review!

I made some more changes and squashed all commits into a single one as you requested.
I thought that at least the postprocessing draw function could be preserved separately because it even uses a different shader in the same way that overlay already has a different draw function.
Anyway, for now in order to reduce the amount of changes in the Renderer class as much as possible, I consolidated all draw commands into a single draw function so it should be the closest the original code now.

Let me know what you think!

src/renderer.cpp
julianxhokaxhiu: Thanks, this logic won't work on FF8 though as in some scenes the view 0 may be inheriting the previous scene. Please consider using another approach. Also, why not clearing the shadow map everytime we start a new frame? Wouldn't that work?

src/renderer.cpp
julianxhokaxhiu: Aha, ok then you can consider doing that in this in the method: https://github.com/julianxhokaxhiu/FFNx/blob/master/src/renderer.cpp#L317

Because it's the best moment as you get the entire scene, drawn already and before showing it on the Backbuffer, you can do all the magic you need to ( basically it's also where the "post" effects are done, which atm are a simple plain rendering of the Framebuffer in the Backbuffer ).

Wouldn't this work? Also this would guarantee you have your own entire attachments cleared and so on.

src/renderer.cpp
julianxhokaxhiu: I do prefer having it named accordingly when the body of the function manipulates only light draw types. When and if that will happen, I do prefer to see a commit changing the name of this method, instead of reserving it :) Thanks

src/renderer.cpp
```diff
@@ -312,6 +334,9 @@ void Renderer::resetState()
     doModulateAlpha();
     doTextureFiltering();
     isExternalTexture();
+    setCastShadow();
+
+    internalState.lightingSettings[0] = enable_lighting;
```
julianxhokaxhiu, line 339: Technically this flag should be set only once in a lifecycle, or can it be resetted through the debug devtool?

src/renderer.cpp
```diff
@@ -596,25 +626,50 @@ void Renderer::shutdown()
 
     bgfx::shutdown();
 }
+void Renderer::clearShadowMap()
```
julianxhokaxhiu, line 629: See the relative comment I left about this topic. We should avoid reserving view IDs.

src/renderer.cpp
```diff
@@ -596,25 +626,50 @@ void Renderer::shutdown()
...
 void Renderer::draw()
 {
     if (trace_all || trace_renderer) trace("Renderer::%s with backendProgram %d\n", __func__, backendProgram);
 
-    // Set current view rect
-    if (backendProgram == RendererProgram::POSTPROCESSING)
-        bgfx::setViewRect(backendViewId, 0, 0, window_size_x, window_size_y);
-    else {
+    auto drawType = getDrawType();
+    if (drawType == DrawType::SHADOW_MAP)
```
julianxhokaxhiu, line 640: Ok now I get what you meant with draw types. Please then change the internalState variable and name it generically too! Thanks :)

src/renderer.cpp
```diff
@@ -631,33 +686,52 @@ void Renderer::draw()
...
-                }
-                else
-                {
-                    if (internalState.bIsMovie) flags |= BGFX_SAMPLER_U_CLAMP | BGFX_SAMPLER_V_CLAMP | BGFX_SAMPLER_W_CLAMP;
-
-                    if (!internalState.bDoTextureFiltering || !internalState.bIsExternalTexture) flags |= BGFX_SAMPLER_MIN_POINT | BGFX_SAMPLER_MAG_POINT | BGFX_SAMPLER_MIP_POINT;
+                if (!internalState.bDoTextureFiltering || !internalState.bIsExternalTexture) flags |= BGFX_SAMPLER_MIN_POINT | BGFX_SAMPLER_MAG_POINT | BGFX_SAMPLER_MIP_POINT;
 
-                    if (flags == 0) flags = UINT32_MAX;
-                }
```
julianxhokaxhiu, line 646: What happened to this entire block of code? Why is it gone?

src/renderer.cpp
```diff
@@ -1357,6 +1604,248 @@ void Renderer::setD3DProjection(struct matrix* matrix)
...
+        1,
+        bgfx::TextureFormat::D32,
+        BGFX_TEXTURE_RT | BGFX_SAMPLER_COMPARE_LEQUAL | BGFX_SAMPLER_U_CLAMP | BGFX_SAMPLER_V_CLAMP | BGFX_SAMPLER_W_CLAMP
+    );
+
+    shadowMapFrameBuffer = bgfx::createFrameBuffer(
+        1,
+        &shadowMapTexture,
+        true
+    );
```
julianxhokaxhiu, line 1721: Please consider reworking this entire code to recreate the framebuffer on the `reset()` method call only.

src/renderer.cpp
CosmosXIII: Yes, lighting can be enabled/disabled from the debug devtools.

src/renderer.cpp
CosmosXIII: I renamed it to LightingDrawType.

src/renderer.cpp
CosmosXIII: No, that won't work because as I explained before, drawing to the shadowmap is a pre-process that needs to be done _**before**_ normal rendering, not as a post-process. The renderFrame method is called _**after**_ normal rendering and as you said it's where post effects are done.

I don't know if you are familiar with the shadowmap algorithm but it is not a post-process. To be fair, field shadows on the background are implemented somehow like a post-process where shadows are drawn by darkening the 2D field background. However, this is a physically incorrect way of doing shadows, but because field backgrounds are just 2D, the only way is to fake it as a post-process. 

On the other hand, characters and the battle background are all 3D models where accurate lighting and shadows can be achieved. Here, shadows are not a post-process but are calculated during normal rendering by fetching the pre-computed shadowmap and using it as an occlusion term to the diffuse and specular components of the luminance calculation:
[https://github.com/julianxhokaxhiu/FFNx/blob/427166ede3b2b33dc5a8c959c63fb9d48d2bc212/misc/FFNx.lighting.sh#L100](url)

Please notice that the shadow factor only applies to the diffuse and specular terms and _**not**_ the ambient term. This calculation cannot be achieved as a post-process in a forward renderer such as the one used in FFNx. By the way, shadow mapping as explained above is nothing unusual, it is how shadows are implemented in most games.

As explained above, the shadowmap rendering is a pre-process step and therefore requires a view id  with a value lower than the normal rendering view id. It is not possible to just use the same view id as normal rendering and call the submit function for shadowmap rendering before normal rendering. This is because the following reasons:
- The shadowmap rendering uses a different framebuffer where the scene is drawn from the light point of view, not the game camera point of view
- The shadowmap needs to be referenced as a texture during normal rendering

Therefore, I think that reserving view id 0 for the shadowmap rendering is the most reasonable way of doing it. As explained before, the only change now is that the 'backendViewId' variable now starts at 1 but that should not change the behavior of the renderer in any way.

src/renderer.cpp
CosmosXIII: See the comment above for the reasons I think reserving view id 0 for shadowmap rendering is reasonable.

src/renderer.cpp
CosmosXIII: I changed he name to internalState variable name to lightingDrawSettings.

src/renderer.cpp
CosmosXIII: Sorry, my mistake, I reverted the changes so it should be as before now.

src/renderer.cpp
CosmosXIII: If I understand correctly, implementing it inside `reset()` would mean that shadow map resolution can only be set during initialization or when the window size is resized. However, shadow map resolution can also be set from the lighting debug window so moving this code to the `reset()` would not work.

src/renderer.cpp
CosmosXIII: I may be missing something but I don't understand how FF8 would use view Id 0 if the `backendViewId` variable now starts at view Id 1 and it is reset to 1 every frame.

src/renderer.cpp
julianxhokaxhiu: Debug code should have its own special workflow. If it needs to reset the entire view, so be it. It's a dev tool afterall.

Generically speaking we develop towards optimizations for the generic experience, and because of this, I rather prefer this code being in the `reset()`. Debug code shall never degrade the overall code quality, just because it's there.

src/renderer.cpp
julianxhokaxhiu: Find a way to interactively reset it through a class method instead of changing a global variable which is then set on each pass. Thanks.

src/renderer.cpp
julianxhokaxhiu: Allright, I understand the whole point and now it starts to make sense.

So, if that's the case I suggest then in order to tackle this PR definitely:
1. Consider moving back the shadow vert and frag logics into separate shader files ( sorry! )
2. Fine by reserving the view id 0, just make sure in the code the shaders attached to that view are only the one required for the shadowmap calculation
3. Once all of the above is done, PLEASE consider testing your own code on FF8 and make sure you didn't break anything

Finally, as you have shielded all the light code into a flag, please make sure the shadowmap processing logics ( methods, code, etc. ) are as isolated as possible from the main code ( which means do not change AT ALL the `draw()` method code, just add your own method call where you need to ). This will make the code also much more easily to be managed.

Additionally please do not bind the whole uniform set of variables if those are required only by the Shadowmap, make sure those are attached only to your own shaders ( which means do not change ANYTHING from the existing core and post frag and vert files ).

Let's see once this whole bunch of changes are done how the code looks like and if it's good, the only part I'll test is how it works on FF8 just to make sure we didn't break anything.

Just to be clear, what I expect after this new set of changes are:
1) Minimal changes to the existing code ( by minimal I mean adding your function calls where needed, but the structure is not changed, not like how the `renderer.cpp` looks like now, or the `common.cpp` too
2) A ton of new files with their own logics
3) A complete new set of lines in the `renderer.cpp` which are your own new methods and members required to accomplish the goal
4) MINIMAL glue code, shielded by the flag and by a check which ensure it's running only on ff7

Looking forward into this :) Sorry for the challange, but this last comment made me recognize the shape of this code and how I expect it to be now.

src/renderer.cpp
julianxhokaxhiu: As a kind reminder then, consider using the prefix `LIGHT_DRAW_TYPE` in your own const names in the shader code too :) Thanks

src/renderer.cpp
julianxhokaxhiu: Replied, thanks

src/renderer.cpp
CosmosXIII: Thank you for understanding the need to reserve view id 0 for shadowmap rendering.
> Consider moving back the shadow vert and frag logics into separate shader files ( sorry! )

May I ask what made you change your mind on this? Also does this refer only to the shadowmap rendering shader or also the shader that rendered the walkmesh for the field shadows?

> Once all of the above is done, PLEASE consider testing your own code on FF8 and make sure you didn't break anything

As I said in a previous comment, FF8 is region blocked and not available for purchase/activation in my country, meaning that it is impossible for me to test if anything is broken in FF8. I would appreciate if you can test it for me, but if you are not available then maybe I can ask for help to someone else on the qhimm community.

> Finally, as you have shielded all the light code into a flag, please make sure the shadowmap processing logics ( methods, code, etc. ) are as isolated as possible from the main code ( which means do not change AT ALL the draw() method code, just add your own method call where you need to ). This will make the code also much more easily to be managed.

If I understand correctly, this means reverting to a similar state as when I first submitted this PR, where a separate drawShadowMap() function was used, is that correct? Also may I ask why this change of mind?

Also you say to not change anything in the draw() function but I need the following two changes for it to work:
1. Bind the shadowmap texture. This could be done without touching the draw function with the Renderer::useTexture() function, is that correct?
2. Set the stencil buffer. There is not currently a way to set the stencil buffer so I was thinking to add some APIs to set the stencil function and reference value from outside the draw() function. However, the actual setting by calling the bgfx functions would be inside the draw() function in similar to how the blend state is set. Would that be ok?

> Additionally please do not bind the whole uniform set of variables if those are required only by the Shadowmap, make sure those are attached only to your own shaders ( which means do not change ANYTHING from the existing core and post frag and vert files ).

You say here to not change anything from the core shaders, but lighting calculations as well as fetching the shadowmap needs to be done here so I'm not sure I understand your request (most of it is already is in include files anyway). 

> Minimal changes to the existing code ( by minimal I mean adding your function calls where needed, but the structure is not changed, not like how the renderer.cpp looks like now, or the common.cpp too

If I understand you correctly, this means reverting to something similar as when I was using separate draw functions such as drawShadowMap() but instead of implementing them inside the Renderer class, I create a different class and source files (lighting.h/cpp) that contains the lighting related draw functions as well as the code found in common.cpp. Is this correct?

> A ton of new files with their own logics

Other than the lighting class files described above, I was thinking to separate the walkmesh extraction/extrusion functions into separate files (currently in field.h/cpp). Other than that I don't see what else would require new files. I would appreciate if you specify what else you want me to separate.

Thank you again for your time reviewing this PR, it is very much appreciated!

julianxhokaxhiu: > Thank you for understanding the need to reserve view id 0 for shadowmap rendering.

Thank you for taking the patience for still working on this. Really, really appreciate it!

> May I ask what made you change your mind on this? Also does this refer only to the shadowmap rendering shader or also the shader that rendered the walkmesh for the field shadows?

As we do reserve an entire view, it makes sense to use also separate shaders. As you will probably fire your own draw calls or whatever needed, you can totally bind your own shaders when calling them, and leave the game shaders to draw as usual. This makes this PR safer to merge as "if I enable the flag, I get something done in the view 0, if not, I still get what I used to have, but starting from view 1". That's basically a way to "kill" and entire block of code with a piece of mind when merging :)

> As I said in a previous comment, FF8 is region blocked and not available for purchase/activation in my country, meaning that it is impossible for me to test if anything is broken in FF8. I would appreciate if you can test it for me, but if you are not available then maybe I can ask for help to someone else on the qhimm community.

Not sure where are you living at the moment, but you can fetch a copy from archive.org. Just saying :D

> If I understand correctly, this means reverting to a similar state as when I first submitted this PR, where a separate drawShadowMap() function was used, is that correct? Also may I ask why this change of mind?

See the answer above. That applies also for this case. Related to the previous state, yes, feel free to encapsulate all your code in a separate function, or better, see the answers below :)

> Bind the shadowmap texture. This could be done without touching the draw function with the Renderer::useTexture() function, is that correct?

No, that function is to bind texture attachments when drawing something. Not sure how the light texture works but if that makes your day, feel free to use it ( as it is! ) in your own function code :)

> Set the stencil buffer. There is not currently a way to set the stencil buffer so I was thinking to add some APIs to set the stencil function and reference value from outside the draw() function. However, the actual setting by calling the bgfx functions would be inside the draw() function in similar to how the blend state is set. Would that be ok?

Absolutely fine to have the Stencil part exactly like we have the blend state for eg. Feel free to implement an enum, a relative function to set which state you want and let the `draw()` API set it based on which is the current enum set. Remember to reset its state in the `reset()` function. That's fine by me!

> You say here to not change anything from the core shaders, but lighting calculations as well as fetching the shadowmap needs to be done here so I'm not sure I understand your request (most of it is already is in include files anyway).

Whatever calculation you need to do, consider doing it in your own new shader vert and frag files. Use those ones as your own greenfield and remember to attach them when drawing :)

> If I understand you correctly, this means reverting to something similar as when I was using separate draw functions such as drawShadowMap() but instead of implementing them inside the Renderer class, I create a different class and source files (lighting.h/cpp) that contains the lighting related draw functions as well as the code found in common.cpp. Is this correct?

Well, that's up to you if it makes it easier. For me it's just fine if you implement a new method inside the Renderer class which does not touch the `draw()` calls or any other API more than what we agreed above ( see stencil enum topic ). But if you think you can "use" the renderer class from the outside, feel free to do so, up to you! If you need to call bgfx APIs though, please consider using the Renderer class instead, I don't want bgfx code to be spreaded all over the files.

> Other than the lighting class files described above, I was thinking to separate the walkmesh extraction/extrusion functions into separate files (currently in field.h/cpp). Other than that I don't see what else would require new files. I would appreciate if you specify what else you want me to separate.

Fine by me, if that has nothing to do with bgfx APIs, feel free to add that logic in separate files! It would make it also easier for you in the future to fire more PRs as that will most probably be your own pillar :)

> Thank you again for your time reviewing this PR, it is very much appreciated!

Again thank you! Looking forward for this bunch of changes :)

src/gl/texture.cpp
julianxhokaxhiu: Glad you found your way out of this. Sorry but I'm a bit short of time in this period. Thank you, really appreciate it!

src/renderer.cpp
julianxhokaxhiu: > Thank you for understanding the need to reserve view id 0 for shadowmap rendering.

Thank you for taking the patience for still working on this. Really, really appreciate it!

> May I ask what made you change your mind on this? Also does this refer only to the shadowmap rendering shader or also the shader that rendered the walkmesh for the field shadows?

As we do reserve an entire view, it makes sense to use also separate shaders. As you will probably fire your own draw calls or whatever needed, you can totally bind your own shaders when calling them, and leave the game shaders to draw as usual. This makes this PR safer to merge as "if I enable the flag, I get something done in the view 0, if not, I still get what I used to have, but starting from view 1". That's basically a way to "kill" and entire block of code with a piece of mind when merging :)

> If I understand correctly, this means reverting to a similar state as when I first submitted this PR, where a separate drawShadowMap() function was used, is that correct? Also may I ask why this change of mind?

See the answer above. That applies also for this case. Related to the previous state, yes, feel free to encapsulate all your code in a separate function, or better, see the answers below :)

> Bind the shadowmap texture. This could be done without touching the draw function with the Renderer::useTexture() function, is that correct?

No, that function is to bind texture attachments when drawing something. Not sure how the light texture works but if that makes your day, feel free to use it ( as it is! ) in your own function code :)

> Set the stencil buffer. There is not currently a way to set the stencil buffer so I was thinking to add some APIs to set the stencil function and reference value from outside the draw() function. However, the actual setting by calling the bgfx functions would be inside the draw() function in similar to how the blend state is set. Would that be ok?

Absolutely fine to have the Stencil part exactly like we have the blend state for eg. Feel free to implement an enum, a relative function to set which state you want and let the `draw()` API set it based on which is the current enum set. Remember to reset its state in the `reset()` function. That's fine by me!

> You say here to not change anything from the core shaders, but lighting calculations as well as fetching the shadowmap needs to be done here so I'm not sure I understand your request (most of it is already is in include files anyway).

Whatever calculation you need to do, consider doing it in your own new shader vert and frag files. Use those ones as your own greenfield and remember to attach them when drawing :)

> If I understand you correctly, this means reverting to something similar as when I was using separate draw functions such as drawShadowMap() but instead of implementing them inside the Renderer class, I create a different class and source files (lighting.h/cpp) that contains the lighting related draw functions as well as the code found in common.cpp. Is this correct?

Well, that's up to you if it makes it easier. For me it's just fine if you implement a new method inside the Renderer class which does not touch the `draw()` calls or any other API more than what we agreed above ( see stencil enum topic ). But if you think you can "use" the renderer class from the outside, feel free to do so, up to you! If you need to call bgfx APIs though, please consider using the Renderer class instead, I don't want bgfx code to be spreaded all over the files.

> Other than the lighting class files described above, I was thinking to separate the walkmesh extraction/extrusion functions into separate files (currently in field.h/cpp). Other than that I don't see what else would require new files. I would appreciate if you specify what else you want me to separate.

Fine by me, if that has nothing to do with bgfx APIs, feel free to add that logic in separate files! It would make it also easier for you in the future to fire more PRs as that will most probably be your own pillar :)

> Thank you again for your time reviewing this PR, it is very much appreciated!

Again thank you! Looking forward for this bunch of changes :)

julianxhokaxhiu:
Just wanted to let you know you might want to rebase on top of the latest commit before you push :) Thanks!

src/renderer.cpp
CosmosXIII: I made some more changes as requested above.

- All lighting calculations are now done on separate new vert/frag shader files.
- Most of the lighting/shadow implementation has been moved to its own source files (lighting.h/cpp).
- Changes to the Renderer class have been reduced as much as possible.
- Added APIs to set the stencil buffer operations

Also, thankfully I was finally able to purchase a copy of FF8. I tested it with the changes of this PR and everything seems to work just fine. Let me know what you think!

src/renderer.cpp
```diff
@@ -556,11 +596,36 @@ void Renderer::init()
...
+
+    backendProgramHandles[RendererProgram::FIELD_SHADOW] = bgfx::createProgram(
+        getShader(vertexFieldShadowPath.c_str()),
+        getShader(fragmentFieldShadowPath.c_str()),
+        true
+    );
+
+    shadowMapProgramHandle = bgfx::createProgram(
```
julianxhokaxhiu, line 617: Can you please add this to the `backendProgramHandles` too? Thanks!

julianxhokaxhiu:
Impressive, we're close in getting this merged! Well done so far.

src/renderer.cpp
CosmosXIII: Done!

julianxhokaxhiu:
Awesome! Now please rebase your branch on top of the latest master so we have a clean state here. And let me know if you're intending to push some more commits, so I can test this branch on top of both FF7 and FF8 :) Cheers!

CosmosXIII:
I rebased on top of latest master as requested. I have quite a few ideas for improvements but I do not intend to make any more commits for this PR (unless you request more changes). Please let me know when you test it if you find any problem.

julianxhokaxhiu (approved):

julianxhokaxhiu:
I'll test this PR this weekend and if all works fine, I'll merge it. So far though the code itself looks good. Well done and thank you for taking the time to polish it!

CosmosXIII:
Thank you for reviewing an testing! Hope people like it!

Illynir:
I tested it too and no problems to report either. It is excellent. Big GG for this monstrous work which allows to add a not negligible visual "depth" in this mythical game. :)

In the options, I would have liked to see an option to manage the "intensity" of the shadows (from the darkest to the most transparent) and of course try to get around the limitation of one direction for the shadow, that it matches the scene where we are.

Anyway, excellent work again. :)

CosmosXIII:
@Illynir Thank you for testing and your feedback!

> In the options, I would have liked to see an option to manage the "intensity" of the shadows (from the darkest to the most transparent) and of course try to get around the limitation of one direction for the shadow, that it matches the scene where we are.

I think this can already be done by adjusting the balance of the direct light intensity and the ambient light intensity. If you are referring to field shadows there is actually a parameter called shadow occlusion that does exactly what you say. Please check it on the FFNx DevTools.

PR #240, opened 2021-05-09
Fixed big log file size and bad fps when lighting is enabled

CosmosXIII:
The big log file size and bad fps was caused because the log was being spammed with bgfx warnings due to setting the same uniforms and texture samplers within the same draw call from the drawToShadowMap() and draw() methods. To fix this, I separated the call to setCommonUniforms() and the texture sampling binding into a new drawWithoutLighting() method. This method then calls the draw() function. Therefore all calls to the draw() function other than the lighting one have been replaced with drawWithoutLighting() but the code flow is exactly the same. To prevent calling the draw() function directly from outside the Renderer class I made it private.

### DIFF
--- a/src/ff7/graphics.cpp
+++ b/src/ff7/graphics.cpp
@@ -359,7 +359,7 @@ void draw_single_triangle(struct nvertex *vertices)
 
 	newRenderer.isTLVertex(true);
 
-	newRenderer.draw();
+	newRenderer.drawWithoutLighting();
 }
 
 void sub_6B2720(struct indexed_primitive *ip)
--- a/src/gl/gl.cpp
+++ b/src/gl/gl.cpp
@@ -113,7 +113,7 @@ void gl_draw_movie_quad_common(uint32_t width, uint32_t height)
 	newRenderer.isTLVertex(true);
 	newRenderer.doTextureFiltering(current_state.texture_filter);
 
-	newRenderer.draw();
+	newRenderer.drawWithoutLighting();
 }
 
 // draw movie frame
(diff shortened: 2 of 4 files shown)


julianxhokaxhiu:
I see that this PR impacts quite a lot the basic Renderer code. I'd need some time to check better the code path and especially if this breaks FF8.

May I ask why do we have to rename the method? The idea is that we want to abstract the class, if you remember from the previous PR. So having a method named "drawWithoutLightning" is far from that.

I'd rather prefer to encapsulate the logic of "light" in its own methods, so if there's a uniform attachment issue, let's find another way to attach those only when a light draw call will be fired. Which by the way, it's completely internal to the class so you might be able to know that.

I'll see if I can find another way to fix it.

CosmosXIII:
I had the feeling that it would not be good like this...
The draw() function always sets the common uniforms and texture samplers but these also need to be set when in drawToShadowMap(). That is why it is being called twice. I can think of two other ways to fix it:
- Have a condition inside the draw() function to only set the uniforms/samplers when lighting is disabled or the postprocessing program is used
- Not calling the draw() function from the drawWithLighting() method. This would require copying most of the implementation in the draw() into the drawWithLighting method. I did not do this at first because I thought it would be better to share the code if possible but if you prefer it to be completely separated this may be best.

julianxhokaxhiu:
There's a third and better way actually. I'm working on it right now, and is a simple "bind" check for uniforms, which I didn't implement at the time as the flow was quite linear and I never had cases like this one on the radar. Not a big deal though, it's time to put that part of code in defensive mode :)

CosmosXIII:
Ok, that seems good but please take into account that other than the uniforms a similar warning also occurs when calling bgfx::setTexture twice on the same draw call. Maybe wrap it so that you can check if its already been set :)

julianxhokaxhiu:
Allright, I fixed this completely in another way, therefore I'll close the PR. Thanks for bringing this up though! Feel free to checkout [this commit](https://github.com/julianxhokaxhiu/FFNx/commit/5ce080ca6a337153d316c10f7127aacd103f3f0d) for more details!

PR #245, opened 2021-05-12
[FF7] Research about fr_e field and shadow problem

LaZar00:
Hi,

after making some changes to the field fr_e, I have arrived to one conclusion (maybe wrong, though).

For understand this, I need to clarify some things:

1.- fr_e field has 2 layers:
     - Layer 0 (type 2, direct color texture) that is the main background.
     - Layer 3 (type 1, palettized texture) that is the shadow layer (Diamond Weapon shadow).
2.- Layer 3 uses blending (over Layer 0, of course), with the Type of Transparency (or blending mode) = 2.
3.- With some tests I have done with the fr_e field with the imported shadow of PSX into PC, I have determined (95% sure):
     - FF7.EXE DO NOT uses blending mode = 2.
     - It seems that always uses blending mode = 1 in any case.
     - I say this because I have tested changing the blending mode to 0, 1 and 3 for Layer 3 of fr_3 (in the tiles data), and the result is the same.

So, I would ask some things to research this a bit more and confirm this theory:

a.) Is possible that someone debug FFNx driver using fr_e field and check if it uses blending mode = 2, really?
b.) I know that fr_e has not a correct Layer 3 (image of a shadow -Diamond Weapon shadow-). But as I have said I have the field with the imported shadow of PSX. If anyone wants to make some tests, please let me know. But I would like to verify first really a.).
c.) If FFNx does not uses finally blending mode = 2 for Layer 3 (or in any other case), could it be possible to implement?

I leave this here for the posterity. So, let's see if we can solve the fr_e broken field.


julianxhokaxhiu:
Please convert this into an issue :) Thank you.

PR #246, opened 2021-05-22
Optimized normal calculation

CosmosXIII:
I removed the double nested for loop so it should be faster now.

### DIFF
--- a/src/gl/gl.cpp
+++ b/src/gl/gl.cpp
@@ -181,50 +181,38 @@ void gl_draw_with_lighting(struct indexed_primitive *ip, struct polygon_data *po
 	if (normaldata == nullptr)
 	{
 		static std::vector<struct point3d> normals;
-		static std::vector<int> adjacentTriIndexes;
-		struct point3d normal, e12, e13, triNormal;
+		static point3d zero = { 0.0f, 0.0f, 0.0f };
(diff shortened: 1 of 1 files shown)


src/gl/gl.cpp
```diff
@@ -181,50 +181,42 @@ void gl_draw_with_lighting(struct indexed_primitive *ip, struct polygon_data *po
...
-		normals.shrink_to_fit();
+		normals.resize(vertexCount);
+		std::fill(normals.begin(), normals.end(), zero);
 
 		// Calculate vertex normals by averaging adjacent triangle normals
 		// Vertex normals are calculated here because battle models dont seem to include normals
-		for (uint32_t idx = 0; idx < ip->vertexcount; idx++)
+		for (uint32_t idx = 0; idx < indexCount; idx+=3)
+		{
+			e12 = e13 = triNormal = {0.0f, 0.0f, 0.0f};
```
julianxhokaxhiu, line 198: I'd use `zero` as well instead of `{0.0f, 0.0f, 0.0f}`

CosmosXIII: I changed it to `zero`.

julianxhokaxhiu (approved):

julianxhokaxhiu:
Well done, thank you!

PR #251, opened 2021-05-24
Added support for extended PBR textures

CosmosXIII:
Added support for extended PBR textures (normal map, roughness, metalness, AO). This is achieved by extending the textures vertically to three times their original height. The first region contains the original texture, the second contains the normal map (RGB channels, alpha unused), the third region contains the other parameters (R: roughness, G: metalness, B: AO, A unused). These extended textures will be provided in IRO mods.

The core FFNx was slightly changed to make it work with the extended textures mods even when lighting is off (otherwise the uvs would be broken and cause the extended textures to show as colors).

### DIFF
--- a/misc/FFNx.frag
+++ b/misc/FFNx.frag
@@ -19,7 +19,7 @@
 //    GNU General Public License for more details.                          //
 /****************************************************************************/
 
-$input v_color0, v_texcoord0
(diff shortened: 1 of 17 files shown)


julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you and well done! Great work so far :)

PR #252, opened 2021-05-27
Fixed incorrect processing of PBR textures

CosmosXIII:
This fixes a mistake I made when processing the PBR textures in the lighting shader. I also removed unnecessary code I used for debug.


julianxhokaxhiu:
Thank you!

PR #259, opened 2021-06-27
Fixed field shadows being drawn over 3D models

CosmosXIII:
Fixed the field shadows being drawn over the 3d models. I also removed all the stencil buffer related API since those are not needed anymore.

### DIFF
--- a/src/gl.h
+++ b/src/gl.h
@@ -28,6 +28,15 @@
 #define LVERTEX 2
 #define TLVERTEX 3
 
+enum DrawOrder
+{
+	DRAW_ORDER_0 = 0,
+	DRAW_ORDER_1,
+	DRAW_ORDER_2,
+
+	DRAW_ORDER_COUNT
+};
+
 struct driver_state
 {
 	struct texture_set *texture_set;
@@ -55,6 +64,7 @@ struct deferred_draw
 	uint32_t vertextype;
 	uint32_t vertexcount;
 	uint32_t count;
+	uint32_t draworder;
 	struct nvertex* vertices;
 	struct point3d* normals;
 	WORD *indices;
@@ -93,7 +103,8 @@ void gl_save_state(struct driver_state *dest);
 void gl_load_state(struct driver_state *src);
 uint32_t gl_defer_draw(uint32_t primitivetype, uint32_t vertextype, struct nvertex* vertices, struct point3d* normals, uint32_t vertexcount, WORD* indices, uint32_t count, struct boundingbox* boundingbox, uint32_t clip, uint32_t mipmap);
 uint32_t gl_defer_sorted_draw(uint32_t primitivetype, uint32_t vertextype, struct nvertex *vertices, uint32_t vertexcount, WORD *indices, uint32_t count, uint32_t clip, uint32_t mipmap);
-void gl_draw_deferred(bool isDrawOpaqueOnly);
+void gl_draw_deferred(DrawOrder draworder);
+void gl_set_projection_viewport_matrices();
 struct boundingbox calculateSceneAabb();
 void gl_draw_sorted_deferred();
 void gl_check_deferred(struct texture_set *texture_set);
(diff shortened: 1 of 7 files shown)


src/gl/gl.cpp
```diff
@@ -246,6 +246,12 @@ void gl_draw_indexed_primitive(uint32_t primitivetype, uint32_t vertextype, stru
 
 	newRenderer.setInterpolationQualifier(current_state.shademode ? RendererInterpolationQualifier::SMOOTH : RendererInterpolationQualifier::FLAT);
 
+	if (vertextype != TLVERTEX)
+	{
+		newRenderer.setD3DProjection(&current_state.d3dprojection_matrix);
+		newRenderer.setD3DViweport(&d3dviewport_matrix);
+	}
+
```
julianxhokaxhiu, line 254: May I ask why did you move this block? Is this really required?

src/gl.h
```diff
@@ -93,7 +103,7 @@ void gl_save_state(struct driver_state *dest);
 void gl_load_state(struct driver_state *src);
 uint32_t gl_defer_draw(uint32_t primitivetype, uint32_t vertextype, struct nvertex* vertices, struct point3d* normals, uint32_t vertexcount, WORD* indices, uint32_t count, struct boundingbox* boundingbox, uint32_t clip, uint32_t mipmap);
 uint32_t gl_defer_sorted_draw(uint32_t primitivetype, uint32_t vertextype, struct nvertex *vertices, uint32_t vertexcount, WORD *indices, uint32_t count, uint32_t clip, uint32_t mipmap);
-void gl_draw_deferred(bool isDrawOpaqueOnly);
+void gl_draw_deferred(int draworder);
```
julianxhokaxhiu, line 106: Use the Enum type directly if you expect to be always that one

src/gl/deferred.cpp
```diff
@@ -291,7 +296,7 @@ uint32_t gl_defer_sorted_draw(uint32_t primitivetype, uint32_t vertextype, struc
 }
 
 // draw deferred models
-void gl_draw_deferred(bool isDrawOpaqueOnly)
+void gl_draw_deferred(int draworder)
```
julianxhokaxhiu, line 299: Same as stated above

julianxhokaxhiu (changes requested):
Overall looks good, there are minor things to clarify and to amend. Great job so far!

src/gl.h
CosmosXIII: Changet it to use the Enum as parameter.

src/gl/deferred.cpp
CosmosXIII: Changed it to use the Enum as parameter.

src/gl/gl.cpp
CosmosXIII: I moved this code to guarantee that the projection and viewport matrices are set before drawing the field shadows. Until now I was drawing other 3d models such as the characters before the field shadows so these matrices were set before drawing the shadows. However, with the changes of this PR, the field shadows (walkmesh) are the first 3d object to be drawn in the frame. By moving this block of code before the draw call deferring I was able to guarantee that these matrices are set before any drawing. Without this change the matrices used when drawing the field shadows have a one frame delay.

As far as I know these matrices are common for all 3d objects so it should be enough to set once at the beginning of the frame. However, I did it this way because it was the method with the less amount of changes required. If this change is not desired I could work around it by adding a new method that sets the matrices for the first 3d object before any drawing. Let me know what you think and thanks for the review as always!

julianxhokaxhiu: To be honest the only part that concerns me is that this code is in common with FF8 which has a tons of tricks in place put here and there to guarantee perfect rendering. This part of the code is one of those and I have no idea if this code change might have an impact ( as you notice there's an if case before this block originally which might return and set basically nothing. Moving this block before that if, now might bring some implications i might not be aware.

If you can find a way to set those in your lighting code class I would appreciate. Thank you!

CosmosXIII: I implemented an alternative way to set the matrices using a new method that is only called when lighting is enabled.

julianxhokaxhiu: Amazing, thank you :)

julianxhokaxhiu (approved):

PR #274, opened 2021-08-29
FF7: Implemented analogue controls in fields

CosmosXIII:
I implemented analogue controls so that the player can move in whatever direction you move the joystick.
I used the "Control Direction" parameter in the flevel: http://wiki.ffrtt.ru/index.php/FF7/Field/Triggers
This parameter defines for each field the orientation the player moves. This PR achieves analogue controls by overriding this parameter every frame so that the player moves in 256 directions instead of 8.


julianxhokaxhiu (approved):
Smart :) I like it, well done!

julianxhokaxhiu:
And thank you!

PR #278, opened 2021-09-12
Implemented linear space and image based lighting (IBL)

CosmosXIII:
This PR contains the following changes:
- Changed lighting so that it is done correctly in linear space instead of gamma space (SRGB). This requires a change to a RGBA16 framebuffer. Final SRGB conversion is done by adding the SRGB flag to the swap buffer. Textures that represent colors are set with the SRGB flag so it is converted to linear space by the GPU.
- Added functionality to load external environment maps for image base lighting (IBL).
- Added specular pbr parameter to control specular reflection of non-metallic materials
- Various shader code fixes

### DIFF
--- a/src/cfg.cpp
+++ b/src/cfg.cpp
@@ -41,6 +41,7 @@ std::string external_voice_path;
 std::vector<std::string> external_voice_ext;
 std::string external_ambient_path;
 std::vector<std::string> external_ambient_ext;
+std::string external_ibl_path;
 bool enable_voice_music_fade;
 long external_voice_music_fade_volume;
 bool save_textures;
@@ -157,6 +158,7 @@ void read_cfg()
 	external_voice_music_fade_volume = config["external_voice_music_fade_volume"].value_or(25);
 	external_ambient_path = config["external_ambient_path"].value_or("");
 	external_ambient_ext = get_string_or_array_of_strings(config["external_ambient_ext"]);
+	external_ibl_path = config["external_ibl_path"].value_or("");
 	save_textures = config["save_textures"].value_or(false);
 	trace_all = config["trace_all"].value_or(false);
 	trace_renderer = config["trace_renderer"].value_or(false);
@@ -346,6 +348,10 @@ void read_cfg()
 	// EXTERNAL AMBIENT EXTENSION
 	if (external_ambient_ext.empty() || external_ambient_ext.front().empty())
 		external_ambient_ext = std::vector<std::string>(1, "ogg");
+
+	// EXTERNAL IBL PATH
+	if (external_ibl_path.empty())
+		external_ibl_path = "lighting/ibl";
 
 	// MOD PATH
 	if (mod_path.empty())
--- a/src/lighting.cpp
+++ b/src/lighting.cpp
@@ -17,6 +17,9 @@
 #include "renderer.h"
 #include "macro.h"
 #include "ff7.h"
+#include "ff8.h"
+#include "field.h"
+#include "cfg.h"
 
 Lighting lighting;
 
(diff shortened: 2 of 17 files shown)


src/cfg.cpp
```diff
@@ -347,6 +349,10 @@ void read_cfg()
 	if (external_ambient_ext.empty() || external_ambient_ext.front().empty())
 		external_ambient_ext = std::vector<std::string>(1, "ogg");
 
+	// EXTERNAL IBL PATH
+	if (external_ibl_path.empty())
+		external_ibl_path = "lighting/IBL";
```
julianxhokaxhiu, line 354: For the sake of convention, can you please make the path all lowercase? Thank you :)

src/renderer.cpp
```diff
@@ -742,6 +744,38 @@ void Renderer::prepareShadowMap()
...
+}
+
+void Renderer::prepareEnvBrdf()
+{
+    static char fullpath[MAX_PATH];
+
+    if (bgfx::isValid(envBrdfTexture))
+        bgfx::destroy(envBrdfTexture);
+
+    sprintf(fullpath, "%s/%s/%s.%s", basedir, external_ibl_path.c_str(), "envBrdf", "dds");
```
julianxhokaxhiu, line 773: If "envBrdf" and "dds" is a known constant, just put it in the string already, no need to pass them as an argument :)

src/renderer.h
```diff
@@ -255,6 +258,10 @@ class Renderer {
     bgfx::TextureHandle shadowMapTexture = BGFX_INVALID_HANDLE;
     bgfx::FrameBufferHandle shadowMapFrameBuffer = BGFX_INVALID_HANDLE;
 
+    bgfx::TextureHandle specularIblTexture = BGFX_INVALID_HANDLE;
+    bgfx::TextureHandle diffuseIblTexture = BGFX_INVALID_HANDLE;
+    bgfx::TextureHandle envBrdfTexture = BGFX_INVALID_HANDLE;
```
julianxhokaxhiu, line 263: Are those textures part of a single "draw" call logic on the GPU? Can we put them for eg. inside of `internalState.texHandlers` ?

CosmosXIII: No, these textures are not part of a single draw call, they are global (common to all draw calls).

src/renderer.cpp
CosmosXIII: I merged it into "envBrdf.dds".

src/cfg.cpp
CosmosXIII: Sorry, my mistake, I changed it to lowercase.

src/lighting.cpp
```diff
@@ -103,6 +106,50 @@ void Lighting::updateLightMatrices(struct boundingbox* sceneAabb)
...
+	switch (mode->driver_mode)
+	{
+	case MODE_BATTLE:
+		if (mode->driver_mode != prev_mode || last_battle_id != ff7_externals.modules_global_object->battle_id)
+		{
+			last_battle_id = ff7_externals.modules_global_object->battle_id;
+
+			sprintf(filename, "bat_%d", last_battle_id);
+			sprintf(specularFullpath, "%s/%s/%s%s", basedir, external_ibl_path.c_str(), filename, "_s.dds");
+			sprintf(diffuseFullpath, "%s/%s/%s%s", basedir, external_ibl_path.c_str(), filename, "_d.dds");
```
julianxhokaxhiu, line 127: Same goes for those two. Sorry I missed them.

src/lighting.cpp
```diff
@@ -103,6 +106,50 @@ void Lighting::updateLightMatrices(struct boundingbox* sceneAabb)
...
+		}
+		break;
+	case MODE_FIELD:
+		if (mode->driver_mode != prev_mode || last_field_id != *ff7_externals.field_id)
+		{
+			last_field_id = *ff7_externals.field_id;
+
+			sprintf(filename, "field_%d", last_field_id);		
+			sprintf(specularFullpath, "%s/%s/%s%s", basedir, external_ibl_path.c_str(), filename, "_s.dds");
+			sprintf(diffuseFullpath, "%s/%s/%s%s", basedir, external_ibl_path.c_str(), filename, "_d.dds");
```
julianxhokaxhiu, line 140: Here as well

src/renderer.cpp
```diff
@@ -742,6 +744,38 @@ void Renderer::prepareShadowMap()
...
+}
+
+void Renderer::prepareEnvBrdf()
+{
+    static char fullpath[MAX_PATH];
+
+    if (bgfx::isValid(envBrdfTexture))
+        bgfx::destroy(envBrdfTexture);
+
+    sprintf(fullpath, "%s/%s/%s", basedir, external_ibl_path.c_str(), "envBrdf.dds");
```
julianxhokaxhiu, line 773: Add the last param directly in the format string as well.

CosmosXIII: Ups, sorry, I see now what you mean. I removed the constant string param.

src/lighting.cpp
CosmosXIII: Fixed this aswell.

src/lighting.cpp
CosmosXIII: Fixed this as well.

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #282, opened 2021-09-20
Fixed transparency by performing blending in SRGB space

CosmosXIII:
Fixed transparency by performing blending in SRGB space. Lighting is still being done in linear space as I wanted, only blending is done in SRGB so that it looks exactly the same as before.


julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #300, opened 2021-11-03
FF7Achievement: Fix and refactor FF7 achievements

tangtang95:
- Fix Gold Chocobo achievement by checking it when the slots are filled through some function under update_field_entities
 - Fix Yuffie and Vincent achievement by checking it when the regular mask of the two character is modified
 - Add FF7 variables to avoid unlocking some achievement randomly (unlock only when the condition actually switch from false to true)
 - Move ff7_load_save_file function to misc.cpp


julianxhokaxhiu:
LGTM

julianxhokaxhiu (approved):

PR #303, opened 2021-11-05
FF7Achievement: Fix and improve Steam achievements

tangtang95:
- Fix cait sith last limit achievement by checking the number of kills
 - Fix bug in defeating weapons achievement by moving the check to the battle-end menu since fanfare sound is not played always
 - Fix bug in mastered all materia in case the player takes one of the unmasterable materia
 - Fix game progress achievement by unlocking them when the respective FMV movie plays
 - Improve 1st limit break achievement by checking it during battle (these also unlock when 1-2 limit breaks are used)

### DIFF
--- a/src/achievement.cpp
+++ b/src/achievement.cpp
@@ -26,6 +26,8 @@
 
 #include "achievement.h"
 
+using std::string;
+
 SteamAchievementsFF7 g_FF7SteamAchievements;
 SteamAchievementsFF8 g_FF8SteamAchievements;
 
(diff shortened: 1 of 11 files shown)


tangtang95:
I have already pushed the new rebased commit. Is there some other issue with the build?

julianxhokaxhiu:
> I have already pushed the new rebased commit. Is there some other issue with the build?

I stopped it because otherwise it would clash with the cache on NuGet. I'll re-run it again as soon as the main build finishes.

tangtang95:
Got it, I thought there was some issue.

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you :)

PR #304, opened 2021-11-09
FF7Achievement: Release

tangtang95:
- Allow unlocking last limit break achievement even when the item quantity is greater than 1
 - Refactor FF7 steam achievement using constexpr and const reference parameters when possible
 - Reorder FF7 achievement list to avoid using a list for swapping indices
 - Add show achievement progress for materia overlord achievement
 - Add debug flag in cfg for the future release of steam achievements

### DIFF
--- a/src/achievement.cpp
+++ b/src/achievement.cpp
@@ -25,6 +25,7 @@
 #include <steamworkssdk/isteamuser.h>
 
 #include "achievement.h"
+#include <numeric>
 
 using std::string;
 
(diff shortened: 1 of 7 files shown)


src/achievement.cpp
```diff
@@ -65,19 +66,19 @@ bool SteamManager::setAchievement(int achID)
...
+        return SteamUserStats()->IndicateAchievementProgress(this->getStringAchievementID(achID), 99, 100);
+        // if(steam_achievements_debug_mode)
+        // {
+        //     return SteamUserStats()->IndicateAchievementProgress(this->getStringAchievementID(achID), 99, 100);
+        // } 
+        // else
+        // {
+        //     SteamUserStats()->SetAchievement(this->getStringAchievementID(achID));
+        //     return SteamUserStats()->StoreStats();
+        // }
```
julianxhokaxhiu, line 81: Better cleanup this, if you need that piece of code you can look up at the git history of this file :) Thanks

//EDIT: I also was wondering, if the flag still makes sense as this code is commented.

julianxhokaxhiu (changes requested):

src/achievement.cpp
tangtang95: I know, then I will remove the flag and the comment for now.

tangtang95:
I wanted to share somewhere this, so, we could have some more feedback. Do you think qhimm forum is a good place?

tangtang95:
> I wanted to share somewhere this, so, we could have some more feedback. Do you think qhimm forum is a good place?

Nevermind, not really into managing a post in a forum ahahaha. I will put information on this PR. Then if you want, you can put it somewhere else

julianxhokaxhiu:
I noticed you have it flagged as "TEST PHASE" but there's no real test or release phase as part of your code ended up in being already in 1.10, therefore consider it already release. In this case, just let your code flow based on flags, therefore the commented section should be uncommented and should behave either in debug mode, or not, based simply on you put in the `FFNx.toml`.

Regarding the announcement, it has already been done in Discord at least and it will be more globally spread and used on the various mod managers, when their relative authors will feel it's ready to be integrated. Until then, it's left on end users to enable it, at their own "risk". What we can guarantee on our end is that the code does what is supposed to :)

Because of this, please make sure every line code in this PR is effective code and ready for massive usage, as much as you can of course. Nothing is perfect and if a fix is still required, PRs are open and so are releases.

tangtang95:
Well, in 1.10, I see the steam achievements as being in the test phase because they still do not unlock the achievement on steam (they just show a progress achievement in order to avoid "ruining" the steam account). I think that the action of unlocking the achievement is almost an irreversible operation since you cannot remove the achievement. That's why I thought it was necessary to have a code tested thoroughly.

Overall, I have tested it in normal situations, but I still do not have much feedback from others. I think that it is ready to be completely released (i.e.truly unlock the achievement on steam)

tangtang95:
Ok, I decided to activate completely the steam achievements. From now on, they will truly unlock the achievement on Steam.

julianxhokaxhiu (approved):
LGTM

julianxhokaxhiu:
Thank you :)

PR #306, opened 2021-11-12
FF7Achievement: Fix game progress achievement

tangtang95:
This improves the end part 1 and end part 2 achievement since they will show after the menu screen of END PART (in which it allows you to save the game). For the end of game achievement, it has been moved to the end of "ending2" movie because the last movie ("ending3") is just before the game exiting which might avoid activating the achievement since it is not a sync call (unlocking the achievement at skip movies has been removed because the "ending2" movie cannot be skipped)
 - Unlock end part 1 and 2 achievement through cd check after seen the respective last movie of each part
 - Unlock end of game achievement on ending2 movie which is the one before the list of credits


julianxhokaxhiu (approved):
LGTM

julianxhokaxhiu:
Thanks!

PR #311, opened 2021-11-14
FF7Achievement: Fix debug flag bug and avoid steam_appid.txt for steam edition

tangtang95:
I've seen that steam_appid.txt is useless if run as steam edition. I also thought to activate steam achievements on steam edition directly (w/o looking at the flag in toml file), but maybe it is too soon.


julianxhokaxhiu:
> but maybe it is too soon.

Why is too soon? I think it's totally correct behavior. What is currently blocking you?

tangtang95:
Yeah, maybe it's fine. I just wanted to have some other feedback.

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you :)

Issue #31, opened 2020-05-15
[ Common ] Full C++ codebase


julianxhokaxhiu:
The driver at the current status is written with a mix of C and C++ code.

Ideally we need to move forward the code to be fully C++, by deprecating malloc/calloc logics, as well as replacing `char[]` with `std::string`.

Those are just hints for the direction, so more topics have to be explored.

The idea is make sure this code is compatible and can be built with both MSVC and MinGW, with no extra effort.

PR #312, opened 2021-11-15
[ IDEA ] Refactor externals functions

tangtang95:
I thought of doing something for all the double declaration and type casting of externals. I just put an example with load_save_file function and another one. These are the advantages and disadvantages that I thought about.
- \+ Avoid all the type casting
- \+ No change in ff7_data.h about get_relative_call() or others.
- \+ The type checking of the compiler still works for the external functions arguments (even the IDE sees it)
-  \- Might increase significantly the compilation time due to the templateù

What do you think? Is it useful? Worth it?

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -250,6 +250,27 @@ struct common_externals
 	uint32_t update_entities_call;
 };
 
+template <typename T>
+struct external_fn;
+
+template<typename R, typename... Args>
+struct external_fn<R(Args...)>{
+	uint32_t address;
+
+	operator uint32_t() const {
+		return address;
+	}
+
+	external_fn<R(Args...)>& operator=(uint32_t address){
+		this->address = address;
+		return *this;
+	}
+
+	R call(Args... args) const {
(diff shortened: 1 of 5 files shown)


julianxhokaxhiu:
I'll try to take a look this evening and will let you know :) Thanks.

Anyway we already have that pattern for calling external functions, just not using C++ templates. This PR may fit with this issue instead #31 but at the current state I wouldn't consider merging it. Maybe we could keep it for a future 2.0 release.

tangtang95:
Yeah, the pattern is the same. It is just a way to avoid all the type checking and have the declaration of the function return type and argument only in one place. At first I thought it avoided the declaration of the function and the caller address (e.g. play_battle_end_music and play_battle_music_win_call), but this does not happen. Maybe, it is possible to add a std::vector inside the struct to keep these.

julianxhokaxhiu:
Keep the branch on your own side and let's consider this PR for a future FFNX 2.0 release, when and if it will happen.

Thank you!

PR #313, opened 2021-11-17
FF7Achievement: Refactor and fix 1st limit break achievement

### LINKED ISSUE #310: [ FF7 ] Limit 1-1 Steam achievement unlocked also by Limit 1-2
As of now, the **1st limit break achievement is unlocked also by limit 1-2** (for example, Cross Slash unlocks Cloud's Braver achievement). Obviously, this is not a problem in a normal FF7 run. 

Basically, right now, the achievement is unlocked whenever the limit counter in battle is increased by 1 (in this function `0x5C930F`) and the current limit level is set to 1. 

Note that a previous method was using the number of limit 1-1 (https://wiki.ffrtt.ru/index.php?title=FF7/Savemap#Character_Record)  (updated at the end of battle with the counter I specified before), but in reality this contains the number of times limit 1 is used. So, the same issue persists.

tangtang95:
- Refactor built-in array into std array
 - Refactor achievement classes into real classes with constructor
 - Refactor ff7 achievement  externals names into more meaningful ones
 - Refactor simple pointer into std unique pointers
 - **Fix ff7_get_inserted_cd function not assigning new inserted cd**
 - **Fix 1st limit break achievement by unlocking it when setting the action id of the actor in battle (Fix #310 )**


src/common.cpp
```diff
@@ -2299,34 +2299,40 @@ uint32_t ff7_get_inserted_cd(void) {
...
+			ffnx_trace("inserted CD: %d, requiredCD: %d\n", insertedCD, requiredCD);
+
+		if(requiredCD == insertedCD + 1)
+			g_FF7SteamAchievements->unlockGameProgressAchievement();
+	}
 
 	if (requiredCD > 0 && requiredCD <= 3) ret = requiredCD;
 
-	insertedCD = ret;
+	*insertedCDAddress = ret;
```
julianxhokaxhiu, line 2335: I get what you're trying to do it here but I'd like to see the patch done differently.

Move the CD declaration variable in the `ff7_externals` and use it to read/write into the location. Also use a proper name for the variable like `insertedCD`, even if it's a pointer to an address, like most of the declarations are in this code anyway. No need to suffix it with `Address`.

Thanks :)

julianxhokaxhiu (changes requested):

src/common.cpp
tangtang95: Now, it should be fine

tangtang95:
I forgot to add the hashtag of the issue I fixed in the commit. I will pushed a new one.

julianxhokaxhiu (approved):
LGTM. Thank you :)

PR #314, opened 2021-11-19
GameHacks: Add auto attack hack for FF7 battles (L2 + R3 or CTRL + A)

tangtang95:
Add a new game hack for FF7 that enables auto attack mode. Each character will attack the 1st enemy. It can be activated outside battle and in-battle; even during the battle menu selection.

### DIFF
--- a/src/ff7/menu.cpp
+++ b/src/ff7/menu.cpp
@@ -23,6 +23,7 @@
 #include "../ff7.h"
 #include "../log.h"
 #include "../achievement.h"
+#include "../gamehacks.h"
 
 void ff7_menu_battle_end_sub_6C9543()
 {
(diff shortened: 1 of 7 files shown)


src/ff7/menu.cpp
```diff
@@ -133,3 +134,27 @@ uint32_t ff7_menu_decrease_item_quantity(uint32_t item_used)
...
+void dispatchAttackCommand(){
+    *ff7_externals.issued_command_id = 0x01;
+    *ff7_externals.issued_action_target_type = 0;
+    *ff7_externals.issued_action_target_index = 4;
+    ((void(*)())ff7_externals.dispatch_chosen_battle_action)();
+}
+
+void ff7_battle_menu_sub_6DB0EE(){
+    ((void(*)())ff7_externals.battle_sub_6DB0EE)();
+    if(gamehacks.isAutoAttack() && (*ff7_externals.battle_menu_state >= 0 && *ff7_externals.battle_menu_state < 19)){
```
tangtang95, line 147: Don't really know if the battle menu state from 0 to 19 (excluded) covers all possible battle menu of the active character. I took this value from SisterRay (https://github.com/FFT-Hackers/SisterRay/blob/master/include/SisterRay/constants.h#L158), and confirmed the 19 in FF7.exe code

tangtang95:
I have seen that double cut materia does not work. Don't know if it is better to have it or not. Do you know any other materia that changes the mechanism? Maybe long range materia...

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #315, opened 2021-11-27
Support of correct camera speed at 60 FPS

tangtang95:
This is what I have done on the camera 60 fps support. It is kinda a mess, but I have seen that it has the correct speed in the battle intro camera movement and also other things which I don't remember or cannot distinguish since the animations are quicker.

I have commented two lines in the ff7_opengl.cpp since they cause crashes, but they are in the original hext patch. While other two multiplication are missing because they cannot be done with this technique.

### DIFF
--- a/misc/FFNx.toml
+++ b/misc/FFNx.toml
@@ -408,6 +408,10 @@ trace_gamepad = false
 #~~~~~~~~~~~~~~~~~~~~~~~~~~~
 trace_achievement = false
 
+# trace_battle_camera - Dump in the logs only APIs that has to do with battle camera
+#~~~~~~~~~~~~~~~~~~~~~~~~~~~
+trace_battle_camera = false
+
 # vertex_log - Dump in the logs current engine vertex data being passed to the GPU for drawing
 #~~~~~~~~~~~~~~~~~~~~~~~~~~~
 vertex_log = false
--- a/src/cfg.cpp
+++ b/src/cfg.cpp
@@ -60,6 +60,7 @@ bool trace_voice;
 bool trace_ambient;
 bool trace_gamepad;
 bool trace_achievement;
+bool trace_battle_camera;
 bool vertex_log;
 bool uniform_log;
 bool show_renderer_backend;
@@ -178,6 +179,7 @@ void read_cfg()
 	trace_ambient = config["trace_ambient"].value_or(false);
 	trace_gamepad = config["trace_gamepad"].value_or(false);
 	trace_achievement = config["trace_achievement"].value_or(false);
+	trace_battle_camera = config["trace_battle_camera"].value_or(false);
 	vertex_log = config["vertex_log"].value_or(false);
 	uniform_log = config["uniform_log"].value_or(false);
 	show_renderer_backend = config["show_renderer_backend"].value_or(true);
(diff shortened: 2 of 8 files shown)


src/ff7/misc.cpp
```diff
@@ -552,6 +552,89 @@ int ff7_field_load_models_atoi(const char* str)
...
+
+void ff7_execute_camera_functions()
+{
+	uint16_t& fn_index = *ff7_externals.camera_index;
+	for(fn_index = 0; fn_index < 16; fn_index++){
+		if(ff7_externals.camera_fn_array[fn_index] != 0){
+			
+			if(new_camera_fn[fn_index])
+			{
+				ffnx_trace(" %s - function started: 0x%x\n", __func__, ff7_externals.camera_fn_array[fn_index]);
```
julianxhokaxhiu, line 596: Please add a new flag `trace_camera` and put these traces behind that flag.

src/ff7/misc.cpp
```diff
@@ -552,6 +552,89 @@ int ff7_field_load_models_atoi(const char* str)
...
+				ff7_externals.battle_camera_data[fn_index].num_camera_steps *= 2;
+				new_camera_fn[fn_index] = false;
+			}
+			else if(post_2_camera_fn[fn_index])
+			{
+				ffnx_trace(" %s - function started: 0x%x\n", __func__, ff7_externals.camera_fn_array[fn_index]);
```
julianxhokaxhiu, line 602: Same as above.

src/ff7/misc.cpp
```diff
@@ -552,6 +552,89 @@ int ff7_field_load_models_atoi(const char* str)
...
+				ff7_externals.battle_camera_data[fn_index].field_4 *= 2;
+				post_2_camera_fn[fn_index] = false;
+			}
+			else if(post_3_camera_fn[fn_index])
+			{
+				ffnx_trace(" %s - function started: 0x%x\n", __func__, ff7_externals.camera_fn_array[fn_index]);
```
julianxhokaxhiu, line 608: Same as above.

tangtang95:
This should patch only the camera movement part (2X camera in the hext patch file). It probably needs some other patches.

julianxhokaxhiu:
Is this patch ready or would you need some more time to work on it?

tangtang95:
> Is this patch ready or would you need some more time to work on it?

Half ready, maybe I should check if it does the same thing as the hext patch. The other thing is that I saw many crashes.

julianxhokaxhiu:
Let's keep it as a draft for now. When you'll feel this is mature on your end, you can mark it as ready and I'll be happy to review.

julianxhokaxhiu:
Amazing work :) Do you think we can add 30FPS too in another PR?

tangtang95:
I don't really know if this hext patch is for 60 fps or 30 fps battle because in FF7 original, there is a 2x difference between battle animation (15 fps) and camera animation (30 fps), so if there is nothing else, in theory, this is the correct camera animation for 30 fps battles. I have timed the battle intro and using this hext patch with `ff7_fps_limiter = 2` has the same timing as `ff7_fps_limiter = 1`

tangtang95:
If there were battle files related to 30 fps battle animation, it will be possible to test it better. Is there no way to generate them easily?

tangtang95:
For now, I have done the same for 30FPS. I think this hext patch makes the 30FPS behave as the original.

julianxhokaxhiu (approved):
Safe enough to try. LGTM :) Thanks!

PR #316, opened 2021-11-30
60FPS: Fix 30 and 60 fps battle camera movement

tangtang95:
- Fix battle intro camera movement for 30 and 60 fps battle
 - Fix battle outro camera movement for 30 and 60 fps battle
 - Fix 60 fps battle camera movement by changing approach and multiply by 4 the related frame values
 - Refactor battle menu fps limiter by using the VirtualProtect before changing value

Using the 60 fps battle animation in `direct/battle` and `direct/magic`, there is still an issue in the camera movement with limit breaks. I assume that the problem is that the 60 fps battle animation are not all 60 fps, but some of them are at 30 fps.


julianxhokaxhiu (approved):
You're a true legend. Thank you!

PR #317, opened 2021-12-01
60FPS: Fix horizontal camera rotation bug and refactor

tangtang95:
- Fix horizontal camera rotation bug in one battle outro and other magic camera animation (opcode 0xF8)
 - Patch opcode 0xE0 like the others (untested: still don't know where this is used)
 - Refactor better camera script data structure


julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #328, opened 2021-12-17
Voice: Fix pause and music fadeout logics

tangtang95:
Try to see if the pause truly does not work. I tried it and it seems that after pausing the voice was still going a little. Then after resuming it started a lot afterwards

### DIFF
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -958,7 +958,6 @@ void common_flip(struct game_obj *game_object)
 
 		*ff7_externals.swirl_limit_fps = 1;
 
-		ff7_handle_voice_playback();
 		ff7_handle_ambient_playback();
 	}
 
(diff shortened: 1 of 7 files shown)


tangtang95:
Still need to decide the filename based on the dialog text of sense, steal and other command that display text. For example, in sense there are 2 or 3 display texts and from what I saw, there is no need to substitute any enemy name or item. So, need to distinguish the different display text using maybe a substring?

julianxhokaxhiu (approved):
LGTM

Is there anything else pending on this? Looks like feature complete to me. Thanks for fixing the pause as well, I totally forgot you had a section ready for that, my bad!

tangtang95:
As I said, sense, steal, and the other commands needs some tweak to work well. If you want you can merge for now

julianxhokaxhiu:
Thank you :)

PR #331, opened 2021-12-19
Voice: Add voice support for display action/info

tangtang95:
- Add voice support for display info about steal, mug, sense, ...
- Disable keeping display text until voice end for display info about steal, mug, and etc
- Add voice support for display action such as magic, limit breaks, ...
- Implement random choice of character playing voice during display actions


julianxhokaxhiu (approved):
LGTM

Amazing, this is going to be a killer feature. Well done :)

julianxhokaxhiu:
Thank you!

PR #333, opened 2021-12-26
60FPS Battle: Many fix and interpolation of battle animations, and implemented pause trick

tangtang95:
This new update should makes battle almost 60 fps non-progressive with glitches on camera and glitches when using summons. Testing right now is not really useful since there could be huge changes when tweaking the pause trick. If you need to group it as a single commit, feel free to tell me. I thought it was better to leave them as multiple commits since it is clearer.

### DIFF
--- a/src/ff7_opengl.cpp
+++ b/src/ff7_opengl.cpp
@@ -31,6 +31,9 @@
 
 unsigned char midi_fix[] = {0x8B, 0x4D, 0x14};
 WORD snowboard_fix[] = {0x0F, 0x10, 0x0F};
+byte y_pos_offset_display_damage_30[] = {0, 1, 2, 3, 4, 5, 6, 6, 7, 7, 8, 8, 8, 8, 7, 7, 6, 6, 5, 4, 3, 2, 1, 0, 0, 0, 0, 0};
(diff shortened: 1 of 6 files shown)


src/ff7_opengl.cpp
```diff
@@ -31,6 +31,12 @@
 
 unsigned char midi_fix[] = {0x8B, 0x4D, 0x14};
 WORD snowboard_fix[] = {0x0F, 0x10, 0x0F};
+byte y_pos_offset_display_damage_30[] = {0, 1, 2, 3, 4, 5, 6, 6, 7, 7, 8, 8, 8, 8, 7, 7, 6, 6, 5, 4, 3, 2};
+//byte y_pos_offset_display_damage_60[] = {0, 0, 1, 1, 2, 2, 3, 3, 4, 4, 5, 5, 6, 6, 6, 6, 7, 7, 7, 7, 8, 8,
```
tangtang95, line 35: I left this one commented since this is the actual interpolation, but I changed it to make it smoother. Don't really know which one is better.

julianxhokaxhiu: Can you please put those lines into a single one instead of two? Thank you

julianxhokaxhiu (approved):
LGTM

julianxhokaxhiu:
Thank you!

PR #336, opened 2022-01-03
60FPS: Fix summon animation bugs and glitches + Add barebone for camera battle controls

tangtang95:
Now every summon in 60 fps mode should be running with no camera glitches and minimal animation glitches.
The second commit is the barebone for camera battle controls that Cosmos wanted


julianxhokaxhiu (approved):
LGTM

PR #337, opened 2022-01-03
FF7: allow to play oa (opening music) from external music instead of the movie

myst6re:


myst6re:
Tested and ready

julianxhokaxhiu (changes requested):
The PR code wise looks good to me. I'm just a bit conflicted on the flag name. What about `ff7_external_opening_music` instead?

myst6re:
> The PR code wise looks good to me. I'm just a bit conflicted on the flag name. What about `ff7_external_opening_music` instead of `ff7_play_opening_music`? Also, what does the PR does in concrete? Looks like it disables the `ob` playback, but it's not clear to me what it does overall.

The PC game does not plays the oa song. In fact the music is directly mixed into the intro movie. To makes it work, the developer has hardcoded a deactivation of oa when the field script tries to play it. And when the movie frame has reached one specific frame, it plays the ob track.

My PR disable this hack, so now the field script plays the oa only for the introduction, and ob when necessary.

> `ff7_play_opening_music`

No problem :) done!

julianxhokaxhiu (approved):
Thank you. All clear :) LGTM

PR #340, opened 2022-01-10
Implemented analogue camera controls in battle mode

CosmosXIII:
I added analogue controls for the battle camera. You can rotate the camera with the controller right stick and zoom-in/ zoom-out with the analogue triggers. Controls are only enabled when no camera animation is being executed.

### DIFF
--- a/src/ff7/camera.h
+++ b/src/ff7/camera.h
@@ -1,0 +1,42 @@
+/****************************************************************************/
+//    Copyright (C) 2021 Cosmos                                             //
+//                                                                          //
+//    This file is part of FFNx                                             //
+//                                                                          //
+//    FFNx is free software: you can redistribute it and/or modify          //
+//    it under the terms of the GNU General Public License as published by  //
+//    the Free Software Foundation, either version 3 of the License         //
+//                                                                          //
+//    FFNx is distributed in the hope that it will be useful,               //
+//    but WITHOUT ANY WARRANTY; without even the implied warranty of        //
+//    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         //
+//    GNU General Public License for more details.                          //
+/****************************************************************************/
+
+#pragma once
+
+#include "../ff7.h"
+
+class Camera
+{
+    public:
+        Camera() = default;
+        ~Camera() = default;
+
+        void setRotationSpeed(float rotX, float rotY, float rotZ);
+        void setZoomSpeed(float speed);
+        void reset();
+
+        void controlCamera(camera_vec3* cameraPosition, camera_vec3* cameraFocusPosition);
+    private:
+        struct point3d rotationSpeed = { 0.0, 0.0, 0.0 };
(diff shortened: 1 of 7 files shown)


julianxhokaxhiu (approved):
LGTM

julianxhokaxhiu:
Thank you!

src/ff7/camera.h
```diff
@@ -0,0 +1,42 @@
...
+        struct point3d rotationOffset = { 0.0, 0.0, 0.0 };
+        float zoomSpeed = 0.0f;
+        float zoomOffset = 0.0f;
+        const float minZoomDist = 5000.0f;
+        const float maxZoomDist = 30000.0f;
+        const float minVerticalAngle = 5.0f;
+        const float maxVerticalAngle = 85.0f;
+};
+
+extern Camera camera;
```
sithlord48, line 42: Newline should be at the EOF

src/ff7/camera.h
```diff
@@ -0,0 +1,42 @@
+/****************************************************************************/
+//    Copyright (C) 2021 Cosmos                                             //
```
sithlord48, line 2: Its 2022 :d and you should use a real name anyone can claim to be "Cosmos"

src/ff7/misc.cpp
```diff
@@ -248,20 +286,24 @@ struct ff7_gamepad_status* ff7_update_gamepad_status()
 			ff7_externals.gamepad_status->button5 = gamepad.IsPressed(XINPUT_GAMEPAD_LEFT_SHOULDER); // L1
 			ff7_externals.gamepad_status->button6 = gamepad.IsPressed(XINPUT_GAMEPAD_RIGHT_SHOULDER); // R1
 			ff7_externals.gamepad_status->button7 = gamepad.leftTrigger > 0.85f; // L2
-			ff7_externals.gamepad_status->button8 = gamepad.rightTrigger > 0.85f; // R2
+
+			// Remap R2 trigger to R3 so that we can use the trigger for zoom-in when analogue controls are enabled
```
sithlord48, line 290: Please do not change controls if a feature is active. 
What happens if this is enabled and i want to remap w/ R2 / L2 do in battle ? 

I suggest
  Either use L3 / R3 for zoom 
                 or 
  Hold R2 or L2 and use the right stick's up / down  for zoom

sithlord48 (commented):
Awesome Feature !

src/ff7/camera.h
julianxhokaxhiu: I updated it overall. Good catch, thanks!

src/ff7/camera.h
julianxhokaxhiu: Not really mandatory, but it's a suggested Unix best practice.

Tharsos83:
Hi, I noticed the following about this feature: The speedhack is activated via controller with L3 + L2/R2 (Playstation). L2/R2 buttons are also used for zooming in fights. This means that you always zoom in during battle when you switch on the speedhack.

PR #351, opened 2022-01-17
60FPS: Interpolate animation effects about textures and materials (spells, enemy moves, ...)

tangtang95:
Most of the spell, enemy moves, summons, and limit breaks effect should be interpolated. However there is a problem with some effect handler functions that calls the function I replaced with a "for" statement (probably particle effects), so, all those particles cannot be identified univocally.

I also fixed the encounter rate as it was too high for both WORLDMAP and FIELD mode. 

I added support for 30FPS battle using the same animation files, but the summons are not working since they do not have the 60FPS animation files. Could fix it in code by checking the actorID, but if in the future summon model will run in 60 FPS there is no need anymore.

### DIFF
--- a/src/common_imports.h
+++ b/src/common_imports.h
@@ -116,6 +116,34 @@ typedef void (gfx_field_EC)(struct game_obj *);
  * downright misleading. Thread with caution!
  */
 
+struct color_ui8
+{
+    byte r;
+    byte g;
+    byte b;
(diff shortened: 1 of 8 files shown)


julianxhokaxhiu (approved):
LGTM

julianxhokaxhiu:
Thank you!

PR #354, opened 2022-01-22
Camera zoom for controllers without analog triggers and deadzone option

CosmosXIII:
Added camera zoom support for controllers without analog triggers as requested by Satsuki.
Also added analog deadzone options for the right stick and analog triggers.


julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #361, opened 2022-01-23
60FPS: Fix FIELD bugs and softlocks

tangtang95:
- Fix partial animation (opcode CANM1, CANM2, CANIM1, CANIM2) wrong final frame value
- Fix scripted movement for field models by interpolating the position without checking for collision
- Fix FMVs animation for 60 FPS by removing script_ASPED modification because the problem was actually in non-interpolated animations


julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #363, opened 2022-01-28
recalcInternals fix

JeffBittle:
This is intended to fix an issue with inconsistent scaling results with certain resolution combinations.  In addition, it should make the internal resolution scaling from the original game resolution more consistent across the board.

This will require that the default value of internal_resolution_scale in the FFNx.toml be changed to 0 or less, which will make the default behavior of this function attempt to find the closest fitting internal resolution to the desired output resolution.

### DIFF
--- a/src/cfg.cpp
+++ b/src/cfg.cpp
@@ -203,7 +203,7 @@ void read_cfg()
 	show_version = config["show_version"].value_or(true);
 	window_size_x = config["window_size_x"].value_or(0);
 	window_size_y = config["window_size_y"].value_or(0);
-	internal_resolution_scale = config["internal_resolution_scale"].value_or(2);
+	internal_resolution_scale = config["internal_resolution_scale"].value_or(0);
 	preserve_aspect = config["preserve_aspect"].value_or(true);
 	fullscreen = config["fullscreen"].value_or(false);
 	refresh_rate = config["refresh_rate"].value_or(0);
--- a/src/renderer.cpp
+++ b/src/renderer.cpp
@@ -445,8 +445,6 @@ bool Renderer::doesItFitInMemory(size_t size)
 
 void Renderer::recalcInternals()
 {
-    bool is16by9 = false;
-
     viewWidth = window_size_x;
     viewHeight = window_size_y;
 
@@ -469,26 +467,26 @@ void Renderer::recalcInternals()
         }
     }
 
-    if ((viewWidth % game_width) > 0 || (viewHeight % game_height) > 0)
+	// If internal_resolution_scale from settings is less than one, calculate the closest fit for the output resolution, otherwise use the value directly
+    if (internal_resolution_scale < 1)
     {
         long scaleW = ::round(viewWidth / (float)game_width);
         long scaleH = ::round(viewHeight / (float)game_height);
 
         if (scaleH > scaleW) scaleW = scaleH;
-        if (scaleW > internal_resolution_scale) internal_resolution_scale = scaleW + 1;
-
-        is16by9 = true;
+        if (scaleW > internal_resolution_scale) internal_resolution_scale = scaleW;
+		if (internal_resolution_scale < 1) internal_resolution_scale = 1;
     }
 
-    // In order to prevent weird glitches while rendering we need to use the closest resolution to native's game one
-    framebufferWidth = is16by9 ? game_width * internal_resolution_scale : viewWidth * internal_resolution_scale;
-    framebufferHeight = is16by9 ? game_height * internal_resolution_scale : viewHeight * internal_resolution_scale;
+    // Use the set or calculated scaling factor to determine the width and height of the framebuffer according to the original resolution
+    framebufferWidth = game_width * internal_resolution_scale;
+    framebufferHeight = game_height * internal_resolution_scale;
 
     framebufferVertexWidth = (viewWidth * game_width) / window_size_x;
     framebufferVertexOffsetX = (game_width - framebufferVertexWidth) / 2;
 
     // Let the user know about chosen resolutions
-    ffnx_info("Original resolution %ix%i, New resolution %ix%i, Internal resolution %ix%i\n", game_width, game_height, window_size_x, window_size_y, framebufferWidth, framebufferHeight);
+    ffnx_info("Original resolution %ix%i, Scaling factor %i, Internal resolution %ix%i, Output resolution %ix%i\n", game_width, game_height, internal_resolution_scale, framebufferWidth, framebufferHeight, window_size_x, window_size_y);
 }
 
 void Renderer::prepareFramebuffer()
--- a/misc/FFNx.toml
+++ b/misc/FFNx.toml
@@ -35,12 +35,13 @@ window_size_y = 0
 
 #[INTERNAL RESOLUTION SCALE]
 # The scale is in multiples of 640x480
-# The scale factor is used to render internally at a larger size then the display, this will then be downscaled to the games current resolution
+# The scale factor is used to multiply the internally used resolution to a given size before it is scaled to the final screen or window resolution.
 # This is required to avoid visual glitches that may happen when the game is not rendered in a 4:3 aspect ratio.
-# This is optional for native 4:3 resolutions, but it will enhance visual quality ( like Nvidia DSR ).
-# Default = 2
+# This is optional for native 4:3 resolutions, but it can enhance visual quality at values higher than the current screen or window resolution ( like Nvidia DSR ).
+# A setting of 0 will attempt to find a scale factor that most closely matches the current screen or window resolution.
+# Default = 0
 #~~~~~~~~~~~~~~~~~~~~~~~~~~~
-internal_resolution_scale = 2
+internal_resolution_scale = 0
 
 #[PRESERVE ASPECT]
 # Preserve original game aspect ratio of (4:3) by adding black bars on the left and right side (if needed)


julianxhokaxhiu:
Can you please squash all your commits into one? Thank you :)

src/cfg.cpp
```diff
@@ -203,7 +203,7 @@ void read_cfg()
 	show_version = config["show_version"].value_or(true);
 	window_size_x = config["window_size_x"].value_or(0);
 	window_size_y = config["window_size_y"].value_or(0);
-	internal_resolution_scale = config["internal_resolution_scale"].value_or(2);
+	internal_resolution_scale = config["internal_resolution_scale"].value_or(0);
```
julianxhokaxhiu, line 206: If you update this value, you should updated it also here: https://github.com/julianxhokaxhiu/FFNx/blob/master/misc/FFNx.toml#L43

src/renderer.cpp
```diff
@@ -469,26 +467,26 @@ void Renderer::recalcInternals()
...
+        if (scaleW > internal_resolution_scale) internal_resolution_scale = scaleW;
+		if (internal_resolution_scale < 1) internal_resolution_scale = 1;
     }
 
-    // In order to prevent weird glitches while rendering we need to use the closest resolution to native's game one
-    framebufferWidth = is16by9 ? game_width * internal_resolution_scale : viewWidth * internal_resolution_scale;
-    framebufferHeight = is16by9 ? game_height * internal_resolution_scale : viewHeight * internal_resolution_scale;
+    // Use the set or calculated scaling factor to determine the width and height of the framebuffer according to the original resolution
+    framebufferWidth = game_width * internal_resolution_scale;
+    framebufferHeight = game_height * internal_resolution_scale;
```
julianxhokaxhiu, line 483: What about 16by9 resolutions? A Framebuffer MUST inherit the same width and height of your window ( be it fullscreen or with a border, doesn't matter )

JeffBittle: The existing code isn't doing anything special for non-4:3 screens anyway.  Ultimately all it's doing is determining what the multiplier of the of the original game resolution needs to be.  The internal_resolution_scale setting as it stands right now is not behaving in a logical fashion.  You would expect such a setting to directly multiply the internal resolution by that amount, and right now it isn't necessarily doing that.

src/cfg.cpp
JeffBittle: Ah ok, I wondered if the FFNx.toml was in there somewhere as well.  I'll make the change there as well.

src/renderer.cpp
julianxhokaxhiu: The lines I've highlighted here in the comment as exactly the issue. You might think you're fixing only the calculation, but as you noticed the result of that calculation ends up to be assigned into a variable that is used to initialize the framebuffer width and height. Have you tested this patch on your own side on top of say these resolutions?
- 640x480 ( 4:3 )
- 1280x1024 ( 4:3 )
- 1280x800 ( 16:10 )
- 1920x1080 ( 16:9 )
- 1680x1050 ( 16:10 )
- 2560×1080 ( 21:9 )

JeffBittle:
Hmm, I've just been doing this through the Github website itself... is it even possible to squash commits directly here?

julianxhokaxhiu:
> Hmm, I've just been doing this through the Github website itself... is it even possible to squash commits directly here?

Don't this so. You need to use Git on your own workstation.

src/renderer.cpp
JeffBittle: Plugging these given values into the existing function, with preserve_aspect true and the assumed default internal_resolution_scale of 2 gives the following results:

640x480
Original resolution 640x480, New resolution 640x480, Internal resolution 1280x960
1280x1024
Original resolution 640x480, New resolution 1280x1024, Internal resolution 2560x1920
1280x800
Original resolution 640x480, New resolution 1280x800, Internal resolution 1280x960
1920x1080
Original resolution 640x480, New resolution 1920x1080, Internal resolution 1280x960
1680x1050
Original resolution 640x480, New resolution 1680x1050, Internal resolution 1280x960
2560x1080
Original resolution 640x480, New resolution 2560x1080, Internal resolution 1280x960

Plugging these values into my proposed function, also with preserve_aspect true and at the proposed auto-calculation default of 0 gives the following results (note that I did make a slight change to the log text, the internal resolution being in the middle and the final output resolution being last made more sense):

640x480
Original resolution 640x480, Internal resolution 640x480,  Output resolution 640x480
1280x1024
Original resolution 640x480, Internal resolution 1280x960,  Output resolution 1280x1024
1280x800
Original resolution 640x480, Internal resolution 1280x960,  Output resolution 1280x800
1920x1080
Original resolution 640x480, Internal resolution 1280x960,  Output resolution 1920x1080
1680x1050
Original resolution 640x480, Internal resolution 1280x960,  Output resolution 1680x1050
2560x1080
Original resolution 640x480, Internal resolution 1280x960,  Output resolution 2560x1080

Right now, my proposed solution tries to match the closest multiple, and in some cases that means it's a little smaller than the output resolution.  I could amend this so that it's always one step over the output resolution so it is always downscaling when attempting to auto-fit, if that would be preferable.  But another thing you should notice in the existing function, in the 1280x1024 example for instance, it is creating a 2560x1920 framebuffer, which is two multiples above the closest fit, instead of a 1920x1440 framebuffer, which would still be a multiple above the final output resolution, so it would still be downscaled without being excessively large.

JeffBittle:
> > Hmm, I've just been doing this through the Github website itself... is it even possible to squash commits directly here?
> 
> Don't think so. You need to use Git on your own workstation.

Hmm, ok.  I do have Git set up, just haven't been using it for this since it was a relatively easy thing to fix.

src/renderer.cpp
julianxhokaxhiu: Did you build your code and test it or are you going theoretical here?

JeffBittle: I mean, the function works.  It's pretty simple all things considered.  But no, I'm not set up to build the DLL right now.  I made a test out of the original function to see what it was doing, and it exactly matched the output of the FFNx.log when setting 7th Heaven to specific resolutions.  I then made my proposed solution in a similar manner to make sure I was getting the desired result, which is how I discovered a very unlikely but possible edge case if someone could actually set their resolution on a particular device lower than 320x240, which could've caused the final internal_resolution_scale value to become 0, so I put in a quick handler for that.

Then I went through the C++ code and made my changes to it, which again, honestly weren't very complex.

JeffBittle:
> > Hmm, I've just been doing this through the Github website itself... is it even possible to squash commits directly here?
> 
> Don't think so. You need to use Git on your own workstation.

Gah, well that's annoying... I squashed the commits into one, but the previous ones are still there.  Hmm...

src/renderer.cpp
julianxhokaxhiu: Please consider always testing your code before submitting any PR. As I said already, the function atomically patched might make sense, but there are other complexities you're not considering. You'll notice the side effects of this patch as soon as you start testing it as a whole.

JeffBittle: There really aren't, this is doing just what the original function was doing just without the bugginess that happened with some resolutions.

JeffBittle: You keep saying things about the framebuffers being related to the final screen size, but that's not what the existing function is doing.  All it's trying to do is figure out what scale of a 4:3 resolution framebuffer to make, and the last two lines are calculating how to center it horizontally on the screen, because then when the renderFrame() function is called, it creates the 4 vertices it needs to create the textured quad that the framebuffer will be written to on screen updates, and those vertices are offset by the correct amount to center them in the screen's width.  That's the only thing, outside of the aspect correction code which I didn't touch at all, that the original function is using the full screen/window resolution for in the first place, to determine the X offset and the width of the vertex framebuffer in relation to the original 640 pixel width.

julianxhokaxhiu: `framebufferWidth` and `framebufferHeight` are used to create the framebuffer, where all game related draw commands will be executed.

JeffBittle: Yes.  That has nothing to do with what I'm saying however.

JeffBittle: Getting the build environment set up now, this takes a while.  I'm also not sure what I can do about the commits, I might end up having to delete the fork entirely and then remake it so I can combine the commits from the start.

JeffBittle: Most of the discussion took place in Discord, but for posterity I'll update this thread.   I have successfully built a version of FFNx with my proposed fix and have tested it in game.  In looking for any sorts of graphical glitches like misalignment or lines through field screens or anything strange in battles, I haven't noticed any while testing at several different resolutions and aspect ratios.

Example of a 4:3 aspect ratio with auto-fit:
1280x960at1280x1024
1280x960at1280x1024_battle
> INFO: Original resolution 640x480, Scaling factor 2, Internal resolution 1280x960, Output resolution 1280x1024

In checking with the FFNx.log file after each different resolution test, I am noticing much more consistent framebuffer sizes being created for the internal resolution with the default auto-fit internal_resolution_scale value of 0 that are much closer to the final output resolution than some of the strangely large framebuffer sizes being created via the previous function (i.e. 1920x1200 is no longer getting a 4x framebuffer of 2560x1920 for no particular reason at the old default scale value of 2).

Example of a 16:10 aspect ratio with auto-fit:
1920x1440at1920x1200
1920x1440at1920x1200_battle
> INFO: Original resolution 640x480, Scaling factor 3, Internal resolution 1920x1440, Output resolution 1920x1200

Additionally, setting the internal_resolution_scale to values above 0 is correctly allowing the internal resolution to be specifically set which could be useful for either performance or aesthetic reasons.  For instance, a setting of 1 allows the game to remain rendered at the original 640x480 resolution before being drawn to the final screen resolution.  This gives more of a retro or classic look, as the game would've appeared on PC back in 1998.
Example of a 16:9 aspect ratio with an explicit scale factor of 1, resulting in a 640x480 framebuffer on the backend:
640x480at1920x1080
640x480at1920x1080_battle
> INFO: Original resolution 640x480, Scaling factor 1, Internal resolution 640x480, Output resolution 1920x1080

In regard to performance concerns, suppose someone with a 1920x1200 screen but a weaker video card was trying to play FF7 via FFNx and found performance to be sluggish because of the 2560x1920 framebuffer being created on the backend.  Currently they would have little recourse in trying to improve performance other than playing in a window or non-native resolution for their display because of the way the function is auto-correcting the internal_resolution_scale value at runtime.  With this fix, they could set the scale value to 2 to make sure they get specifically a 1280x960 framebuffer which their card could potentially render more easily.

For the above screenshots I decided to stick with no mods whatsoever, just to show a consistent look, although I did leave the lighting mode enabled.  However, even with high resolution textures like SYW, I still saw no graphical artifacts in field scenes or within battles.  Make sure to open the images in a full screen view in order to see the details of the game rendering, at small size the downscaling will make them look mostly the same.

PR #364, opened 2022-01-29
recalcInternals fix and new defaults

JeffBittle:
This includes both a fix to how the recalcInternals function works as well as the new default value of 0 for the settings files.

A huge thank you to Cosmos for helping me get my tangled git branch all sorted so I could actually submit this PR as a single commit, as requested.  Unfortunately it did require closing the initial PR to resubmit.

### DIFF
--- a/src/renderer.cpp
+++ b/src/renderer.cpp
@@ -445,7 +445,7 @@ bool Renderer::doesItFitInMemory(size_t size)
 
 void Renderer::recalcInternals()
 {
-    bool is16by9 = false;
+    long scale_factor = internal_resolution_scale;
 
     viewWidth = window_size_x;
     viewHeight = window_size_y;
@@ -469,26 +469,26 @@ void Renderer::recalcInternals()
         }
     }
 
-    if ((viewWidth % game_width) > 0 || (viewHeight % game_height) > 0)
+    // If internal_resolution_scale from settings is less than one, calculate the closest fit for the output resolution, otherwise use the value directly
+    if (internal_resolution_scale < 1)
     {
         long scaleW = ::round(viewWidth / (float)game_width);
         long scaleH = ::round(viewHeight / (float)game_height);
 
         if (scaleH > scaleW) scaleW = scaleH;
-        if (scaleW > internal_resolution_scale) internal_resolution_scale = scaleW + 1;
-
-        is16by9 = true;
+        if (scaleW > internal_resolution_scale) scale_factor = scaleW;
+        if (scale_factor < 1) scale_factor = 1;
     }
 
-    // In order to prevent weird glitches while rendering we need to use the closest resolution to native's game one
-    framebufferWidth = is16by9 ? game_width * internal_resolution_scale : viewWidth * internal_resolution_scale;
-    framebufferHeight = is16by9 ? game_height * internal_resolution_scale : viewHeight * internal_resolution_scale;
+    // Use the set or calculated scaling factor to determine the width and height of the framebuffer according to the original resolution
+    framebufferWidth = game_width * scale_factor;
+    framebufferHeight = game_height * scale_factor;
 
     framebufferVertexWidth = (viewWidth * game_width) / window_size_x;
     framebufferVertexOffsetX = (game_width - framebufferVertexWidth) / 2;
 
     // Let the user know about chosen resolutions
-    ffnx_info("Original resolution %ix%i, New resolution %ix%i, Internal resolution %ix%i\n", game_width, game_height, window_size_x, window_size_y, framebufferWidth, framebufferHeight);
+    ffnx_info("Original resolution %ix%i, Scaling factor %i, Internal resolution %ix%i, Output resolution %ix%i\n", game_width, game_height, scale_factor, framebufferWidth, framebufferHeight, window_size_x, window_size_y);
 }
 
 void Renderer::prepareFramebuffer()
--- a/misc/FFNx.toml
+++ b/misc/FFNx.toml
@@ -35,12 +35,13 @@ window_size_y = 0
 
 #[INTERNAL RESOLUTION SCALE]
 # The scale is in multiples of 640x480
-# The scale factor is used to render internally at a larger size then the display, this will then be downscaled to the games current resolution
+# The scale factor is used to multiply the internally used resolution to a given size before it is scaled to the final screen or window resolution.
 # This is required to avoid visual glitches that may happen when the game is not rendered in a 4:3 aspect ratio.
-# This is optional for native 4:3 resolutions, but it will enhance visual quality ( like Nvidia DSR ).
-# Default = 2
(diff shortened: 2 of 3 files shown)


JeffBittle:
I have successfully built a version of FFNx with my proposed fix and have tested it in game.  In looking for any sorts of graphical glitches like misalignment or lines through field screens or anything strange in battles, I haven't noticed any while testing at several different resolutions and aspect ratios.

Example of a 4:3 aspect ratio with auto-fit:
1280x960at1280x1024
1280x960at1280x1024_battle
> INFO: Original resolution 640x480, Scaling factor 2, Internal resolution 1280x960, Output resolution 1280x1024

In checking with the FFNx.log file after each different resolution test, I am noticing much more consistent framebuffer sizes being created for the internal resolution with the default auto-fit internal_resolution_scale value of 0 that are much closer to the final output resolution than some of the strangely large framebuffer sizes being created via the previous function (i.e. 1920x1200 is no longer getting a 4x framebuffer of 2560x1920 for no particular reason at the old default scale value of 2).

Example of a 16:10 aspect ratio with auto-fit:
1920x1440at1920x1200
1920x1440at1920x1200_battle
> INFO: Original resolution 640x480, Scaling factor 3, Internal resolution 1920x1440, Output resolution 1920x1200

Additionally, setting the internal_resolution_scale to values above 0 is correctly allowing the internal resolution to be specifically set which could be useful for either performance or aesthetic reasons.  For instance, a setting of 1 allows the game to remain rendered at the original 640x480 resolution before being drawn to the final screen resolution.  This gives more of a retro or classic look, as the game would've appeared on PC back in 1998.
Example of a 16:9 aspect ratio with an explicit scale factor of 1, resulting in a 640x480 framebuffer on the backend:
640x480at1920x1080
640x480at1920x1080_battle
> INFO: Original resolution 640x480, Scaling factor 1, Internal resolution 640x480, Output resolution 1920x1080

In regard to performance concerns, suppose someone with a 1920x1200 screen but a weaker video card was trying to play FF7 via FFNx and found performance to be sluggish because of the 2560x1920 framebuffer being created on the backend.  Currently they would have little recourse in trying to improve performance other than playing in a window or non-native resolution for their display because of the way the function is auto-correcting the internal_resolution_scale value at runtime.  With this fix, they could set the scale value to 2 to make sure they get specifically a 1280x960 framebuffer which their card could potentially render more easily.

For the above screenshots I decided to stick with no mods whatsoever, just to show a consistent look, although I did leave the lighting mode enabled.  However, even with high resolution textures like SYW, I still saw no graphical artifacts in field scenes or within battles.  Make sure to open the images in a full screen view in order to see the details of the game rendering, at small size the downscaling will make them look mostly the same.

JeffBittle:
I have been informed by Satsuki_Yatoshi that the sorts of graphical issues they were seeing in rendering during development were actually happening with high resolution textures, especially at 2x and 4x scales on 1920x1080, and that those are the sorts of circumstances where I should be looking for issues while using this proposed solution.  I will update this post with appropriate screenshots to document my findings.

julianxhokaxhiu (approved):
LGTM - Screenshots seems to confirm the issue is not there if touched. Were you able to test your own build with @satsukiyatoshi? Did he found anything critical?

JeffBittle:
> LGTM - Screenshots seems to confirm no visual glitches may appear after this patch. Were you able to test your own build with @satsukiyatoshi? Did he found anything critical?

The build of FFNx with my changes has been sent to @satsukiyatoshi, so I'm hoping we can get some tests with different hardware just to confirm no visual glitches, as I only have an AMD video card myself.  And different resolution and internal resolution scale combinations of course.

src/renderer.cpp
```diff
@@ -469,26 +467,26 @@ void Renderer::recalcInternals()
         }
     }
 
-    if ((viewWidth % game_width) > 0 || (viewHeight % game_height) > 0)
+    // If internal_resolution_scale from settings is less than one, calculate the closest fit for the output resolution, otherwise use the value directly
+    if (internal_resolution_scale < 1)
```
julianxhokaxhiu, line 471: After some testing with @satsukiyatoshi he found out that the autoscaling works only once, when you run the game, then doesn't matter how many times you resize the window, this value will never change.

This is because the IF here, despite the function code being called everytime the window resolution changes, it will never be hit the moment it gets assigned a value inside the if.

Because of this you need to find another way to make sure this code is being hit everytime IF the user decided to put 0 as value. Also, please find a way that doesn't bring another flag in the toml file.

Thank you!

julianxhokaxhiu (changes requested):

satsukiyatoshi:
Current patch is working to set on the fly internal resolution but crash the game is using a very small resolution (trying to set ultra big internal resolution):

[00000126] INFO: Original resolution 640x480, Scaling factor 109, Internal resolution 4224x52320, Output resolution 185x223
[00000126] ERROR: [Device Lost] s_renderD3D11->m_device->CreateTexture2D(&desc, 0, &m_rt2d) FAILED 0x80070057

[00000127] ERROR: [Device Lost] s_renderD3D11->m_device->CreateTexture2D(&desc, kk == 0 ? 0 : srd, &m_texture2d) FAILED 0x80070057

[00000127] ERROR: [Device Lost] s_renderD3D11->m_device->CreateShaderResourceView(m_ptr, &srvd, &m_srv) FAILED 0x80070057

JeffBittle:
> Current patch is working to set on the fly internal resolution but crash the game is using a very small resolution (trying to set ultra big internal resolution):
> 
> [00000126] INFO: Original resolution 640x480, Scaling factor 109, Internal resolution 4224x52320, Output resolution 185x223 [00000126] ERROR: [Device Lost] s_renderD3D11->m_device->CreateTexture2D(&desc, 0, &m_rt2d) FAILED 0x80070057
> 
> [00000127] ERROR: [Device Lost] s_renderD3D11->m_device->CreateTexture2D(&desc, kk == 0 ? 0 : srd, &m_texture2d) FAILED 0x80070057
> 
> [00000127] ERROR: [Device Lost] s_renderD3D11->m_device->CreateShaderResourceView(m_ptr, &srvd, &m_srv) FAILED 0x80070057

This was a little tricky, but apparently our attempts to use what we thought was a little more efficient variable type didn't really agree with how round() returns values.  I switched them back to long type and now the window can be successfully resized all the way down to absolute minimum, 120x0, without causing a crash.

julianxhokaxhiu (approved):
LGTM Code wise - @satsukiyatoshi good to go on your side too?

satsukiyatoshi:
form my test no bug , crash , or parasite line, looks good to me ATM

julianxhokaxhiu:
Thank you to both :)

JeffBittle:
Very cool to see.  And a big thanks to you both as well for the help in getting this worked out!  My first code contribution to an open source project.

PR #365, opened 2022-01-29
Optimized walkmesh shadow calculations to be done only once per field

CosmosXIII:
Optimized walkmesh shadow calculations to be done only once when changing fields.


julianxhokaxhiu (approved):
LGTM

PR #367, opened 2022-01-31
60FPS: Fix split and join opcode movement and rotation

tangtang95:
This makes the movement of characters during split and join opcode like the OG 30 FPS, which should avoid softlocks related to this.
It also fixes a bug that made the movement during split and join even slower.


julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #375, opened 2022-02-06
FF8: handle vram

myst6re:


julianxhokaxhiu (approved):
Thank you very much for this contribution!

PR #377, opened 2022-02-08
FF7: Fixed field shadows in movies

CosmosXIII:
Thanks to vertex2995 we were able to get the view matrix used in movies. Using this matrix for field shadow rendering fixes the problem where shadows would not be rendered in their correct position.

### DIFF
--- a/src/common_imports.h
+++ b/src/common_imports.h
@@ -196,18 +196,6 @@ struct struc_77
 	struct struc_173 struc_173;
 };
 
-// Field camera axis
-typedef struct {
-	signed short x;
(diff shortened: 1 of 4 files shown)


julianxhokaxhiu:
Fantastic work, can I ask if you can please update your commit message and prefix it with `FF7:` please? Thank you :)

CosmosXIII:
> Fantastic work, can I ask if you can please update your commit message and prefix it with `FF7:` please? Thank you :)

Done :)

julianxhokaxhiu (approved):
Thank you!

PR #380, opened 2022-02-09
FF7: Fixed field shadow crash when changing game modes

CosmosXIII:
Fixed field shadow crash when changing game modes. I reverted the pointer check for the level data that was there before and worked just fine.


julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you :)

PR #381, opened 2022-02-11
Lighting: Fixed lighting shader output

CosmosXIII:
Fixed a little mistake from the previous fix where it was setting the wrong output variable.


julianxhokaxhiu:
But is this actually rendering in HDR mode? The line below you removed is actually required to have the proper color rendered in Linear space when HDR is enabled. I think we need to re-work this patch.

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #389, opened 2022-02-14
Voice: Fix BATTLE character action voices bugs

tangtang95:
- Fix voice volume not being set when using character action voices
 - Fix rerouting of voice filenames
 - Move voice battle pause and resume mechanism in common_flip like ambient sounds


julianxhokaxhiu:
Good catch on those fixes. Thanks :)

PR #394, opened 2022-02-16
60FPS: Fix WORLDMAP wait frames in the script + Small fix of FF8

tangtang95:


src/movies.cpp
```diff
@@ -349,6 +349,8 @@ void ff8_prepare_movie(uint32_t disc, uint32_t movie)
 			if(res > 0) camOffset += res;
 		}
 
+		fclose(camFile);
```
julianxhokaxhiu, line 352: Shouldn't this be inside the `if`? If you're on disk 4 this code will crash as it is.

julianxhokaxhiu (changes requested):

src/movies.cpp
tangtang95: Which if? Right now I'm closing it after it has been read inside the while

julianxhokaxhiu: The if block just above this line. Where the `fopen` happens :)

julianxhokaxhiu (changes requested):

src/movies.cpp
tangtang95: If you are talking about the if block on line 339, that enters only if the file is not opened correctly. I don't think there is the need to close it when it is not opened

julianxhokaxhiu: Can you just drop this commit for now and we can re-evaluate later? Thanks!

tangtang95: Done

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #402, opened 2022-02-21
FF7: Improve worldmap footsteps + Add worldmap highwind sound

tangtang95:
The goal is to add a loop sound to highwind and then change it like Tiny Bronco (changing the pitch???) when it's moving. Or maybe we could only add sound when highwind is moving

### DIFF
--- a/src/sfx.cpp
+++ b/src/sfx.cpp
@@ -531,7 +531,7 @@ void sfx_process_footstep(bool is_player_moving)
 	}
 }
 
-void sfx_play_wm_footstep(int player_model_id, int player_walkmap_type)
+void sfx_process_wm_footstep(int player_model_id, int player_walkmap_type)
 {
 	static time_t last_playback_time, current_playback_time;
 	float pace = 0.3f;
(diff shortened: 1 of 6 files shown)


src/sfx.cpp
```diff
@@ -564,6 +564,24 @@ void sfx_play_wm_footstep(int player_model_id, int player_walkmap_type)
...
+		if(is_highwind_motor_on)
+			ff7_sfx_set_panning_on_channel(32, 7);
+		else
+		 	ff7_sfx_set_panning_on_channel(0, 7);
+
+		char track_name[64];
+		sprintf(track_name, "sfx_highwind");
+		if(!nxAudioEngine.isSFXPlaying(7))
+			nxAudioEngine.playSFX(track_name, default_sfx_id, 7, 0.0f);
+	}
```
julianxhokaxhiu, line 582: Can you please add this block here?
```cpp
else
{
  common_externals.play_sfx(default_sfx_id);
}
```

I'd also move the block
```cpp
if(is_highwind_motor_on)
  ff7_sfx_set_panning_on_channel(32, 7);
else
  ff7_sfx_set_panning_on_channel(0, 7);
```

Out side the if so it's common for both cases. LMK what you think.

julianxhokaxhiu (changes requested):
Outside of a minor change request, the rest looks solid. Thank you :) Looking forward!

src/sfx.cpp
tangtang95: I didn't want to add sound for non-external sfx because I don't think there is a good sound. Tiny bronco one is kinda bad for highwind. I also need to find the original id for that sfx

tangtang95: Cannot modify panning on non-external sfx even by using `common_externals.set_sfx_panning_on_channel` function. Probably the channel value is wrong.

julianxhokaxhiu: Yes, channel 7 doesn't exist in vanilla. I don't remember which one is used for bronco so you have to use the same. I think it's 5 but not 100% sure.

Regarding the else block, allright, if it's intended fine by me.

tangtang95:
I have modified a lot the logic by following this function `75E6CC` in FF7.exe which is used by worldmap to play looped SFX

src/sfx.cpp
```diff
@@ -564,6 +564,36 @@ void sfx_play_wm_footstep(int player_model_id, int player_walkmap_type)
...
+		{
+			char track_name[64];
+			sprintf(track_name, "sfx_highwind");
+			nxAudioEngine.playSFX(track_name, default_sfx_id, default_channel, 0.0f, true);
+		}
+		else
+		{
+			common_externals.play_sfx_effects(64, default_sfx_id, 0, 0, 0);
+		}
+		playing = true;
```
julianxhokaxhiu, line 586: This variable should be set manually on the `common_externals.play_sfx_effects` as it doesn't return a value. But on the external case, it should be assigned based on the return value of `nxAudioEngine.playSFX`

julianxhokaxhiu (changes requested):

src/sfx.cpp
tangtang95: Done

julianxhokaxhiu (approved):
LGTM

julianxhokaxhiu:
Thank you!

PR #403, opened 2022-02-23
60FPS: Fix Great Glacier counter steps before fainting

tangtang95:
Also refactor MOVIE FPS fix by moving it all into ff7/field.cpp


julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #408, opened 2022-03-05
60FPS: Fix new rotation steps index not starting from zero in specific situation

tangtang95:
There are specific occasions where the opcode of turngen and turn rotation is stopped in the middle of the execution. When using 60 FPS, the new `rotation_steps_idx` is not resetted to 0. This will cause a softlock since the `rotation_steps_idx` is greater than `rotation_n_steps` and it won't stop rotating. In original `rotation_steps_idx` this is resetted to 0, otherwise it won't start rotating, so to fix it we just set the new `rotation_steps_idx` to 0 at the start of the rotation.


julianxhokaxhiu:
Thank you!

PR #409, opened 2022-03-05
60FPS: Rework the 60FPS rotation movement logic

tangtang95:
Since there could be other special occasion in which the previous rotation movement logic had issues, I redid the
new rotation movemeng logic by removing all the hooks where rotation steps are set. Instead, the only modification
is on the rotation update where the increment of the steps index is done only every 2 frame in 60FPS.
This solution is much simpler and less prone to bugs and glitches.


julianxhokaxhiu:
Thank you :)

PR #412, opened 2022-03-11
FF8 - Voice feature

myst6re:


src/voice.cpp
```diff
@@ -643,6 +645,60 @@ void ff7_display_battle_action_text()
...
+			simulate_OK_disabled[window_id] = false;
+			current_dialog_id = -1;
+		}
+	}
+
+	// Auto close the message if it was voice acted and the audio file has finished playing
+	if (current_opcode_message_status[window_id].is_voice_acting && !nxAudioEngine.isVoicePlaying(window_id))
+	{
+		current_opcode_message_status[window_id].is_voice_acting = false;
+		if (!simulate_OK_disabled[window_id]) simulate_OK_button = true;
```
myst6re, line 696: @julianxhokaxhiu do you know how to implement this for FF8?

julianxhokaxhiu: I think I can, I just need to figure out which is the relative mapped variable once the input is parsed, but as I did RE that part we can add it easily.

julianxhokaxhiu (approved):

julianxhokaxhiu:
The PR looks good enough for me to be merged as it is, fine by you if I do?

julianxhokaxhiu:
Thank you!

PR #416, opened 2022-03-17
60FPS: Fix bug in Cid attack SFX

tangtang95:
Whenever Cid attacked in 60FPS, it played Cloud attack SFX due to a bug in Cid animation script and mismatch timing in 60FPS. This is fixed by removing this SFX from playing at all only in 60FPS


julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #422, opened 2022-03-26
FF8: Vibration modding

myst6re:


julianxhokaxhiu (changes requested):
Overall the PR looks amazing. The only thing I would change is the path to tomls. Can you please consider using `vibrate/ff8` instead of `vibrate` as lookup path? I'd like to keep an open door for FF7 in the future :)

Thank you!

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #426, opened 2022-03-26
FF8: Adding more vibration to battle, like the R1 trigger

myst6re:


misc/FFNx.vibrate.ff8.battle.toml
```diff
@@ -0,0 +1,135 @@
+# FFNx Vibration Engine config file - FF8 Field module
```
julianxhokaxhiu, line 1: This should be `FF8 Battle module`

misc/FFNx.vibrate.ff8.world.toml
```diff
@@ -1,3 +1,32 @@
+# FFNx Vibration Engine config file - FF8 Field module
```
julianxhokaxhiu, line 1: This should be `FF8 World module`

julianxhokaxhiu (changes requested):

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #427, opened 2022-03-30
FF8: Configure game path and data drive via FFNx.toml

myst6re:


julianxhokaxhiu (approved):

julianxhokaxhiu:
I am not against this PR, I am just wondering what is the actual real use case. Do you have one at least where this PR makes sense and targets that scenario? Thank you!

myst6re:
> I am not against this PR, I am just wondering what is the actual real use case. Do you have one at least where this PR makes sense and targets that scenario? Thank you!

It allows more than one installation of FF8, I can switch between japan and french version easily. Without it I need to change the path the the game installation in the register.

julianxhokaxhiu:
Allright, fair enough. Good to be merged by me then. Thanks!

PR #428, opened 2022-04-01
FF8: more debug output + Fix crash

myst6re:


src/cfg.cpp
```diff
@@ -217,7 +217,7 @@ void read_cfg()
 	enable_lighting = config["enable_lighting"].value_or(false);
 	prefer_lighting_cpu_calculations = config["prefer_lighting_cpu_calculations"].value_or(true);
 	ff7_external_opening_music = config["ff7_external_opening_music"].value_or(false);
-	ff7_more_debug = config["ff7_more_debug"].value_or(false);
+	more_debug = config["more_debug"].value_or(false) || config["ff7_more_debug"].value_or(false);
```
julianxhokaxhiu, line 220: Keep only the flag that exists in the `FFNx.toml`. Therefore please remove this or logic, thank you :)

julianxhokaxhiu (changes requested):

src/cfg.cpp
myst6re: You don't want to keep the compatibility with old config files?

julianxhokaxhiu: No, that was never the case so far. If a flag changes, the code changes with it.

myst6re: Okay, as it is not a sensible flag, I can agree with you, but sometimes it is better to keep a compatibility layer for our users :-)

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you :)

PR #432, opened 2022-04-03
FF7: Fix deadzone bug for XINPUT analog triggers

tangtang95:
Always find it strange that deadzone never worked. The value of rightTrigger and leftTrigger is actually in between `[0, 1]`. https://github.com/julianxhokaxhiu/FFNx/blob/master/src/gamepad.cpp#L114-L115


tangtang95:
Wait on merging, I think it needs a fix also on the statement inside the if

julianxhokaxhiu:
Can you please rebase this branch? Azure apparently got stuck and never built this branch so far. Thanks!

tangtang95:
Done!

julianxhokaxhiu:
Thank you!

PR #442, opened 2022-05-16
Lighting: Fix field shadow glitch for 3d models

tangtang95:
Remove the special condition, in lighting vertex shader, used to skip original battle shadow that caused the field shadow triangle glitch


julianxhokaxhiu:
Can you please update the changelog as well?

julianxhokaxhiu:
Thank you :)

PR #449, opened 2022-06-20
60FPS: Implement smooth FIELD background movement

tangtang95:
Change the update mechanism related to following the player in FIELD mode by moving the background and the 3d models.
All positions of the 4 layers of the background are computed using floating point variables. This is also done for the 3d model world coordinates.
This removes all the wobbling background and 3d models. To avoid changing the original game, this smoothness in FIELD movement is applied only when using FULL 30 FPS or FULL 60 FPS mode.

### DIFF
--- a/src/ff7.h
+++ b/src/ff7.h
@@ -1940,7 +1940,7 @@ struct ff7_shake_bg_data
 	uint8_t do_shake;
 	uint8_t shake_phase;
 	char amp_index;
-	uint8_t shake_curr_value;
+	char shake_curr_value;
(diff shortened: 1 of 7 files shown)


julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

tangtang95:
Forgot to add it to the changelog. Do you want me to open a new PR for that?

julianxhokaxhiu:
Yes please, I'd appreciate

PR #451, opened 2022-06-21
60FPS: Fix MSAA tiles tearing by using centroid UV texture coords

tangtang95:
@CosmosXIII found this fix (https://www.opengl.org/pipeline/article/vol003_6/) for MSAA issue. This centroid evaluation is more computationally intensive and also less accurate, but it avoids the bad pixels. The best might be to use centroid evaluation only on 2D elements, but from testing, there is no much difference. For now, we could test and see what the users think and feel with this.

Since MSAA is no more problem, I increased the background position precision to 1/32. I don't think a higher precision is gonna make the background smoother.

### DIFF
--- a/misc/FFNx.varying.flat.def.sc
+++ b/misc/FFNx.varying.flat.def.sc
@@ -1,5 +1,5 @@
 flat vec4 v_color0    : COLOR0    = vec4(0.0, 0.0, 0.0, 0.0);
-vec2 v_texcoord0 : TEXCOORD0 = vec2(0.0, 0.0);
+centroid vec2 v_texcoord0 : TEXCOORD0 = vec2(0.0, 0.0);
 vec4 v_position0 : TEXCOORD4 = vec4(0.0, 0.0, 0.0, 0.0);
 vec4 v_shadow0 : TEXCOORD5 = vec4(0.0, 0.0, 0.0, 0.0);
 vec3 v_normal0 : NORMAL = vec3(0.0, 0.0, 0.0);
--- a/misc/FFNx.varying.smooth.def.sc
+++ b/misc/FFNx.varying.smooth.def.sc
@@ -1,5 +1,5 @@
 smooth vec4 v_color0    : COLOR0    = vec4(0.0, 0.0, 0.0, 0.0);
-vec2 v_texcoord0 : TEXCOORD0 = vec2(0.0, 0.0);
+centroid vec2 v_texcoord0 : TEXCOORD0 = vec2(0.0, 0.0);
 vec4 v_position0 : TEXCOORD4 = vec4(0.0, 0.0, 0.0, 0.0);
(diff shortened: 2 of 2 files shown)


julianxhokaxhiu:
We'd need to test this on FF8 too intensively before merging, as it might fix FF7 but might break FF8.

@myst6re when you have some time would you be abe to give this PR a spin? Thank you.

myst6re:
MSAA is already mostly broken on FF8, today enabling antialiasing is discouraged. I think this change will be more a fix for ff8 than a break.

I try to test it tonight though

tangtang95:
I have tested more this patch and I noticed that with internal resolution x6 without MSAA, or low MSAA (e.g. 2x) still has line tearing between tiles. So it's not fully fixed. Even with lower precision steps of 1/8

tangtang95:
Actually, the settings with internal resolution x6 and without MSAA is already broken in the current canary version. While IRx8 is fine without MSAA. Anyway this tiles tearing is very common when trying to move the camera/world coordinates smoothly with floats, they usually fix it by packing the tiles in a way to add padding between them.

CosmosXIII:
I also thought this was a driver bug but after reading how MSAA works I don' t think that is the case anymore.
As explained in the article, with MSAA enabled texels outside the triangle can be fetched near the triangle edges.
When working with extremely packed tile textures as in FF7 there is a high chance some samples fetch outside what is safe.
The fact that this patch does not fix all cases does not mean it is not the correct thing when working with extremely packed tiled textures.
Fetching outside of the safe areas can happen in other cases but with MSAA this becomes a lot more likely.
I think while this does not fix all cases, it fixes the problems introduced by MSAA.

I don't have an AMD card so I can not test it but maybe it would be useful to test if with MSAA enabled the rendered image is actually any different or it is doing nothing (in other words, are the edges actually antialiased?)

tangtang95:
My bad, pushed on the wrong branch. Let me fix it

julianxhokaxhiu:
Can you please rebase this branch? Thank you

julianxhokaxhiu:
So I tested this PR locally and I'm still seeing some horizontal lines on top of

> [00000024] INFO: Original resolution 640x480, Scaling factor 3, Internal resolution 1920x1440, Output resolution 2560x1377

So I guess this PR doesn't fix all possible cases, and IMHO we should just accept it's an Nvidia driver bug :(

tangtang95:
I noticed that scaling factor equals to multiple of 2 works quite well either with or without MSAA. While with all scaling factor, it should work with MSAAx16 (for example, scaling factor of 6 works well with MSAAx16 but not well with MSAA off or MSAAx2)

tangtang95:
If you are using MSAA, with renderdoc you get multiple samples of the framebuffer. Some samples might have no lines, while other samples have lines

julianxhokaxhiu:
Thank you!

PR #455, opened 2022-06-23
60FPS: Add smooth background movement in scripted scenarios

tangtang95:
With this, the background and 3d models movement are also smooth for scripted movements coming from the flevel


julianxhokaxhiu:
Thank you!

PR #456, opened 2022-06-24
60FPS: Set ideal precision to 1/10 for background positions

tangtang95:
This together with the centroid UV coordinate qualifier should avoid all tearing lines for whatever
combination of internal resolution scale (from 1x to 8x) and MSAA values.


tangtang95:
For now I think it works perfectly, at least from my testing.

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #466, opened 2022-07-09
60FPS: Fix model position glitch for some summons

tangtang95:
This is due to some timing issue that happens in 60 fps


julianxhokaxhiu (approved):

julianxhokaxhiu:
Thanks!

PR #467, opened 2022-07-13
FF7Achievement: Fix some bugs about cait sith and emerald weapon

tangtang95:
Also there was some warning on log telling that some externals were already defined. So I cleaned up those.


julianxhokaxhiu:
Marvelous work as always, thank you :)

PR #475, opened 2022-08-09
Lighting: draw call order fix and antialiasing sanity check

CosmosXIII:
This PR has the following lighting fixes:
- reworked field shadow rendering to preserve original draw call order. This should fix some bugs like the black screen after Cloud falls from Sector 5 when using vanilla assets and incorrect draw order in the temple of ancients clock room
- disabled antialiasing when lighting enabled to prevent incorrect behavior of the depth buffer backup/recover logic for field shadow rendering


julianxhokaxhiu:
Thank you!

PR #476, opened 2022-08-10
Lighting: fixed visual glitch when enabling both lighting and anti-aliasing

CosmosXIII:
This PR fixes visual glitch when enabling both lighting and anti-aliasing.
By disabling the z-write for the field shadow draw call it works just fine so no need to backup/recover the depth buffer anymore.


julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #478, opened 2022-08-12
Lighting: Fixed draw call order for framebuffer effects

CosmosXIII:
This PR fixes the framebuffer effects when lighting is enabled. 
The problem until now was that the blit draw call for the framebuffer effect was being executed at the wrong timing (before anything was drawn to the screen and therefore it was all black). By deferring the blit draw call in the same way as the other draw calls, we can reproduce the original intended order of execution. The same applies for the clear and movie rendering draw calls.

In other words, with this fix Titan finally found his floor :)

### DIFF
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -1168,6 +1168,8 @@ void common_flip(struct game_obj *game_object)
 // clear_all below
 void common_clear(uint32_t clear_color, uint32_t clear_depth, uint32_t unknown, struct game_obj *game_object)
 {
+	if(gl_defer_clear_buffer(clear_color, clear_depth, game_object)) return;
+
 	uint32_t mode = getmode_cached()->driver_mode;
 
 	if(trace_all) ffnx_trace("dll_gfx: clear %i %i %i\n", clear_color, clear_depth, unknown);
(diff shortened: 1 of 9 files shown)


julianxhokaxhiu:
Ok the way the PR is being is going to 100% break the blitting logic in FF8 which was very hard to get working. So I'll not be able to merge this soon as it needs deep testing on FF8.

@myst6re would you be able to help here? I'll try to as well on my own side, but having a pair of eyes more wouldn't hurt. Thanks!

myst6re:
Yes, tonight! (UTC+2)

CosmosXIII:
> Ok the way the PR is being made, is going to 100% break the blitting logic in FF8 which was very hard to get working. So I'll not be able to merge this soon as it needs deep testing on FF8.
> 
> @myst6re would you be able to help here? I'll try to as well on my own side, but having a pair of eyes more wouldn't hurt. Thanks!

Actually I put an if statement to check if it is running FF7 and lighting is enabled in all the newly added functions for deferring draw calls. For example here:
https://github.com/julianxhokaxhiu/FFNx/pull/478/commits/64bacb470471c3333c8fcfe3b5266d74ec95a4ff#diff-4d7701eea9c0aaa722960664de09f03f20a80841aa17a61e48422147de61fff2R142

So for FF8 it should return false and work as before.

julianxhokaxhiu:
Thanks, good to know. If that is enough then we should not notice anything :)

julianxhokaxhiu:
I went through the final boss battle in FF8 which makes extensive usage of blitting, as well as the battle whirl, and I didn't notice any degradation. Well done @CosmosXIII! Let's wait for an ACK from myst as well :)

julianxhokaxhiu (approved):

julianxhokaxhiu:
Can you please rebase? We're good for merging this :)

CosmosXIII:
> Can you please rebase? We're good for merging this :)

Done!

julianxhokaxhiu:
Thank you!

PR #481, opened 2022-08-14
Widescreen: Added config logic and zoom implementation

CosmosXIII:
## Summary

This PR adds the logic to load and use a widescreen config file and implements a zoom logic to fill the 16:9 screen for fields that are not wide enough to extend all the way to 16:9. The changes of this PR are behind checks so that it only affects FF7.

Below is the work contributed by vertex:
- Widescreen: Add hook function just before draws related to text and text boxes

### Motivation

This will allow to increment the number of fields that can be displayed in widescreen mode and therefore improve consistency.
This is optional so when no config file is provided it will work as before.

### ACKs

### DIFF
--- a/misc/FFNx.blit.frag
+++ b/misc/FFNx.blit.frag
@@ -1,0 +1,26 @@
+/****************************************************************************/
+//    Copyright (C) 2022 Cosmos                                             //
+//                                                                          //
+//    This file is part of FFNx                                             //
+//                                                                          //
+//    FFNx is free software: you can redistribute it and/or modify          //
+//    it under the terms of the GNU General Public License as published by  //
+//    the Free Software Foundation, either version 3 of the License         //
+//                                                                          //
+//    FFNx is distributed in the hope that it will be useful,               //
+//    but WITHOUT ANY WARRANTY; without even the implied warranty of        //
+//    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         //
+//    GNU General Public License for more details.                          //
+/****************************************************************************/
+
+$input v_texcoord0
+
+#include <bgfx/bgfx_shader.sh>
+#include "FFNx.common.sh"
+
+SAMPLER2D(tex_0, 0);
+
+void main()
+{
+	gl_FragColor = texture2D(tex_0, v_texcoord0.xy);
+}
--- a/misc/FFNx.blit.vert
+++ b/misc/FFNx.blit.vert
@@ -1,0 +1,32 @@
+/****************************************************************************/
+//    Copyright (C) 2022 Cosmos                                             //
+//                                                                          //
+//    This file is part of FFNx                                             //
+//                                                                          //
+//    FFNx is free software: you can redistribute it and/or modify          //
+//    it under the terms of the GNU General Public License as published by  //
+//    the Free Software Foundation, either version 3 of the License         //
+//                                                                          //
+//    FFNx is distributed in the hope that it will be useful,               //
+//    but WITHOUT ANY WARRANTY; without even the implied warranty of        //
+//    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         //
+//    GNU General Public License for more details.                          //
+/****************************************************************************/
+
+$input a_position, a_texcoord0
+$output v_texcoord0
+
+#include <bgfx/bgfx_shader.sh>
+
+void main()
+{
+    vec4 pos = a_position;
+
+    pos.w = 1.0 / pos.w;
+    pos.xyz *= pos.w;
+    pos = mul(u_proj, pos);
+
+    gl_Position = pos;
+    v_texcoord0 = a_texcoord0;
+}
+
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -55,6 +55,8 @@
 #include "achievement.h"
 #include "game_cfg.h"
 
+#include "ff7/widescreen.h"
+
 #include "ff8/vram.h"
 #include "ff8/vibration.h"
 
@@ -835,6 +837,8 @@ int common_create_window(HINSTANCE hInstance, struct game_obj* game_object)
 					vram_init();
 					vibration_init();
 				}
+
+				if(aspect_ratio == AR_WIDESCREEN) widescreen.init();
 
 				// enable verbose logging for FFMpeg
 				av_log_set_level(AV_LOG_VERBOSE);
--- a/src/ff7/field.cpp
+++ b/src/ff7/field.cpp
@@ -586,48 +586,58 @@ void field_clip_with_camera_range_float(vector2<float>* point)
 {
 	field_trigger_header* field_triggers_header_ptr = *ff7_externals.field_triggers_header;
 	float half_width = 160;
-	if(is_fieldmap_wide())
+	auto camera_range = field_triggers_header_ptr->camera_range;
+
+	if(aspect_ratio == AR_WIDESCREEN && (is_fieldmap_wide() || widescreen.getMode() == WM_ZOOM))
 	{
-		int cameraRange = field_triggers_header_ptr->camera_range.right - field_triggers_header_ptr->camera_range.left;
-		half_width = ceil(wide_viewport_width / 4.f);
+		camera_range = widescreen.getCameraRange();
+
+		// This centers the background if necessary
+		int cameraRangeSize = camera_range.right - camera_range.left;
+		half_width = 160 + std::min(53, cameraRangeSize / 2 - 160);
 	}
 
-	if (point->x > field_triggers_header_ptr->camera_range.right - half_width)
-		point->x = field_triggers_header_ptr->camera_range.right - half_width;
-	if (point->x < field_triggers_header_ptr->camera_range.left + half_width)
-		point->x = field_triggers_header_ptr->camera_range.left + half_width;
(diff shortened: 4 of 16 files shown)


src/ff7/widescreen.cpp
```diff
@@ -24,12 +24,28 @@
...
+// This function should be called at each frame after drawing backgrounds and 3d models
+void ff7_field_draw_gray_quads_sub_644E90() {
+    ff7_externals.field_draw_gray_quads_644E90();
+
+    if(gl_defer_zoom()) return;
+
+    if(aspect_ratio == AR_WIDESCREEN && widescreen.getMode() == WM_ZOOM)
+        newRenderer.zoomBackendFrameBuffer();
+
+    ffnx_trace("FIELD: draw gray quads\n");
```
tangtang95, line 46: I left this trace here, but it was meant to be deleted

src/ff7/widescreen.cpp
```diff
@@ -230,4 +248,67 @@ void ff7_widescreen_hook_init() {
...
+    if(pName == 0) return;
+
+    std::string _name(pName);
+    auto node = config[_name];
+    if(node)
+    {
+        camera_range.left = node["left"].value_or(0);
+        camera_range.right = node["right"].value_or(0);
+        camera_range.bottom = node["bottom"].value_or(0);
+        camera_range.top = node["top"].value_or(0);
```
tangtang95, line 303: You are overwriting all camera ranges. So if there is no config.toml, it's not gonna work at all (everything becomes 0). I think it's better to overwrite only the camera ranges that are inside the toml file.

src/renderer.cpp
```diff
@@ -1181,18 +1189,18 @@ void Renderer::setScissor(uint16_t x, uint16_t y, uint16_t width, uint16_t heigh
...
         // This removes the black bars on the top and bottom of the screen
-        if(y = 16 && height == 448)
+        if(y == 16 && height == 448)
         {
             scissorOffsetY = getInternalCoordY(0.0);
             scissorHeight = getInternalCoordY(480);
         }
 
         // This sets a scissor offset for field with width not enough to fill the screen in 16:9
-        field_trigger_header* field_triggers_header_ptr = *ff7_externals.field_triggers_header;
```
tangtang95, line 1191: Better to rewrite this section. I would do it like this:
```cpp
struct game_mode* mode = getmode_cached();
switch(mode) {
   case MODE_CREDITS:
       // modify scissorOffsetX only
    case MODE_FIELD:
       // is_movie_playing
      
       // remove black bars

      // widescreen.getMode(), In your code I saw == WM_DISABLED, is it correct?
   default:
      // maybe do not apply scissorOffsetX shift only if x == 0 and game_width == 640 or find another way.
}

```

src/ff7/field.cpp
```diff
@@ -586,48 +586,58 @@ void field_clip_with_camera_range_float(vector2<float>* point)
...
+
+	if(aspect_ratio == AR_WIDESCREEN && (is_fieldmap_wide() || widescreen.getMode() == WM_ZOOM))
 	{
-		int cameraRange = field_triggers_header_ptr->camera_range.right - field_triggers_header_ptr->camera_range.left;
-		half_width = ceil(wide_viewport_width / 4.f);
+		camera_range = widescreen.getCameraRange();
+
+		// This centers the background if necessary
+		int cameraRangeSize = camera_range.right - camera_range.left;
+		half_width = 160 + std::min(53, cameraRangeSize / 2 - 160);
```
tangtang95, line 597: This doesn't seem to work well with fields that are already big enough. Half width shouldn't be depending on camera range. If camera range is 660, you are going to set `half_width` to 340, which is obviously going outside the background

tangtang95: Oh my bad, it's the minimum of those two. But this can also be done at config.toml level by increasing the camera range right and left to 213 and -213

julianxhokaxhiu: Do we need some more work here then?

src/ff7/widescreen.cpp
julianxhokaxhiu: Well in this case if there's or not a config toml, the `if` guard will definitely not allow the code to be executed. Although, if there's at least one node defined, but only with one property, for eg.:
```toml
[field_name]
left = 123
```
In this case all the other fields will be filled with 0, even though they were not defined.

If this is the intended behavior, fine by me, if not, please consider checking if the field exists while reading it to know if you have to override it or not.

src/renderer.cpp
```diff
@@ -233,6 +233,8 @@ void Renderer::updateRendererShaderPaths()
     fragmentShadowMapPath += ".smooth" + shaderSuffix + ".frag";
     vertexFieldShadowPath += ".smooth" + shaderSuffix + ".vert";
     fragmentFieldShadowPath += ".smooth" + shaderSuffix + ".frag";
+    vertexBlitPath += ".smooth" + shaderSuffix + ".vert";
+    fragmentBlitPath += ".smooth" + shaderSuffix + ".frag";
```
julianxhokaxhiu, line 237: Technically it should not matter because when we blit we don't use the vertex colors at all, but shouldn't this technically be `.flat` instead?

src/renderer.cpp
julianxhokaxhiu: I agree, it makes the code much more cleaner.

misc/FFNx.blit.frag
```diff
@@ -0,0 +1,27 @@
...
+
+#include <bgfx/bgfx_shader.sh>
+#include "FFNx.common.sh"
+
+SAMPLER2D(tex_0, 0);
+
+void main()
+{
+	vec4 color = texture2D(tex_0, v_texcoord0.xy);
+	gl_FragColor = color;
```
julianxhokaxhiu, line 26: You can optimize this by simply doing
```glsl
gl_FragColor = texture2D(tex_0, v_texcoord0.xy);
```
GPU drivers will be much happier too.

julianxhokaxhiu (changes requested):
I left a bunch of comments here and there. Some parts don't really need an explicit change, but others do. Feel free to explore each conversation to know which one.

If there's a need to talk about any of this, let's do it on Discord.

src/ff7/widescreen.cpp
CosmosXIII: Deleted the trace, thanks.

src/ff7/widescreen.cpp
CosmosXIII: Added the `if` tests to check whether each of the parameters exists or not.

src/renderer.cpp
CosmosXIII: Rewrited this section with a switch statement and changed some conditions to fix known bugs.
As it is now framebuffer effects work fine and the limit break boxes are now fixed too!

src/ff7/field.cpp
CosmosXIII: It could be done in the config manually, but I think the centering it is better to do it through code because it would reduce the size of the config file and it would be easier to adapt it for other aspect ratios such as 16:10 in the future.

src/renderer.cpp
CosmosXIII: Changed it to `.flat`

misc/FFNx.blit.frag
CosmosXIII: Done! GPU drivers happy now :)

src/ff7/field.cpp
tangtang95: For now, bigger aspect ratio such as 21:9 are not compatible in few spots. The hardest to fix is worldmap (i.e. clouds textures).

julianxhokaxhiu: Ok let's keep this as TODO for the future. But for now, we're good :)

misc/FFNx.blit.vert
```diff
@@ -0,0 +1,33 @@
...
+{
+    vec4 pos = a_position;
+    vec2 coords = a_texcoord0;
+
+    pos.w = 1.0 / pos.w;
+    pos.xyz *= pos.w;
+    pos = mul(u_proj, pos);
+
+    gl_Position = pos;
+    v_texcoord0 = coords;
```
julianxhokaxhiu, line 31: Sorry missed this one...
```glsl
v_texcoord0  = a_texcoord0
```
Thanks!

CosmosXIII: Done!

julianxhokaxhiu (approved):

src/common.cpp
```diff
@@ -954,6 +958,8 @@ void common_flip(struct game_obj *game_object)
 	time_t last_seconds = last_frame.time;
 	struct game_mode *mode = getmode_cached();
 
+	if(aspect_ratio == AR_WIDESCREEN) widescreen.initParamsFromConfig();
```
tangtang95, line 961: This seems an overkill. There is a better place to put this. There should be an initialize variables in ff7/field.cpp.

tangtang95 (changes requested):

src/common.cpp
tangtang95: That function is called at the start of field mode. But I'm not sure if the camera range are already available. Need to test it

CosmosXIII: it worked fine, thanks

julianxhokaxhiu:
Thank you!

julianxhokaxhiu:
Patch has been dropped because of many issues on default 4:3 AR. Feel free to open another PR.

PR #482, opened 2022-08-15
Widescreen: Added config logic and zoom implementation

CosmosXIII:
## Summary

This PR adds the logic to load and use a widescreen config file and implements a zoom logic to fill the 16:9 screen for fields that are not wide enough to extend all the way to 16:9. The changes of this PR are behind checks so that it only affects FF7.

Below is the work contributed by vertex:
- Widescreen: Add hook function just before draws related to text and text boxes

### Motivation

This will allow to increment the number of fields that can be displayed in widescreen mode and therefore improve consistency.
This is optional so when no config file is provided it will work as before.

### ACKs


julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #486, opened 2022-08-27
FF7: Move vertical center hext into FFNx code as toml config option

tangtang95:
## Summary

Move the  vertical center hext into toml config option for FF7. 

### Motivation

In this way, we can use the same code to enable it by default for widescreen. Otherwise if there is also this flag to consider in widescreen, the code becomes more complex.

### ACKs


tangtang95:
I have only moved two hext address patches because the third one is useless

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #488, opened 2022-09-02
Refactor FF7 field.cpp + 60FPS fix for partial animation opcode

tangtang95:
## Summary

Huge refactor on `src/ff7/field.cpp`: refactor FF7 field.cpp into multiple files

Created a new namespace for all the files related to field.

The new list of modules/files is:
 - background: for field background related stuff
 - enter: related to `field_enter` sub run only once when transitioning to FIELD mode
 - model: for field models related stuff
 - opcode: related to field opcodes
 - utils: common or generic utils related to multiple modules
 - defs: functions that are exported outside of ff7_field namespace
 - field: main file that contains unrelated stuff to the other files and the hook init function

There is also a 60FPS fix I had done related to partial animation opcode: it is just a way to align better passing from one animation to the next one.

### Motivation

`field.cpp` has become huge, and quite hard to follow. So, might as well refactor everything into multiple files so it's easier also to understand.

### ACKs


julianxhokaxhiu (approved):

julianxhokaxhiu:
Wow, huge PR but definitely worth it! Thank you :)

PR #489, opened 2022-09-03
FF7: Refactor battle, field, and world modules

tangtang95:
## Summary

Refactor the battle files in a similar way than field files: into folders and using a namespace. The same also for world modules.
Moreover, changed the field namespace into `ff7::field`

### Motivation

It was a mess before. Not that with this refactor it won't be :)

### ACKs


julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #494, opened 2022-09-11
FF7: Improve texture loading speed and caching

tangtang95:
- Fix default handle index for DDS texture create (use FFNx one which returns 0 if not found)
 - Remove as much as possible the number of reads on the texture file (Primarily the `stat` function check)
 - Overhaul the texture loading by dividing clearly the animated texture loading and the normal texture loading.

## Summary

Performance improvement for loading textures. The goal is to decrease the number of reads to the file. Instead of verifying the file existence, immediately try to load it. This reduce the number of reads on the file and it will return 0 if the texture file hasn't been loaded.

Also moved up the check for caching for animated texture. Also once an animated texture has been resolved (even with missing texture), this is saved on the hash table cache.

### Motivation

Due to performance issue in 7th with missing textures (e.g. basketball minigame) and cold start on some field map.

### ACKs

- [x] I test it on 7th new overhaul and also with 7th caching disabled (the one on Wrap.cs)
- [ ] I only tested it with 60 fps mod + SYW texture mod. We still need to test it with other mods. Also if someone else could try to check the performance improvement, it would be helpful

### DIFF
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -1326,7 +1326,7 @@ void common_unload_texture(struct texture_set *texture_set)
 	// Destroy animated textures
 	if (gl_set->is_animated)
 	{
-		for(std::map<std::string,uint32_t>::iterator it = gl_set->animated_textures.begin(); it != gl_set->animated_textures.end(); ++it) {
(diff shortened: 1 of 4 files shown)


julianxhokaxhiu (approved):

julianxhokaxhiu:
Nice one and well thought :) Thank you!

PR #498, opened 2022-09-15
Revert animated_textures key to string and use key based on partial path

tangtang95:
## Summary

Revert back animated_textures map key to string because there are conflicts by using only the hash. 

### ACKs


julianxhokaxhiu (approved):

julianxhokaxhiu:
Amazing! Thank you

PR #501, opened 2022-09-25
Add clear depth on BATTLE mode before drawing UI (Related #131)

### LINKED ISSUE #131: [ FF7 ] Scorpion  Sentinel drawn over UI
Yeah, the first boss.
It's the original Steam release (1.0.9), FFNx 1.8.0, and the only changed settings are:
`renderer_backend = 0`
`fullscreen = true`
`preserve_aspect = true`
`enable_vsync = true`
`linear_filter = false`
`enable_antialiasing = 0`
`enable_anisotropic = true`

And, just in case, important PC specs: Windows 10 2004 x64, NVIDIA GTX 460 (driver version 391.35).

tangtang95:
## Summary

 - Clear the frame buffer depth
 - Add `bgfx::resetView` on all used viewId at each frame to avoid view collision at the next frame to allow to have dynamic number of views at each frame.
 - Change logic to select battle UI elements to defer. Use a more clear condition.

### ACKs

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,10 @@
 
 - Full commit list since last stable release: https://github.com/julianxhokaxhiu/FFNx/compare/1.14.0...master
 
+## FF7
+
+- Renderer: Fixed graphical glitch happening in battle when 3d models rendered in front of UI (https://github.com/julianxhokaxhiu/FFNx/issues/131)
+
 # 1.14.0
 
 - Full commit list since last stable release: https://github.com/julianxhokaxhiu/FFNx/compare/1.13.0...1.14.0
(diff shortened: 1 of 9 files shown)


tangtang95:
There are some draw text that needs to be deferred

tangtang95:
I have modified the condition behind deferring battle UI elements. Before it wasn't really clear what it was deferring. Now, we choose only 2 draw related to 2 graphics object to defer which are necessary for ESUI. Other graphics objects to defer doesn't seem necessary. If we find some bug, we can easily add it.

julianxhokaxhiu (dismissed):

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #502, opened 2022-09-25
Voice: Add play voice for enemy actions during BATTLE mode

tangtang95:
## Summary

Add play voice for enemy actions in BATTLE mode. The file requested is `voice/_battle/enemy_XXXX/cmd_YY_ZZZZ.ogg` where `XXXX` is the enemy ID of the battle, `YY` is the command ID of the enemy, and `ZZZZ` is the action ID of the enemy. 

 - enemy ID number is in in HEX format
 - YY should be always 20. It's the command ID for enemy action

### Motivation

Echo-S mod seems to need it

### ACKs

### DIFF
--- a/src/voice.cpp
+++ b/src/voice.cpp
@@ -183,7 +183,7 @@ bool play_battle_cmd_voice(byte char_id, cmd_id command_id, std::string tokenize
 	return playing;
 }
 
-bool play_battle_action_voice(byte char_id, byte command_id, short action_id)
+bool play_battle_char_action_voice(byte char_id, byte command_id, short action_id)
 {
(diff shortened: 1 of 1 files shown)


tangtang95:
Maybe command ID can be drop since it should be the same for all enemies

tangtang95:
> Maybe command ID can be drop since it should be the same for all enemies

Actually it's better to leave it so, there is a way to easily play a voice for whatever action of the enemy.

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thanks!

PR #516, opened 2022-12-29
60FPS: Fix Climhazzard random crash on poisoned enemy

tangtang95:
## Summary

With 60 FPS mod, when using Climhazzard against enemy with status (e.g. poison), can cause a crash. This crash does not happen always, but somehow "randomly". I do not know the real cause, but I just disabled any decorator wrapper on the battle effect function that cause the crash. After testing it more than 10 times, I didn't get any crash.

### Motivation

A user reported it.

### ACKs


julianxhokaxhiu (approved):

julianxhokaxhiu:
Can you please update also the Changelog?

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #518, opened 2022-12-30
FF7: Added logic to update a month string variable for the time cycle system

CosmosXIII:
## Summary

This PR adds the logic to update a month string variable for the time cycle system.

### Motivation

This allows the month string update to be managed by FFNx instead of requiring a field script to update. It also allows it to be updated when in other modes such as battle and worldmap.

### ACKs


julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #508, opened 2022-10-30
FF7: Fix 3D model drawn over UI on BATTLE mode if lighting enabled

CosmosXIII:
Fixes #131

## Summary

The same fix as below but for lighting enabled:
https://github.com/julianxhokaxhiu/FFNx/pull/501

### Motivation

The motivation is to make Kuraudo happy. The expected outcome is Kuraudo being happy.

### ACKs

- [ ] I have updated the [Changelog.md](https://github.com/julianxhokaxhiu/FFNx/blob/master/Changelog.md) file -> changelog was already changed by vertex
- [x] I did test my code on FF7
- [ ] I did test my code on FF8

PR #525, opened 2023-01-11
FF7: Fix FIELD mode text box flickering rendering

tangtang95:
## Summary

Apply bgfx reset view only when it is FF7 and in battle mode.

### Motivation

The reset view caused a regression bug on FIELD mode where some text box when closed caused flickering. By applying the reset view only in battle mode, we avoid the issue and still keep this fix https://github.com/julianxhokaxhiu/FFNx/pull/508

### ACKs

### DIFF
--- a/src/renderer.cpp
+++ b/src/renderer.cpp
@@ -1098,8 +1098,10 @@ void Renderer::show()
 
     bgfx::dbgTextClear();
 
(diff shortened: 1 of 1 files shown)


julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #534, opened 2023-02-03
Video: Fix incorrect gamma for FMV playback

ChthonVII:
## Summary

Add SMPTE170M linearization function and use it when YUV texture is a movie.
FMVs are almost certain to be bt601 or bt709, so we need this to properly convert them to linear RGB.

### Motivation

Why are you doing this? FMVs should play back correctly.

### ACKs

**I HAVE DONE NONE OF THESE THINGS.** I don't have a Windows build environment set up for this right now. Sorry.


julianxhokaxhiu:
@CosmosXIII please review as well. Thank you!

CosmosXIII:
I'm not familiar with this process but my understanding is it was being converted first from YUV to sRGB, then from sRGB to linear.
Does it need to do this too in this case?
Also I know you said you don't have the environment to build it but some before/after screenshots would be great.

CosmosXIII:
I tested this branch with SYW upscales but the blacks are still crushed:
image

ChthonVII:
I was able to get a build environment up and test this. It appears to be functioning as intended.

julianxhokaxhiu:
@ChthonVII make sure you're not using HDR for testing, because if that's the case, you have to look for [something else](https://github.com/julianxhokaxhiu/FFNx/blob/master/misc/FFNx.post.frag#L38-L41) to patch in the entire flow.

CosmosXIII:
what I can say is that the movies definitely look better in Windows Media Player or VLC. But I'm not getting the same quality with this PR, the blacks are still crushed with SYW.

ChthonVII:
I was able to get a build environment up and running.

This commit appears to be working as intended.
Blacks are still squashed because of #535. (Also it looks like maybe you forgot to flush your shader cache. I forgot the first time I tested it.)

This fix + fixing #535 + correcting some slightly off entries in the YUV->RGB matrix result in something that looks like correct _TV-range_ playback.
Example:
TVrangefixed

The showstopping problem now is that [swscale won't do a TV->PC range conversion with 10bit input](https://github.com/julianxhokaxhiu/FFNx/issues/536#issuecomment-1417960332) (like SYW's upscales), so we'll have to detect that case and do it ourselves.

julianxhokaxhiu:
Out of curiosity, did you check how does `mpv` for eg. handles this on their side? They do render as well using the GPU so I'd be curious if we could align to them on that regard.

CosmosXIII:
> This fix + fixing #535 + correcting some slightly off entries in the YUV->RGB matrix result in something that looks like correct _TV-range_ playback. 

Ok, can you commit those changes too so we can test it?

ChthonVII:
I'm afraid I made a mistake here. Using the SMPTE170M would be correct _for the original videos_, but it's not correct for SYW's AI upscales. The correct inverse gamma function to use for those is the inverse of whatever gamma function his ERGSAN upscaler thingy uses. I don't know what that is. But it's clearly not SMPTE170M, since it gives use posterization on the star field. The sRGB function looks right.

This raises a bothersome issue: The video files that everyone's using are out of spec. You're not supposed to use BT601 with some other transfer function. I guess the best thing to do is to use the sRGB function and document that we're expecting out-of-spec videos.

Where I am now:
- Reverting this and going back to sRGB inverse gamma function +
- Fix #535 +
- YUV->RGB matrix fixes +
- modify movies.cpp to _NEVER_ attempt a range conversion; just pass through and tell the shader what the range is
=
- correct output for 8-bit TV-range yuv420 (range conversion is handled in the YUV->RGB matrix)
- correct output for 8-bit PC-range yuvj420
- too dark for 10-bit anything. Swscale is doing something wrong to the color range when it converts down to 8-bit. My only guess it's doing an unrequested tv->pc range conversion, and then the shader is doing another one on top. I'll investigate that tomorrow.

ChthonVII:
The plot thickens:
* When configured for pixel format conversion with no range conversion, swscale is nevertheless doing a TV->PC range conversion on SYW's file (10-bit, TV-range, but _missing the range metadata_). If we make movies.cpp tell the shader that swscale's output is full range, we get correct output.
* When configured for pixel format conversion with no range conversion, swscale is nevertheless doing **_TWO_** TV->PC range conversions if the file is 10-bit, TV range, and has correct metadata. What. The. Hell?
* When configured for pixel format conversion with no range conversion, swscale is nevertheless doing a TV->PC range conversion if the file is 10-bit, PC range, and has correct metadata.

So, the good news is that correct playback for the files that most people are using is in reach.
The bad news is that the behavior with files with correct metadata is really, really wrong. I'll experiment with swscale's parameters in hopes of finding some combination that doesn't trigger this behavior.

eve-atum:
Before you go tearing your hair out trying to solve all cases, keep in mind that the people creating these upscales are hobbyists and fans doing this in their spare time using off-the-shelf software, and likely have minimal (or zero) domain knowledge.  So rather than building your solution around their (likely default) software settings, I would suggest building it to handle things correctly in both the OG format (25 year old AVI files), and capable of reading the metadata attached to anything that comes out in the future.  Satsuki Yatoshi (creator of the SYW upscales) is an active member of the community, and if you provide the tools and knowledge, it'll be easy for him (and anyone else) to do things the "right way" on the next iteration.

ChthonVII:
@eve-atum,
Ironically swscale's behavior is buggier when the file has correct metadata. Doing some things SYW's "wrong" way may be the only way to work around these bugs for 10-bit inputs.

eve-atum:
I understand where you're coming from, but the proper fix is to get ffmpeg's stack to behave correctly, rather than working around it.  Being FOSS, we can be fairly confident that someone over there is going to realize it's messed up (perhaps you've already flagged it to them?) and fix their bug, which will then turn your workaround into a new bug on our side once we bump the dependency with the upstream repair.

ChthonVII:
I believe I've got it working. Posted more of a write up to [the consolidated issue thread](https://github.com/julianxhokaxhiu/FFNx/issues/536#issuecomment-1421043470).

@eve-atum: I don't foresee the ffmpeg people fixing this without a total rewrite of swscale, which I don't think they have the capacity to do any time soon, or likely ever.

The root of the problem is that swscale is guessing what color range conversion is needed based on metadata and pixel format, and then automatically doing it. But it's often guessing wrong. On top of that, sometimes it does _two_ conversions in the same direction (which ruins the video). On top of that, in most cases it ignores the user-supplied parameters that are supposed to control the color range conversion. So you generally can't prevent it from doing an incorrect conversion, or make it do one it doesn't already want to do. Once I finally figured this out, the fix is pretty straightforward: _Use a different target pixel format that doesn't trigger automatic color range conversions._ Then we can just do it in the shader, like we're already set up to do. There remains one case where the source pixel format is triggering an unwanted automatic conversion. Fortunately, however, that's also the one case where it will obey the supplied parameters telling it not to.

eve-atum:
Ouch.. fair 'nuff then.  I would think that with how many other projects depend on ffmpeg, that google and canonical, at the very least, would want to ensure it operates correctly.

ChthonVII:
Spoke too soon. Please see [here](https://github.com/julianxhokaxhiu/FFNx/issues/536#issuecomment-1425106361) and give thoughts.

misc/FFNx.frag
```diff
@@ -72,28 +74,42 @@ void main()
...
+            }
 
-            color.rgb = toLinear(color.rgb);
+            
+            if (is170MGamma){
+                color.rgb = saturate(toLinearSMPTE170M(color.rgb));
+            }
+            else {
+                color.rgb = saturate(toLinear(color.rgb));
+            }
```
julianxhokaxhiu, line 112: I think this block of changes ( other than the matrix transformation which now let's say has more precision in it ) I think is not correct. Are you using `saturate` because you want to reach the 2.4 gamma out of the 2.2, or am I missing something here?

misc/FFNx.lighting.frag
```diff
@@ -92,28 +94,41 @@ void main()
...
+                color.rgb = saturate(instMul(mpeg_rgb_transform, yuv));
+            }
 
-            color.rgb = toLinear(color.rgb);
+            if (is170MGamma){
+                color.rgb = saturate(toLinearSMPTE170M(color.rgb));
+            }
+            else {
+                color.rgb = saturate(toLinear(color.rgb));
+            }
```
julianxhokaxhiu, line 131: Same story here.

src/video/movies.cpp
```diff
@@ -57,7 +57,13 @@ uint32_t movie_frames;
...
 double movie_fps;
 double movie_duration;
-bool movie_jpeg_range = false;
+bool fullrange_input = false;
+bool okpixelformat = false;
+bool okcolorspace = false;
+bool yuvjfixneeded = false;
+bool use170Mgamma = false;
+bool isdeepandtv = false;
+AVPixelFormat targetpixelformat = AV_PIX_FMT_YUV444P;
```
julianxhokaxhiu, line 66: Do we need all these flags as global vars? Can we put some of them locally to the function where they are used?

src/video/movies.cpp
```diff
@@ -178,8 +184,92 @@ uint32_t ffmpeg_prepare_movie(char *name, bool with_audio)
...
+            break;
+        case AVCOL_TRC_SMPTE170M:
+        case AVCOL_TRC_BT709: // same as SMPTE170M
+        case AVCOL_TRC_IEC61966_2_4: // same as 709
+        case AVCOL_TRC_BT1361_ECG: // not quite sure on this one, but think it's the same curve
+            use170Mgamma = true;
+            if (trace_movies)
+            {
+                ffnx_trace("prepare_movie: SMPTE170M transfer function detected\n");
+            }
```
julianxhokaxhiu, line 243: Please compact this if to a one-liner. Thank you

src/video/movies.cpp
```diff
@@ -215,23 +305,64 @@ uint32_t ffmpeg_prepare_movie(char *name, bool with_audio)
...
+            int *coefs_in;
+            int srcRange, dstRange;
+            int brightness, contrast, saturation;
+            sws_getColorspaceDetails(sws_ctx, &coefs_in, &srcRange, (int**)&dummy, &dstRange, &brightness, &contrast, &saturation);
+            
+            /*
+            if (trace_movies){
+                ffnx_trace("prepare_movie: sws_getColorspaceDetails() detected srcRange %i, dstRange: %i\n", srcRange, dstRange);
+            }
+            */
```
julianxhokaxhiu, line 330: Same here, please remove the dead code if not meant to be there.

src/video/movies.cpp
```diff
@@ -215,23 +305,64 @@ uint32_t ffmpeg_prepare_movie(char *name, bool with_audio)
...
-    int dummy[4];
-    int srcRange, dstRange;
-    int brightness, contrast, saturation;
-    sws_getColorspaceDetails(sws_ctx, (int**)&dummy, &srcRange, (int**)&dummy, &dstRange, &brightness, &contrast, &saturation);
-    const int* coefs = sws_getCoefficients(SWS_CS_DEFAULT);
-    dstRange = srcRange = movie_jpeg_range = true; // this marks that values are according to yuvj
-    sws_setColorspaceDetails(sws_ctx, coefs, srcRange, coefs, dstRange, brightness, contrast, saturation);
+		// if we need a colorspace conversion, set it up here
+        // this would also be the place to set up color range conversion, if it worked -- which it doens't
+        if (!okcolorspace /*|| !fullrange_input*/ || yuvjfixneeded || isdeepandtv){
```
julianxhokaxhiu, line 319: If the flag is not used, feel free to remove it. Otherwise if this PR is a WIP, I'll convert it as a Draft.

src/video/movies.cpp
```diff
@@ -289,8 +420,12 @@ void ffmpeg_stop_movie()
...
 {
 	uint32_t upload_width = strides[num];
-	uint32_t tex_width = num == 0 ? movie_width : movie_width / 2;;
-	uint32_t tex_height = num == 0 ? movie_height : movie_height / 2;
+    // Use these for yuv444:
+    uint32_t tex_width = movie_width;
+	uint32_t tex_height = movie_height;
+    // Use these for yuv420:
+	//uint32_t tex_width = num == 0 ? movie_width : movie_width / 2;;
+	//uint32_t tex_height = num == 0 ? movie_height : movie_height / 2;
```
julianxhokaxhiu, line 428: Please remove the dead code if not meant to be used.

src/video/movies.cpp
```diff
@@ -325,11 +460,14 @@ void draw_yuv_frame(uint32_t buffer_index)
 
 	newRenderer.isMovie(true);
 	newRenderer.isYUV(true);
-	newRenderer.isFullRange(movie_jpeg_range);
-	gl_draw_movie_quad(movie_width, movie_height);
+	newRenderer.isFullRange(fullrange_input);
+	//newRenderer.isFullRange(true); // use this to test swscale's range conversion (or lack thereof) by forcing shader to assume full range
```
julianxhokaxhiu, line 464: Also here.

julianxhokaxhiu (changes requested):
I left an early review. The PR overall looks good. I left some first impressions so feel free to take a look at single comments. Please consider fixing your code indentation using the file current format ( either with spaces or tabs, we don't have it consistent yet but I'm aiming to do that soon, but don't mix and match ).

I'll finally move this PR as Draft and once you'll be happy with it we can re-review it again and have a test before merging it. So far thank you for your hard work, I can sense the passion you're putting into this and I look forward to get this merged :)

src/video/movies.cpp
ChthonVII: Indeed, they do not. I relocated those that could be.

src/video/movies.cpp
```diff
@@ -178,8 +184,92 @@ uint32_t ffmpeg_prepare_movie(char *name, bool with_audio)
...
+            if (trace_movies)
+            {
+                ffnx_trace("prepare_movie: SMPTE170M transfer function detected\n");
+            }
+            break;
+        default:
+            ffnx_error("prepare_movie: unsupported transfer (inverse gamma) function\n");
+			ffmpeg_release_movie_objects();
+			goto exit;
+    }
```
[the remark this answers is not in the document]
ChthonVII: Think I fixed it. Hard to tell because it looked correct in my editor to begin with.

src/video/movies.cpp
ChthonVII: Done

src/video/movies.cpp
```diff
@@ -215,23 +305,64 @@ uint32_t ffmpeg_prepare_movie(char *name, bool with_audio)
...
-			AV_PIX_FMT_YUVJ420P,
-			SWS_FAST_BILINEAR | SWS_ACCURATE_RND,
+			//AV_PIX_FMT_YUVJ420P,// don't use yuvj target pixformats. swscale goes crazy and does extra, un-asked-for TV->PC conversions 
+			targetpixelformat,
+            //SWS_FAST_BILINEAR | SWS_ACCURATE_RND, // might need to go back to this if performance issues are encountered. Unlikely, I hope.
+            SWS_LANCZOS | SWS_ACCURATE_RND | SWS_FULL_CHR_H_INT,
 			NULL,
 			NULL,
 			NULL
 		);
```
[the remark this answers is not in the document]
ChthonVII: Done

src/video/movies.cpp
ChthonVII: Done.

src/video/movies.cpp
ChthonVII: Done.

src/video/movies.cpp
ChthonVII: I was keeping it as context for the accompanying comment explaining why we're not simply requesting a color range conversion from swscale. I'll delete it and just expand the comment.

satsukiyatoshi:
For information:
My upscale i've used the jpsxdec software to extract fmv so my pack is accurate.
After extraction i upsclaed and transcoded to x264 8bits (yv12 colorspace)

ChthonVII:
@satsukiyatoshi: Would you mind sharing more about your workflow? I'd really like to identify where the limited TV-range and 2.2 gamma came from.

Is this correct?
1. Use jpsxdec to create .png sequence and audio.
2. Use [cupscale](https://github.com/n00mkrad/cupscale) to upscale? Or are you using your own tool?
3. Use avisynth to read in upscaled .png sequence, stabilize, serve frames. Would you mind sharing one of your avisynth scripts?
4. Frameserve from avisynth to x264 to encode.
5. Encode audio with ???
6. Mux audio and video with mkvtoolsnix.
7. Rename .mkv file to .avi.

As best I can tell (and I've spent a lot of time looking at them now), your latest set of 30fps files have the following properties:
* BT601 color matrix
* SMPTE-C color gamut
* limited TV-range (!)
* YUV420P10 pixel format (not 8-bit, not yv12)
* pure 2.2 gamma curve (!) (expected SMPTE 170M piecewise function)
* no metadata in the video stream

Of these properties, three are unexpected or problematic:
* The TV range is just wrong. I suspect this might have come from calling [avisynth's ConvertToYUV420() function](http://avisynth.nl/index.php/Convert#YUV420) after reading in the upscaled .png sequence. The default behavior includes a PC->TV range conversion. You need to use `matrix = "PC.601"` to get full-range bt601 yuv output.  
* The 2.2 gamma is really unexpected. One would normally expect to see the SMPTE 170M piecewise function paired with bt601. This was introduced at the last point in the workflow where linear RGB was converted to gamma space RGB. Unless you were using linear RGB somewhere in your avisynth script, I suspect it happened inside cupscale. (In which case we may not be able to fix it and may have to just live with it.)
     * (An unfortunate side effect of having to live with 2.2 gamma is that most tools won't be able to do a correct gamut conversion (because they'll assume the SMPTE 170M function), and therefore won't be able to do a correct color matrix conversion. So we must stay in bt601+SMPTE-C. I believe that maybe [fmtconv](https://amusementclub.github.io/fmtconv/doc/fmtconv.html#transfer) could do it via its `gcor` parameter, but I'm not sure.)
* The lack of metadata makes it hard to know how to correctly play these files back. x264 has some [parameters for setting the metadata](https://manpages.ubuntu.com/manpages/focal/man1/x264.1.html). Particularly, `--range`, `--colormatrix`, and  `--transfer` would be very helpful.
     * Unfortunately, x264 doesn't have a setting for `--transfer` that means "pure 2.2 gamma curve." (Because it's not in [the h264 spec](https://aomediacodec.github.io/av1-spec/av1-spec.pdf); see page 117.) So we're probably stuck with "unspecified" there.

satsukiyatoshi:
Here's my process for 30fps files:

-Use jpsxdec to create .png sequence and audio.
-Use ESRGan with my own command line tool to do some spacial / palette cleaning.
-Use avisynth to do some preprocessing (mainly temporal stabilisation) from rgb png to rgb png.
-Use ESRGan with my own command line tool to do upscaling.
-Use avisynth to do some postprocessing depending upscaled source from rgb png to raw avi rgb.
-Doubled fps with TopazLab video (Prores codec 4:4:2 as no raw provided by this software) or Flowframes (raw RGB avi).
-Use aviynth to convert source to yv12 and frameserve it to x264 8 bits version (superslow profile + crf 18)
-Encode audio with ffmpeg.
-Mux audio and video with mkvtoolsnix.
-Rename .mkv file to .avi

no 16-235 level tv limit used
no 10 bit x264 encoding done

i compared my psx jap's ff7 into fmv (real rgb cable) platinuim editon (if it matters) with my 30fps intro fmv played with mpc-hc + madvr on the same Tv (in computer mode, calibrated with no enhancement at all) and to me the colors/luma/contrasts look accurate.
i compared also on my benq w2700 (with a scart > hdmi converter) and the psx fmv looks sligthly darker thans my pack's one but it may be because of the converter.

ChthonVII:
@satsukiyatoshi: Are we talking about the same files here? I'm looking at the files unpacked from the "[Tsunamods] SYW V6 FMV (30fps)" iro. They are absolutely, beyond any doubt, TV-level and YUV420P10 pixel format. Look:
Is10bit
histogram

satsukiyatoshi:
don't know why it is reported this way, i told you the way i processed my 30 fps fmv pack.
if you own a psx with ff7, compare the playing from my pack with mpc-hc VS the playing of the psx and you'll see it's realy accurate.

ChthonVII:
> don't know why it is reported this way

I've got some pretty good guesses:
* Why YUV420 instead of YV12? Because the x264 encoder doesn't support YV12 output. If you feed it YV12 input, it will convert to YUV420. (Which is OK because that conversion is lossless.)
* Why 10-bit instead of 8-bit? Two possibilities:
     * You've got your encoder binary files crossed up. Maybe you accidentally used the unintended one at encode time. Maybe you crossed up the versions when you downloaded them. Maybe whoever supplied the downloads crossed them up when they uploaded them.
     * You've got an all-in-one binary. I don't know about Window, but x264 for Linux went to an all-in-one binary a few years back. It reproduces the bit-depth of its input unless you tell it otherwise.
* Using a 10-bit encoder instead of 8-bit is fine. Some people would even say it's a superior choice for all use cases. And they're probably right.
     * But it does mean that your cfr setting is unnecessarily low because the scale is different.
* What's in those extra 2 bits? Depends...
     * If you converted to 10-bit _before_ compressing the range, then it's about ~1.7 bits of original color data and ~0.3 bits of dithering noise. If the player uses the correct order of operations, the color data can be recovered at playback time.
     *  If you converted to 10-bit _after_ compressing the range, then it's all dithering noise.
* How did the range get compressed? Like I said before, some avisynth functions include pc->tv range compression _by default_, unless you expressly set a parameter saying otherwise. For (possibly relevant) example, if you just use `ConvertToYV12()` with no "matrix" specified, that's equivalent to `ConvertToYV12(matrix = "Rec601")`, **which includes a pc->tv range compression**. You need to expressly use `ConvertToYV12(matrix = "PC.601")` to avoid the range compression. (See [docs](http://avisynth.nl/index.php/Convert#YUV420)) If you'd like to send me your avisynth scripts, I'd be happy to hunt down the unwanted range conversion for you.
* Why does it look OK in mpc-hc despite the limited TV range? Because mpc-hc interprets the absence of color range metadata as "tv range," so it is expanding the range at playback time. This gets the colors that survived back in their proper places. **But you've still lost ~14% of the original color data, and you've got extra banding because of that.** 
     * Still don't believe it? Try editing the stream metadata to claim it's full range and watch what mpc-hc does. Or grab my branch and change movies.cpp line 452 from `newRenderer.isFullRange(fullrange_input);` to `newRenderer.isFullRange(true);` to see the input passed through without range expansion. These files are, beyond any question, limited tv-range videos.
     * After the unwanted range compression is removed, don't forget to inform the encoder to set the metadata with `--input-range pc --range pc`. Otherwise you'll get an incorrect extra expansion done at playback. 
* What about gamma? After reading your comments, and reviewing the jpsxdec source, and messing with jpsxdec a bit, I'm convinced that nothing in your workflow altered the gamma. The unexpected gamma is already present in the PS1 files. I probably need to do more to deal with that...

In sum, the only big problem on your end is the range compression. You've put so incredibly much work into preserving these videos accurately, but then you're losing 14% of your data to some undesirable avisynth default behavior.

satsukiyatoshi:
I won't do all the FMV again ^^'.
To my eyes it's more than enought accurate VS psx played fmv

ChthonVII:
> I won't do all the FMV again ^^'.
> To my eyes it's more than enought accurate VS psx played fmv

Please permit me to try one time to persuade you with a picture. Please save the picture below, zoom in nearest neighbor to 200% or 250% or so, and take a close look at the bottom of the mirrored chunk of wall in the middle. You'll see that the right side has a lot of banding, while the left side has much less. The right side is the file from the "[Tsunamods] SYW V6 FMV (30fps)" iro, with its range expanded like it would be for playback, flipped horizontally for easier comparison. The left side is my attempt at repairing it, re-encoded at crf 18. Most, maybe even all, of the banding like this was caused by the accidental range compression. While I'm pleasantly surprised how much banding my script was able to repair, this re-encode is not nearly as good as what you could accomplish by avoiding the range compression in the first place. Please reconsider.

sywviewopenfix_readback vpy - 3460

Here's a 3x nearest neighbor zoom of the very bottom:
bandingzoom3x

satsukiyatoshi:
As i say i won't because:
-ATM i have almost no free time
-I have also other projects (FF7 and not FF7 related)
-I have a familly
-In your exemple it's a x3 nearest neighbor picture, in motion it's almost invisible and as the fmv resize is not nearest either, it's smooth this banding too (and not speaking about most modern display tend to boost contrast to look better OOB and it's crushing lots of details and 99% user never set it up in an accurate way)

So yes it can be better (always) but the reality is that ingame i'm not sure that even 0.1% user will notice it.
I like the way you want to have the best possible quality and the best fidelity to the game, but i've done some many tweaks to each FMV, most FMV are a merge of several avs scripts (sometime i used a script for 2s to 3s) i haven't kept the whole process, script, models, filters.... so do it again would be from scratch, and as a gain is almost no visible in a real gameplay session ...

ChthonVII:
I see. Then I won't bother you about it again.

julianxhokaxhiu:
Just a quick review again on top of all the updates that have happened during this week:
- I noticed your code has still indentation issues, please check your changes here on GitHub to spot them ( https://github.com/julianxhokaxhiu/FFNx/pull/534/files )
- I noticed you added a new flag to allow users to customize the gamma for videos. Is that really necessary? I mean I'm not against it but for some videos that will run couple of minutes once in a while on an entire gameplay, is it really to push this complexity on them? Remember that this flag will then be required to be supported on 7th Heaven and good luck explaining everyone how to use it...can I suggest using the metadata if possible and otherwise fallback to a well known gamma value that is our internal safe default? I'd prefer NOT having this flag in place.
- I noticed you're still having commented code around. Please remember to remove that one once you'll be ready for a real review. We do have git for history, we don't need to keep commented code sections.

On top of all of this I just want to say thank you for your own dedication. I appreciate it and I look forward to the final implementation. Keep going!

ChthonVII:
Sigh. Another headache. Please see [the latest post in ye old 536](https://github.com/julianxhokaxhiu/FFNx/issues/536#issuecomment-1440217829).

julianxhokaxhiu:
From now all conversations will happen on #536 to keep this PR easier to understand.

misc/FFNx.toml
```diff
@@ -271,6 +271,14 @@ ffmpeg_video_ext = "avi"
 #~~~~~~~~~~~~~~~~~~~~~~~~~~~
 external_movie_audio_ext = "ogg"
 
+#[MOVIE DEFAULT GAMMA]
+# What gamma to use for movies if the file's metadata doesn't specify one.
+# Larger values are darker.
+# A value of ~2.0 approximately complies with NTSC broadcast television standards of the Playstation 1 era.
+# A value between ~2.2 and ~2.4 approximates the actual behavior of typical CRT television sets from that era.
+#~~~~~~~~~~~~~~~~~~~~~~~~~~~
+movie_default_gamma = 2.2
```
julianxhokaxhiu, line 280: As I stated I don't like the fact we have a flag for this part. Please consider removing it and defining a well known standard that we use as default fallback, if any other approach doesn't work ( metadata, guessing, etc. ). Thank you.

src/video/movies.cpp
julianxhokaxhiu: Thanks

julianxhokaxhiu (changes requested):
Aside of the flag, everything else so far looks good and safe enough to be merged for me. Let's try to find a way to solve it if possible. Thank you!

PR #543, opened 2023-02-19
Common: Use Game loop to render the FFNx logo

myst6re:
## Summary

- Fix FFNx logo game freeze by moving it before the squaresoft logo using the same timing logic (the consequence is the publisher movie will be before the FFNx logo)
- Add the ability to skip the FFNx logo, like the squaresoft logo
- Fix jp version + hook cleaning

### ACKs

### DIFF
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -932,11 +932,6 @@ uint32_t common_init(struct game_obj *game_object)
 
 	proxyWndProc = true;
 
-	if (!ff8) {
-		// Small rendering loop to draw the FFNx logo before the game starts
-		drawFFNxLogo(game_object);
-	}
-
 	return true;
 }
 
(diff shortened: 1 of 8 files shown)


julianxhokaxhiu:
Weird how this is happening to you again. I clearly remember having fixed this on this commit: https://github.com/julianxhokaxhiu/FFNx/commit/08d80920a3ad14a6f278e616f08d30fd4626b759

Are you using the latest master status on your own end?

myst6re:
> Weird how this is happening to you again. I clearly remember having fixed this on this commit: [08d8092](https://github.com/julianxhokaxhiu/FFNx/commit/08d80920a3ad14a6f278e616f08d30fd4626b759)
> 
> Are you using the latest `master` branch code on your own end?

Yes, when I start the game without the publisher video (you can rename/remove the file publish.pak to skip this video), and then I try to close the game, it hang. Looking at the code, I can imagine that the game is waiting for the render init to finish before sending any events to Windows.

> I like the idea of skipping the logo with CTRL+S though, can we implement it generically also for FF7?

Honestly I don't know, I've done it for FF8 because I know well how pubintro loop works

julianxhokaxhiu:
A-ha ok, didn't consider that edge case. But is it a valid one? I'd assume the game to always be installed correctly, and as long as it is it will work fine. I'd prefer finding one way to handle this code instead of having to inject it in custom ways per game edition. Can't this code be independent of the game? that's the main purpose basically of it. Run an intro logo and then start the game.

Regarding the CTRL+S combo we could add it on the renderer loop directly I guess, maybe I can do some tests somewhere this week.

myst6re:
> A-ha ok, didn't consider that edge case. But is it a valid one? I'd assume the game to always be installed correctly, and as long as it is it will work fine. I'd prefer finding one way to handle this code instead of having to inject it in custom ways per game edition. Can't this code be independent of the game? that's the main purpose basically of it. Run an intro logo and then start the game.

I checked with the publisher movie, and even the freeze is less noticeable, it is still there.

myst6re:
I pushed the FF7 version, they look alike because in both cases I hook to the begin scene call to draw the logo.

julianxhokaxhiu:
Fantastic, thank you! Is it ready for merge? Or do you still have pending fixes?

myst6re:
> Fantastic, thank you! Is it ready for merge? Or do you still have pending fixes?

It is ready :)

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #545, opened 2023-02-20
FF8: Enhance field texture replacement

myst6re:
## Summary

Try to load texture files from `field/mapdata/{first two letters of the field name}/{field name}/{field name}_{texture number}`
if the image from `field/mapdata/{field name}/{field name}` is not found.

+ PMP (field effects) texture replacement support at `field/mapdata/{field name}/{field name}_pmp`
+ Tiny improvements for VRAM texture replacements
+ adding helper functions to check if a file exist and to measure the elasped time a functions (in microseconds)

### Motivation

Existing mods on Tonberry or Demastered can have a lot of work to adapt for the background replacement mecanism in FFNx.

### ACKs

### DIFF
--- a/src/utils.h
+++ b/src/utils.h
@@ -126,3 +126,15 @@ inline int getRandomInt(int min, int max)
 
     return distrib(gen);
 }
+
+inline std::chrono::time_point<std::chrono::high_resolution_clock> highResolutionNow()
+{
+    return std::chrono::high_resolution_clock::now();
+}
+
+inline long double elapsedMicroseconds(std::chrono::time_point<std::chrono::high_resolution_clock> startTime)
+{
+    return std::chrono::duration<long double, std::micro>(highResolutionNow() - startTime).count();
+}
+
+bool fileExists(const char *filename);
--- a/Changelog.md
+++ b/Changelog.md
@@ -15,7 +15,8 @@
 
 - Common: Fix startup hang on launch
 - Common: Fix jp version crash
-- Graphics: Add Field texture replacement ( https://github.com/julianxhokaxhiu/FFNx/pull/542 )
+- Graphics: Add Field texture replacement ( https://github.com/julianxhokaxhiu/FFNx/pull/542 https://github.com/julianxhokaxhiu/FFNx/pull/545 )
+- Graphics: Fix wrong texture replacements in battle
 
 # 1.15.0
 
--- a/src/audio.cpp
+++ b/src/audio.cpp
@@ -131,9 +131,7 @@ bool NxAudioEngine::getFilenameFullPath(char *_out, T _key, NxAudioEngineLayer _
 
 bool NxAudioEngine::fileExists(const char* filename)
 {
-	struct stat dummy;
-
-	bool ret = (stat(filename, &dummy) == 0);
+	bool ret = ::fileExists(filename);
 
 	if (!ret && (trace_all || trace_music || trace_sfx || trace_voice || trace_ambient))
 		ffnx_warning("NxAudioEngine::%s: Could not find file %s\n", __func__, filename);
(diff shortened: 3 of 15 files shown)


myst6re:
Found a bug

myst6re:
It seems that the bug is already present in master and it not introduced by this Pull Request, so I guess it is ready to merge, I'll create an issue.

src/utils.h
```diff
@@ -126,3 +126,18 @@ inline int getRandomInt(int min, int max)
...
+
+inline long double elapsedMicroseconds(std::chrono::time_point<std::chrono::high_resolution_clock> startTime)
+{
+    return std::chrono::duration<long double, std::micro>(highResolutionNow() - startTime).count();
+}
+
+bool fileExists(const char *filename)
+{
+    return GetFileAttributes(filename) != INVALID_FILE_ATTRIBUTES;
+}
```
julianxhokaxhiu, line 143: Do we really need this help? Atm we are using [`stat()`](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/stat-functions?view=msvc-170) everywhere as I think it's the fastest call we can make on the OS to check if a file exists. I'd be happy to port all the [`stat()`](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/stat-functions?view=msvc-170) calls into a unique utility function but we need to verify which is the fastest call ( and compatible one with 7th too ).

myst6re: I can use stat, no problem. But is there benchmarks of this? Stat should use GetFilesAttributes in its implementation, so I don't think it is faster.

julianxhokaxhiu: So when I started to use it around, this was my ref thread: https://stackoverflow.com/a/12774387

But looks like we might want to use the native `std::filesystem::exists` way as it's meant to be exactly for that purpose, so we should hope that optimizations done at that level are smart enough to be "the fastest" option.

I'm searching on the `ucrt` source code for what `stat` uses but atm I can't find any trace of `GetFileAttributes` which seems to be used by the [`access()`](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/access-waccess?view=msvc-170) API.

myst6re: Your reference does not have a benchmark on Windows. For me "stat()" is a POSIX call, Microsoft implemented it on top of its own syscalls, implying an overhead, with CreateFileW as you found.

Do you think if I use stat in fileExists, I could replace all file exists helpers in FFNx?

julianxhokaxhiu: I think so yes, and it should keep the compatibility with 7th too.

julianxhokaxhiu: One more update: `std::filesystem:exists` seems to use the [`GetFileAttributesExW`](https://github.com/microsoft/STL/blob/main/stl/src/filesystem.cpp#L872) API so this would break 7th compatibility as well. We could adopt it, but we would need to ship a new stable 7th version that hooks this API, test that everything works as expected and then finally be able to ship an FFNx stable round. It is doable but it requires some effort.

So for now going with `stat` is the only safe way to ensure everything we build so far continues to work.

src/utils.h
```diff
@@ -126,3 +127,21 @@ inline int getRandomInt(int min, int max)
...
+{
+    return std::chrono::duration<long double, std::micro>(highResolutionNow() - startTime).count();
+}
+
+bool fileExists(const char *filename)
+{
+    struct stat dummy;
+
+    // Use stat to keep compatibility with 7th Heaven
+	return stat(filename, &dummy) == 0;
```
julianxhokaxhiu, line 146: Can you fix the indentation here please? :D

myst6re: sure

myst6re:
Ready, and #547 should be fixed too

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #559, opened 2023-04-01
Lighting: Disable lighting texture for STAGE02_T06

AxlRose-RX:
I added "STAGE02_T06" (new texture) to my version of Grasslands, but would help a lot to not have to include a  config.toml in my .iro if by default this file had the extra/new texture excluded, is part of the skybox

## Summary

Brief explanation of the pull request.

### Motivation

Why are you doing this? What use cases does it support? What is the expected outcome?

### ACKs


julianxhokaxhiu:
@CosmosXIII Can you please review? Thanks!

CosmosXIII:
Why did you need to have an extra texture? why not just replace the skybox texture that was already there?

julianxhokaxhiu:
Re-reading the PR description actually I think the PR itself doesn't make much sense. It's a new texture for a custom model you're providing It's not a standard game file that requires a fix. To be honest I'd be against merging this, but I'll give you a chance @AxlRose-RX to explain to us the use case, and answer to @CosmosXIII question.

AxlRose-RX:
yes is an extra texture/3D model, with different UVs that basically no other mod has for this battle scene, so adding it won't affect others people at all. It will ONLY affect whoever uses my mod and I just didn't want to add my own lighting.toml for this then it cancelling other people's lighting.toml. Or if other people's lighting.toml was loaded, my version of Grasslands will get the shadows on the sky box because is not excluded

image

CosmosXIII:
I don't know any one else providing lighting.toml in their mods so I don't think this is a problem right now.
It may become a problem in the future but we may also have a better/different way to adjust lighting per texture (nothing definite but my idea is having a per-texture material config and we could include an option to disable shading).

AxlRose-RX:
that will be great or read the global + lighting.toml from all the mods that the players have enabled?

Lets say Gaia 7 has:
disable_lighting_textures = [ "STAGE02_T99", .....

Mine has only
disable_lighting_textures = [ "STAGE02_T06",

Final will include all from global + T99 and T06. a "-STAGE02_T80" minus sign could tell the game to remove one too, etc and let the mod order do its thing

julianxhokaxhiu:
Ok looks clear to me that this PR doesn't make sense to be merged as again it's about a custom model. We might consider enhancing the exclusion list logic and/or refine how the lighting makes use of those in the future. Consider adding your own `config.toml` on the mod for the time being :)

PR #583, opened 2023-07-14
Update crashdump.cpp

eve-atum:
Adjust crash message to suggest checking DEP settings as the first troubleshooting step


julianxhokaxhiu:
I'm sorry but this has nothing to do with FFNx. The crash can happen for multiple reasons and DEP could be just one of them.

Still, thanks :)

PR #584, opened 2023-07-21
FF8: Add more overridable textures for Worldmap + fixes

### LINKED ISSUE #565: [ FF8 DE NV ] Cannot enter worldmap
**Describe the bug**
The game hang when entering worldmap. Not logs, no crash. The wmx file opening seems the last known step.
This is happening only with DE nvidia patch.

**To Reproduce**
Enter the worldmap with DE nvidia patch.

**Additional notes**
I disabled vram and worldmap modules, and the bug is still there.

myst6re:
## Summary

- Adding ability to override textures from wmset section 39 (rails and roads), 40 and 42 (vehicles).
- Restore ability to mod "texl" textures, even with ff8_worldmap_internal_highres_textures option enabled. + bug fixes
- Add save_textures_legacy to save texture in legacy format instead of recommended one
- Fixes #565
- Texture reload hack mini optimization on worldmap

### ACKs

### DIFF
--- a/src/cfg.cpp
+++ b/src/cfg.cpp
@@ -52,6 +52,7 @@ long external_voice_music_fade_volume;
 bool enable_voice_auto_text;
 bool enable_auto_run;
 bool save_textures;
+bool save_textures_legacy;
 bool trace_all;
 bool trace_renderer;
 bool trace_movies;
@@ -197,6 +198,7 @@ void read_cfg()
 	external_widescreen_path = config["external_widescreen_path"].value_or("");
 	external_time_cycle_path = config["external_time_cycle_path"].value_or("");
 	save_textures = config["save_textures"].value_or(false);
+	save_textures_legacy = config["save_textures_legacy"].value_or(false);
 	trace_all = config["trace_all"].value_or(false);
 	trace_renderer = config["trace_renderer"].value_or(false);
 	trace_movies = config["trace_movies"].value_or(false);
@@ -342,7 +344,7 @@ void read_cfg()
 		hext_patching_path += "/de";
 		break;
 	case VERSION_FF8_12_DE_NV:
-		hext_patching_path += "/de";
+		hext_patching_path += "/de_nv";
 		break;
 	case VERSION_FF8_12_SP:
 		hext_patching_path += "/es";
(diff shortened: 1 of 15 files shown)


src/ff8/texture_packer.cpp
```diff
@@ -42,6 +42,8 @@ void TexturePacker::cleanVramTextureIds(const TextureInfos &texture)
 
 void TexturePacker::cleanTextures(ModdedTextureId previousTextureId, bool keepMods)
 {
+	ffnx_trace("%s: previousTextureId=%d keepMods=%d\n", __func__, previousTextureId, keepMods);
```
julianxhokaxhiu, line 45: Remember to put an `if` if it's intended to be kept.

myst6re: Yes, it is not intended to be kept

src/cfg.cpp
```diff
@@ -342,7 +344,7 @@ void read_cfg()
 		hext_patching_path += "/de";
 		break;
 	case VERSION_FF8_12_DE_NV:
-		hext_patching_path += "/de";
+		hext_patching_path += "/de_nv";
```
myst6re, line 347: #565

src/ff8/vram.cpp
```diff
@@ -455,7 +521,11 @@ uint32_t ff8_field_read_map_data(char *filename, uint8_t *map_data)
 	snprintf(tex_filename, MAX_PATH, "field/mapdata/%s/%s", get_current_field_name(), get_current_field_name());
 
 	if (save_textures) {
-		ff8_background_save_textures(tiles, mim_texture_buffer, tex_filename);
+		if (save_textures_legacy) {
+			ff8_background_save_textures_legacy(tiles, mim_texture_buffer, tex_filename);
+		} else {
+			ff8_background_save_textures(tiles, mim_texture_buffer, tex_filename);
+		}
```
julianxhokaxhiu, line 528: You can probably simplify this condition to something like:
```cpp
if (save_textures_legacy)
  ff8_background_save_textures_legacy(tiles, mim_texture_buffer, tex_filename);
else if (save_textures)
  ff8_background_save_textures(tiles, mim_texture_buffer, tex_filename);
```
This way if someone enables the legacy part only it will dump those, if it enables `save_textures` it will dump the FFNx ones. What do you think?

myst6re: Maybe it makes sense with more legacy texture names, okay fine I make the change

myst6re:
Ready!

julianxhokaxhiu (approved):

PR #587, opened 2023-08-30
FF7: Fix direct color

myst6re:
## Summary

Fix black color on field backgrounds

### Motivation

The PC rendering is different from the PSX rendering.

The PSX data clearly shows black pixels when the 15th bit is active in the 16-bit direct color (0x8000).
The PC version converted this black pixel to 0x0821 in data and at runtime it is converted again to 0x0403 (dark purple).

We also need to distinguish the black pixel to the transparent one using color key (0x0000).

#### PSX rendering
PSX rendering

#### FFNx rendering (and original PC)
FFNx rendering (and original PC)

#### Comparison of black

black-comparison

### ACKs

### DIFF
--- a/src/ff7_opengl.cpp
+++ b/src/ff7_opengl.cpp
@@ -291,6 +291,11 @@ void ff7_init_hooks(struct game_obj *_game_object)
 		patch_code_int(ff7_externals.field_init_viewport_values + 0x6E, 240);
 	}
 
+	// ########################
+	// field direct color black
+	// ########################
+	patch_code_short(uint32_t(ff7_externals.field_convert_type2_layers) + 0xE3, 0x8000);
+
 	// #####################
 	// worldmap footsteps
 	// #####################
--- a/Changelog.md
+++ b/Changelog.md
@@ -5,6 +5,7 @@
 ## FF7
 
 - Ambient: Allow ambient effects to playback in fields that use movies as background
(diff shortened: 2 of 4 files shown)


src/ff7_opengl.cpp
```diff
@@ -291,6 +291,11 @@ void ff7_init_hooks(struct game_obj *_game_object)
 		patch_code_int(ff7_externals.field_init_viewport_values + 0x6E, 240);
 	}
 
+	// ########################
+	// field direct color black
+	// ########################
+	patch_code_short(0x63F325 + 0xE3, 0x421);
```
myst6re, line 297: Currently my patch replace the faulty color by grey, this is not enough to get the original rendering

2023-08-30 23_49_35-Final Fantasy VII (Direct3D11) devel

I tried set this value to 0, but in this case the pixel is transparent, not black, so you will have artifacts between bg layers

myst6re: A darker blue works well for this field
`patch_code_short(0x63F325 + 0xE3, 0x1);`
2023-08-31 13_20_25-Final Fantasy VII (Direct3D11) devel

myst6re:
2023-09-02 16_17_48-Final Fantasy VII (Direct3D11) devel
Done! This is the fix on FFNx

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #593, opened 2023-09-08
60FPS: Fix char status menu rendering speed in FF7

tangtang95:
## Summary

In 60 FPS battles, currently the status menu skips the character status when a character has more than one status. 
This patch fixes it.

### Motivation

💯 

### ACKs


tangtang95:
Sorry for the mess with changing commits, title and description. I thought that refactoring the battle menu fix (using the 30fps approach would work), but unfortunately it cause another crash with ESUI (this crash is not easy to debug, so I don't want to go into this rabbit hole). So, I am back with the previous solution

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #594, opened 2023-09-08
FF8: Fix battle stage texture dump + option to use music files from remastered

myst6re:
## Summary

Fix battle stage texture dump with save_textures on.

option to use music files from remastered => if you extract the ogg files from remastered archives, you obtain files named like:
- 008s-antena.ogg
- 018s-julia.ogg
It is the same format as the original game, but the extension is replaced from sgt to ogg. Also this does not use xg files naming

### Motivation

I expected at first that Roses and Wine mod was using the same file name pattern than the original game.
That's not the case, we need to maintain a list of filename, that I don't want to.

But as a quickwin I can add the option for users who want to import oggs files from the remastered version to the 2013 version.

### ACKs


julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #592, opened 2023-09-07
30FPS: Fix crash conflict in FF7 with ESUI mod

tangtang95:
## Summary

Fix crash in battle mode when using 30fps mod and ESUI mod

### Motivation

Because yes!

### ACKs

- [ ] I have updated the [Changelog.md](https://github.com/julianxhokaxhiu/FFNx/blob/master/Changelog.md) file
- [x] I did test my code on FF7
- [ ] I did test my code on FF8

PR #597, opened 2023-09-09
FF7: Refactor 60FPS battle menu fix and fix ESUI compatibility

tangtang95:
## Summary

Related to https://github.com/julianxhokaxhiu/FFNx/pull/592 and https://github.com/julianxhokaxhiu/FFNx/pull/593. This refactoring of battle menu 60fps fix follows the approach used in 30fps. It requires less patches and it should give better results in compatibility and coherence. 

Also in the second commit I put a workaround for 30 fps + ESUI. I am not really a fan of it, but it does the job. 

### ACKs

### DIFF
--- a/src/ff7.h
+++ b/src/ff7.h
@@ -3236,25 +3236,19 @@ struct ff7_externals
 
 	// battle menu
 	uint32_t display_battle_menu_6D797C;
-	void (*display_cait_sith_slots_handler_6E2170)();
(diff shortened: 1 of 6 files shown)


julianxhokaxhiu (approved):

julianxhokaxhiu:
Thanks a lot for trying to find the best fix around this :) I agree the workaround is not the best but the condition around it is still FFNx related somehow so I'm fine with it. Well done!

PR #601, opened 2023-09-13
FF8: Increase max texture size to 16384 for external textures + Minor fixes

myst6re:
## Summary

- Non-us version does not use similar addresses in Worldmap.
- When an external texture contains an almost opaque alpha, like 254, the pixel is displayed black, force an opaque color to show the real color of the pixel instead
- texl textures (high res worldmap textures) can only be upscaled to 1024x1024, because source height is incorrect. But it should be incorrect as it replaces low res worldmap textures. So I decided to increase MAX_SCALE a lot. Now it is possible to upscale texl to 16384 (which is the max texture size on my machine).

### Motivation

Crash => no crash

### ACKs


julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #607, opened 2023-09-23
FF7: Implemented GPU-based original game lighting that better matches PSX version

CosmosXIII:
## Summary

I re-implemented the original lighting done by the game in the shaders instead of the original CPU-based implementation.
In doing so I found that the light directions used by the game were inverted compared to the PSX version, so the new modes fix that too. The new modes better match better the PSX version lighting but there are small differences in contrast. Part of this is due to FFNx normal auto-calculation so those may not fully match.

### Motivation

To make people happy. Also, if we end up doing smooth skinning, doing lighting on the GPU is necessary so this paves the path for that.

### ACKs


src/renderer.cpp
```diff
@@ -394,6 +411,9 @@ void Renderer::resetState()
     setOverallColorGamut(enable_ntscj_gamut_mode ? COLORGAMUT_NTSCJ : COLORGAMUT_SRGB);
     setGamutOverride();
     setGammaType();
+	setGameLightData();
```
julianxhokaxhiu, line 414: Can you please fix the indentation?

misc/FFNx.lighting.frag
```diff
@@ -104,9 +113,12 @@ uniform vec4 iblData;
 #define isPbrTextureLoaded FSTexFlags.y > 0.0
 #define isIblTextureLoaded FSTexFlags.z > 0.0
 
+#define gameLightingMode gameLightingFlags.x
+#define GAME_LIGHTING_PER_PIXEL 2
+
 void main()
 {
-	vec4 color = vec4(toLinear(v_color0.rgb), v_color0.a);
+	vec4 color = v_color0;
```
julianxhokaxhiu, line 121: Can you please fix the indentation here to match the lines below?

julianxhokaxhiu (changes requested):
Minor indentation issues, but the PR looks good to me. Thanks, looking forward to it :)

src/renderer.cpp
CosmosXIII: Fixed :)

misc/FFNx.lighting.frag
CosmosXIII: Fixed :)

julianxhokaxhiu:
Thank you :)

Issue #474, opened 2022-08-08
[ FF7 ] Collection of widescreen 16:9 bugs


CosmosXIII:
This is the collection of bugs related to the new widescreen 16:9 aspect ratio mode.

**Common**
- [x] 16:9 aspect ratio stretches to fill the screen when displayed on an ultrawide monitor

**Field mode**
- [ ]  Some fields show a back bar on the top/bottom of the screen
- [ ]  The chocobo black screen in the Chocobo Farm does not show
- [ ]  The wind overlay in Gaea's cliff does not scale to fill the screen
- [ ]  In fship_22, the vertical scrolling stops too early
- [ ] Aerith's Great Gospel Limit 4 shows the black background in the edges

**Battle mode**
- [ ] Compact Battle UI has black background on the bottom right and left
- [x] Limit break boxes extend beyond the intended window
- [ ] Pollensalta at Northern Cave - The full screen spell texture is not following true widescreen

**Worldmap mode**

**Other modes**
- [x] shop characters/text to appear cut off
- [ ]  Condor minigame is still 4:3

Before reporting new bugs/glitches, make sure you are running the latest canary version of FFNx. Also double check if the bug or glitch is not already in this list

KuraudoXS:
7th 2.6.14
FFNx 1.13.0.8

Other Mod - Compatibility issue
1) Compact Battle UI has black background right and left.

Fields
1) As soon as you step in the Highwind the animation background is not wide, resulting in black background right and left.

AnsgarTOdinson:
16:9 mode stretches to fill the screen when displayed on an ultrawide monitor.  Native 4:3 mode does not.  This applies to all modes.

tangtang95:
@KuraudoXS found this field map 'junbin21' having some problem with top camera range. Need to fix it in future config.toml

Sdub-12:
**PLEASE NOTE:** If you come here to report a crash while you were using 7th Heaven, please do on their relative support channels. NO SUPPORT will be provided, and your issue will be close immediately.

**Describe the bug**
16:9 Mode is causing shop characters/text to appear incorrectly/moved. The issue seems to be specifically to shop prices. Will attach screenshots. Using Canary build.

**To Reproduce**

1. Go to Gamedriver Settings
2. Enable 16:9 Widescreen
3. Begin Game
4. Enter Shop interface
5. See error

**Expected behavior**
I expected the prices to appear normally as is with 4:3

**Screenshots**

[...]

PR #613, opened 2023-10-01
FF7: Fix Widescreen glitches for Bahamut Zero and Pollensalta

tangtang95:
## Summary

Fix widescreen glitches (related to #474) for:
 - Bahamut Zero stars background positioning and size
 - Pollensalta cold breath attack: moving background and moving white dots 

Regarding Pollensalta fix for the moving background size: this only works when in 30/60fps mode because I used a function hooked only when in 60 fps mode. I don't think there is other way to do it, but I think the correct solution should be to fix the texture page which I suppose it is inside files with `.s` as extension (in this case, `special/hubuki/kemu.s`). So I left the solution for 30/60fps just as a temporary fix.

### Motivation

I don't know why either

### ACKs


julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you :)

Issue #622, opened 2023-11-06
[ FF7 ] Character name screen does not work correctly when using a gamepad in Steam Input mode


maciej-trebacz:
**Describe the bug**

**To Reproduce**

1. Go to Steam settings, in the Controller tab select "Enable Steam Input for [your gamepad type]"
2. Go to the Library screen, select Final Fantasy VII and click on the little controller icon on the right, in the same row where the PLAY button is
3. Configure Steam Input with your controller, for example go to community layouts and find "FINAL FANTASY VII DS4/5"
4. Launch FF7 and find a spot where a character is about to be named
5. Try using the controller to navigate the character naming screen, notice that pressing X or O on the controller enters letters like "c" or "v" instead of the actual highlighted letter

**Expected behavior**
The game normally expects you to use cursor keys to select characters and then confirm the selection with [OK], you can also use [CANCEL] to delete characters. and [START] to move the cursor to the section on the right to confirm your entry. On the original Steam driver this works both on keyboard, on a controller in normal mode and in the Steam Input mode. Direct character entry from the keyboard is **not** possible on this version.

**Video**
https://github.com/julianxhokaxhiu/FFNx/assets/1614514/88f07c62-b09d-40fd-a6df-ffb23146f7e0

On this video I press [X] and [O] on my controller and notice how it inputs "c" and "x" instead.

**Additional context**
Apparently direct keyboard input *was* a feature of the original 1998 PC release, but was removed in the Steam version (the original Steam driver patched out three functions responsible for it: 719ce4, 718710 and 719d03) possibly exactly because they were in conflict with the Steam Input, which emulates keyboard presses and so the game interpreted them as direct letter input instead of [OK] / [CANCEL]

PR #627, opened 2023-11-20
FF7: Fix character naming screen not working correctly when Steam Input is enabled

maciej-trebacz:
## Summary

Fixes issue #622 by patching out functions that enable direct keyboard input on character naming screens. This is what Aali driver originally did and this change apparently wasn't ported to FFNx. See linked issue for more details.

### ACKs

### DIFF
--- a/src/ff7_opengl.cpp
+++ b/src/ff7_opengl.cpp
@@ -77,12 +77,6 @@ void ff7_init_hooks(struct game_obj *_game_object)
 	// Load Models atoi function
 	replace_call_function(ff7_externals.field_load_models_atoi, ff7_field_load_models_atoi);
 
-  // Restore Steam release behavior on character name screen when using gamepads in Steam Input mode
-  // Aali driver used to patch out these three functions to fix this issue
-  replace_function(ff7_externals.set_default_input_settings_save, noop);
-  replace_function(ff7_externals.keyboard_name_input, noop);
-  replace_function(ff7_externals.restore_input_settings, noop);
-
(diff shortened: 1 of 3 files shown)


src/ff7_opengl.cpp
```diff
@@ -77,6 +77,12 @@ void ff7_init_hooks(struct game_obj *_game_object)
 	// Load Models atoi function
 	replace_call_function(ff7_externals.field_load_models_atoi, ff7_field_load_models_atoi);
 
+  // Restore Steam release behavior on character name screen when using gamepads in Steam Input mode
+  // Aali driver used to patch out these three functions to fix this issue
+  replace_function(ff7_externals.set_default_input_settings_save, noop);
+  replace_function(ff7_externals.keyboard_name_input, noop);
+  replace_function(ff7_externals.restore_input_settings, noop);
```
julianxhokaxhiu, line 84: Having those here will impact 1998 release as well. Better to move it where we do check if the game is the steam edition.

julianxhokaxhiu (changes requested):
Thanks for the PR. I'm not against it but the way it is currently shape may create an issue with the non-Steam editions of the game. Left a comment on the impacted block.

src/ff7_opengl.cpp
maciej-trebacz: @julianxhokaxhiu fixed.

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #632, opened 2024-01-03
FF8: Enable 360 analog input

myst6re:
## Summary

Send analog input to the core game. Per module, not globally, because analog already works well as a d-pad.

- Field module: send analog values for characters movements (it automatically overrides d-pad input sent by the driver)
- World module: send analog values for everything (left and right sticks), but override d-pad input only for characters movements

### Motivation

PlayStation version always supported analog input for character movements, not only 8 directions, but a range of 255*255 values.
The game also implements an auto walk/run feature depending of the strength of the analog input.
Finally in worldmap, vehicles can be driven using the right stick along with the left stick.

https://finalfantasy.fandom.com/wiki/Final_Fantasy_VIII_version_differences#Microsoft_Windows

### ACKs

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -39,6 +39,8 @@
 ## FF8
 
 - Hext: `fix_uv_coords` hext files are not needed anymore ( https://github.com/julianxhokaxhiu/FFNx/pull/619 )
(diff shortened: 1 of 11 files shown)


julianxhokaxhiu:
Thank you!

Issue #460, opened 2022-06-29
[ FF7 ] Battle camera control bugs


tangtang95:
This is a list of bugs I encountered with battle camera controls. Just for remembering, I list it here so we can check it out.

There are some occasion during the start of the battle, where the camera upon taking control (https://github.com/julianxhokaxhiu/FFNx/blob/master/src/ff7/camera.cpp#L265-L272) it snaps in one frame to another camera position. For example, this can be replicated against the thiefs enemies outside the church of Aerith on the road towards Aerith's house. It's not related to the `abs() < 1`, but maybe it's about this (https://github.com/julianxhokaxhiu/FFNx/blob/master/src/ff7/camera.cpp#L323-L338). What happens if the `battle id` is the same as the previous battle? 

There is also the bug of camera control during the camera end script movement (during fanfare) in the battle against the icicle enemy. 

Also, why https://github.com/julianxhokaxhiu/FFNx/blob/master/src/ff7/camera.cpp#L317 the parameter `cameraFocusPosition` is not used or overwritten?

@CosmosXIII

julianxhokaxhiu:
@CosmosXIII is this bug still valid?

CosmosXIII:
@julianxhokaxhiu Sorry for the very late reply but yes this bug is still valid. I'll do some testing and see if I can fix it.

PR #637, opened 2024-01-13
FF7: New camera control in field mode using the right stick

CosmosXIII:
## Summary

This PR adds the functionality to scroll the camera with the right analog stick. The camera goes back smoothly to it's usual position once the analog stick is released.

This PR also has some other minor fixes.
- Fixes the problems raised in #460 -> was not able to reproduce the bugs but this should be safer than before
- Reduced lighting specular because it was too washed-out in some places

### Motivation

Because we need more rainforest, not less, I'm serious.

### ACKs


julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #641, opened 2024-01-17
FF8: Add option to display playstation icons

myst6re:
## Summary

2024-01-18 00_25_46-Final Fantasy VIII

### Motivation

2024-01-18 00_29_25-Final Fantasy VIII

### ACKs

### DIFF
--- a/misc/FFNx.toml
+++ b/misc/FFNx.toml
@@ -705,6 +705,11 @@ ff8_fix_uv_coords_precision = true
 #~~~~~~~~~~~~~~~~~~~~~~~~~~~
 ff8_external_music_force_original_filenames = false
 
+# Use icon textures for gamepad buttons instead of "B1", "B2", etc...
+# By default the PS1 icons are used from the game data.
+#~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ff8_use_gamepad_icons = false
+
 ## GAME INSTALLATION OPTIONS
 
 #[APP PATH]
--- a/src/ff8_opengl.cpp
+++ b/src/ff8_opengl.cpp
@@ -505,6 +505,105 @@ LPDIJOYSTATE2 ff8_update_gamepad_status()
 	return ff8_externals.dinput_gamepad_state;
 }
 
+int ff8_get_input_device_capabilities_number_of_buttons(int a1)
+{
+	return xinput_connected ? 10 : std::min<DWORD>(joystick.GetCaps()->dwButtons, 10);
+}
+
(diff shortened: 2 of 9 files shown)


myst6re:
The PR is not ready because I need to rewrite the draw_icon function from the game to keep the keyboard display
2024-01-18 00_28_39-Final Fantasy VIII

misc/FFNx.toml
```diff
@@ -705,6 +705,10 @@ ff8_fix_uv_coords_precision = true
 #~~~~~~~~~~~~~~~~~~~~~~~~~~~
 ff8_external_music_force_original_filenames = false
 
+# Use PlayStation icons for gamepad buttons
+#~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ff8_use_playstation_icons = false
```
myst6re, line 710: @julianxhokaxhiu are you ok with the name?

julianxhokaxhiu: Can the playstation icons be replaced using the external texture approach?

myst6re: Yes, if the option ff8_use_playstation_icons is enabled, it will use the icons.tex texture, which is overridable via external texture

julianxhokaxhiu: Ah nice. So would it make use to name it something like `ff8_use_gamepad_icons`? Thinking about the possibility to override the tex file with an Xbox layout for example via a custom mod. Although by default will use the PSX icons, keeping the name generic might also benefit the mod guide for eg. What do you think?

I know it's a minor thing but I'm open for suggestions in case. Thanks!

src/ff8_opengl.cpp
```diff
@@ -505,6 +505,105 @@ LPDIJOYSTATE2 ff8_update_gamepad_status()
 	return ff8_externals.dinput_gamepad_state;
 }
 
+int ff8_get_input_device_capabilities_number_of_buttons(int a1)
+{
+	return xinput_connected ? 16 : joystick.GetCaps()->dwButtons;
```
julianxhokaxhiu, line 510: Since the memory allocation is only for 16 buttons, wouldn't it make sense to cap to 16 this capability anyway? Or at least check if `dwButtons < 16 ? dwButtons : 16` or something like this.

myst6re: The reality is worse: the game would never save more than 10 buttons in its mapping anyway

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #657, opened 2024-02-18
FF7: Fixed menu not working when using the buggy

CosmosXIII:
## Summary

Fixed menu not working when using the buggy and external worldmap. For some reason the world_is_player_model_bitmask function does not work so instead I made it check the model id values manually.

### Motivation

To return the original behavior where the player can use the menu while using the buggy.

### ACKs

### DIFF
--- a/src/ff7/world/player.cpp
+++ b/src/ff7/world/player.cpp
@@ -408,7 +408,7 @@ namespace ff7::world {
             {
                 if(player_model_id == HIGHWIND)
                     ff7_externals.world_run_special_opcode_7640BC(6);
-                else if(ff7_externals.world_is_player_model_bitmask(CLOUD | TIFA | CID | BUGGY) && ff7_externals.world_get_player_walkmap_type() != 14)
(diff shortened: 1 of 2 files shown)


src/ff7/world/player.cpp
```diff
@@ -408,7 +408,9 @@ namespace ff7::world {
             {
                 if(player_model_id == HIGHWIND)
                     ff7_externals.world_run_special_opcode_7640BC(6);
-                else if(ff7_externals.world_is_player_model_bitmask(CLOUD | TIFA | CID | BUGGY) && ff7_externals.world_get_player_walkmap_type() != 14)
+                // For some reason function world_is_player_model_bitmask does not work so instead check the model id values manually
+                // else if(ff7_externals.world_is_player_model_bitmask(CLOUD | TIFA | CID | BUGGY) && ff7_externals.world_get_player_walkmap_type() != 14)
```
julianxhokaxhiu, line 412: Do we need the comment section above? I'd prefer having the fix directly with the reason explained in the PR if possible. Thanks!

CosmosXIII: done

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you :)

PR #670, opened 2024-02-29
FF8: Fix texture animation only partially animated

myst6re:
## Summary

- When the external texture is upscaled (which is the common case), texture copy is not scaled up to the full size of the copy
- Adding some rules in the .reg file to allow alt+tab and windows key press (FF8 2000 only). It will ask Windows to disable all compatibility patches, which are not needed anyway with FFNx

### ACKs


julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #672, opened 2024-03-03
Music: Add new relative_speed option in the config layer

myst6re:
## Summary

External Music: ability to set music speed via config

### Motivation

Kurodo asked me if it was possible to sync "tb" music track (main theme alternative / Cloud's mind) with the scene, like in PS1
https://www.youtube.com/watch?v=ELt0-heDj0I&t=20020s

The PC version is faster than the PSX version, one solution would be speed up the music. That's increase the pitch a little, I'm not a big fan of this solution. But everyone seems fine with it, for example the modder Postscriptthree released a fix that just does just that, DLPB also did this in the past.

Since this patch is really light, and can be used by modders for whetever creative reasons (I enjoy listening to the slow version of costa del sol, for example), so let's do it!

### ACKs

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,10 @@
 
 - Full commit list since last stable release: https://github.com/julianxhokaxhiu/FFNx/compare/1.18.1...master
 
+## Common
+
+- Music: Add new [relative_speed](https://github.com/julianxhokaxhiu/FFNx/blob/master/misc/FFNx.music.toml#L29-L31) option in the config layer ( https://github.com/julianxhokaxhiu/FFNx/pull/672 )
+
 ## FF8
 
 - Graphics: Fix texture animations by copy only partially animated ( https://github.com/julianxhokaxhiu/FFNx/pull/670 )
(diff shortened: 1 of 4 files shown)


julianxhokaxhiu (approved):

Changelog.md
```diff
@@ -2,6 +2,10 @@
 
 - Full commit list since last stable release: https://github.com/julianxhokaxhiu/FFNx/compare/1.18.1...master
 
+## Common
+
+- External music: ability to set music speed via config ( https://github.com/julianxhokaxhiu/FFNx/pull/672 )
```
julianxhokaxhiu, line 7:
```suggestion
- Music: Add new [relative_speed](https://github.com/julianxhokaxhiu/FFNx/blob/master/misc/FFNx.music.toml#L29-L31) option in the config layer ( https://github.com/julianxhokaxhiu/FFNx/pull/672 )
```

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thanks!

PR #675, opened 2024-03-10
FF8: Do not use hashes to name vibration config files

myst6re:
## Summary

Use the current magic id to identify battle vibrate sequences

### Breaking change

Vibration config file are not named with hashes anymore, but with the magic id like this:

```
battle_effect_{magicID}.toml
```

### Motivation

There are several pools of vibration sequences:
- the common one (for the pause menu)
- the field one (for setvibrate opcode)
- the world one (for vehicles)
- and a lot for battle effects

Before this PR I wasn't able to identify which battle effect is linked to which vibrate pool.
So I used hashes over the data to identify most of them.

### ACKs


julianxhokaxhiu (approved):

julianxhokaxhiu:
Lovely discovery :) Thank you!

Issue #677, opened 2024-03-22
[FF7 Steam] Corrupted ff7input.cfg


petfriendamy:
Sometimes, with FFNx installed, the ff7input.cfg file will get corrupted, overwriting all of the user's keybinds and replacing them with what seems to be some sort of default controls on the numpad. The speedrunning community has found a consistent way to reproduce this behavior:

1) Start a new game
2) Enter the config menu (it does not seem to matter if you change any settings)
3) Exit the config menu
4) Doing this will create the corrupted ff7input.cfg file

Multiple users have reported this behavior.
[ff7input (corrupt).cfg.zip](https://github.com/julianxhokaxhiu/FFNx/files/14725195/ff7input.corrupt.cfg.zip)

julianxhokaxhiu:
Can you please confirm if you're able to repro the issue since we merged #683 ? Thanks!

petfriendamy:
Looks like it's been fixed, thanks!

PR #683, opened 2024-03-31
FF7: Prevent input configuration corruption on Steam edition

maciej-trebacz:
## Summary

Replaces the default config bitmask value from 0x41 to 0x45, which sets the default Controller type to "Customize" instead of "Normal" in FF7 Steam.

### Motivation

FF7 Steam locks the Controller setting to "Customize" because the "Normal" control scheme is broken (it uses numpad controls). However there is [a bug](https://github.com/julianxhokaxhiu/FFNx/issues/677) that can cause your config file to get corrupted and try to load the "Normal" control scheme anyways. This is because the config bitmask value that is loaded on New Game is initialized to `0x41` with the third bit responsible for switching from "Normal" to "Customize" not set.

This change patches this bitmask to set that bit on so the game always uses "Customize" control scheme.

### ACKs


maciej-trebacz:
Since I obviously don't know how to Git properly please squash this instead of merging.

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thanks a lot :)

PR #685, opened 2024-04-14
FF8: Magic textures wrongly dumped when save_textures option is disabled

myst6re:
## Summary

Magic textures wrongly dumped when save_textures option is disabled.

**Important:** If you have installed [FFNx-v1.19.0.0](https://github.com/julianxhokaxhiu/FFNx/releases/tag/1.19.0), please remove the directories `mods/Textures/battle` and `mods/Textures/magic` and reinstall your battle mods if you had any.

### ACKs

### DIFF
--- a/src/saveload.cpp
+++ b/src/saveload.cpp
@@ -76,6 +76,12 @@ void save_texture(const void *data, uint32_t dataSize, uint32_t width, uint32_t 
 	char filename[sizeof(basedir) + 1024];
 	uint64_t hash;
 
+	if (!save_textures) {
+		ffnx_warning("Save texture skipped because the option \"save_textures\" is disabled (name=%s).\n", name);
+
+		return;
+	}
+
 	if (is_animated)
 	{
 		char xxhash_filename[sizeof(basedir) + 1024];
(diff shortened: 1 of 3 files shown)


src/saveload.cpp
```diff
@@ -76,6 +76,12 @@ void save_texture(const void *data, uint32_t dataSize, uint32_t width, uint32_t
 	char filename[sizeof(basedir) + 1024];
 	uint64_t hash;
 
+	if (!save_textures) {
+		ffnx_warning("Save texture skipped because the option \"save_textures\" is disabled (name=%s).\n", name);
```
julianxhokaxhiu, line 80: This will create another massive log statement, so I'd suggest prefixing it with another if:
```suggestion
if (trace_all) ffnx_warning("Save texture skipped because the option \"save_textures\" is disabled (name=%s).\n", name);
```
Fine to add other `trace_` vars if needed for this case.

myst6re: At first no, this log will not appear as everytime we use save_texture, we check if save_textures is enabled before. In the future maybe, if we forgot to check before

julianxhokaxhiu: Ah ok I see now the angle of this. I thought you wanted to short circuit it in the function itself, but if it's supposed to be just a warning while we develop makes sense :) Thanks!

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thanks!

PR #686, opened 2024-04-17
FF8: Fix game over crash with external textures

myst6re:
## Summary

Ignore PNGs without alpha (for now)

### Motivation

Converting PNGs adds a little overhead, it is not big, but we don't want to promote PNGs in mods

### ACKs

### DIFF
--- a/src/image/image.cpp
+++ b/src/image/image.cpp
@@ -118,10 +118,18 @@ bool loadPng(const char *filename, bimg::ImageMip &mip, bimg::TextureFormat::Enu
     _width = png_get_image_width(png_ptr, info_ptr);
     _height = png_get_image_height(png_ptr, info_ptr);
 
+    if (color_type == PNG_COLOR_TYPE_RGB && (targetFormat == bimg::TextureFormat::BGRA8 || targetFormat == bimg::TextureFormat::RGBA8)) {
+        ffnx_warning("%s: PNG files without alpha is not supported, please convert it to RGBA for improved performance\n", __func__);
+
+        return false;
+    }
+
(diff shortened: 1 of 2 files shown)


src/image/image.cpp
```diff
@@ -134,6 +140,8 @@ bool loadPng(const char *filename, bimg::ImageMip &mip, bimg::TextureFormat::Enu
 
     data = (uint8_t*)driver_calloc(datasize, sizeof(uint8_t));
 
+    ffnx_info("%s: data_size=%d width=%d height=%d bit_depth=%d color_type=%X\n", __func__, datasize, _width, _height, bit_depth, color_type);
```
julianxhokaxhiu: Do we want to log this every time or only when `trace_loaders` is enabled?

myst6re:
Ready

src/image/image.cpp
myst6re: We do not want this log at all I guess, do you want it?

julianxhokaxhiu: Not sure, if it helps debugging why not. Does it help you in case for future crashes to be reported?

myst6re: Okay then, I re-add it

julianxhokaxhiu: Maybe consider marking it as `ffnx_trace` and put it with the usual `trace_all || trace_loaders` if. That would be more than enough

myst6re: Okay sorry I marked resolved before reading your message, it should be good now

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #687, opened 2024-04-21
FF8: Allow mod textures per VRAM page

myst6re:
## Summary

Allow mod textures per VRAM page. This is like how [Tonberry](https://forums.qhimm.com/index.php?topic=15945.0) works, but no hashes needed.

File name format: the vram page id is added and if there are several textures in the vram page, most of them are put in the filename `{texture_names}_{vram page id}_{palette x}_{palette y}`
Example: `A0STG101.X-0_16_0_255`

### Motivation

Even if VRAM pages are fragmented and can contains several unrelated textures, they could be useful for modders.

- They are GPU accelerated, meaning that DDS files are uncompressed on the GPU without blocking the CPU
- Some part of some big textures could be reduced in size, especialy in battle where some VRAM pages are animated, and other not
- A lot of existing mods could be ported easily to FFNx

### Remaining Issues

- [X] Palette index can change for the same texture in FF8. Override texture by palette index is possible but can lead to glitches in battle or worldmap => fixed by using palette coordinates in vram instead
- [ ] File names could colide if several textures were found, because I choose to truncate the filename if it is too big
- [ ] I only use texture file names to compose the vram page file name, so if textures are in several different directories, I use the first directory name as a directory for the vram page file
- [ ] Field background is not stable, prefer use the tonberry compatibility layer

### ACKs

### DIFF
--- a/src/ff8/vram.cpp
+++ b/src/ff8/vram.cpp
@@ -47,6 +47,7 @@ int next_psxvram_x = -1;
 int next_psxvram_y = -1;
 int next_psxvram_pal_x = -1;
 int next_psxvram_pal_y = -1;
+texture_page *current_tex_page = nullptr;
 int next_texl_id = 0;
 int next_texture_count = -1;
 int next_do_not_clear_old_texture = false;
@@ -196,11 +197,13 @@ int read_vram_to_buffer_parent_call1(struc_50 *psxvram, texture_page *tex_page, 
 
 	next_psxvram_x = (x >> (2 - bpp)) + ((rel_pos & 0xF) << 6);
 	next_psxvram_y = y + (((rel_pos >> 4) & 1) << 8);
+	current_tex_page = tex_page;
 
 	int ret = ff8_externals.sub_464F70(psxvram, tex_page, x, y, w, h, bpp, rel_pos, a9, target);
 
 	next_psxvram_x = -1;
 	next_psxvram_y = -1;
+	current_tex_page = nullptr;
 
 	return ret;
 }
@@ -227,11 +230,13 @@ int read_vram_to_buffer_parent_call2(texture_page *tex_page, int rel_pos, int a3
 
 	next_psxvram_x = (tex_page->x >> (2 - tex_page->color_key)) + ((rel_pos & 0xF) << 6);
 	next_psxvram_y = tex_page->y + (((rel_pos >> 4) & 1) << 8);
+	current_tex_page = tex_page;
 
 	int ret = ((int(*)(texture_page *, int, int))ff8_externals.sub_4653B0)(tex_page, rel_pos, a3);
 
 	next_psxvram_x = -1;
 	next_psxvram_y = -1;
+	current_tex_page = nullptr;
 
 	return ret;
 }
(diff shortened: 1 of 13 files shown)


src/ff8/vram.cpp
```diff
@@ -1547,4 +1652,7 @@ void vram_init()
 	replace_call(ff8_externals.battle_enter + 0x35, engine_set_init_time);
 	// Clear texture_packer on every module exits
 	replace_call(ff8_externals.psxvram_texture_pages_free + 0x5A, clean_psxvram_pages);
+	// Free pc_name in tex_header
+	replace_call(0x464680 + 0x98, free_tex_header);
```
myst6re, line 1656: TODO

julianxhokaxhiu:
Wow, this looks like the holy grail of all possible solutions. Thanks myst, looking forward!

src/ff8/vram.cpp
```diff
@@ -291,36 +301,141 @@ int read_vram_to_buffer_with_palette2_parent_call(int a1, uint16_t rel_pos, int1
...
+	if (trace_all || trace_vram) ffnx_trace("%s: vram_pos=(%d, %d) target=0x%X target_w_in_bytes=%d w=%d h=%d bpp=%d\n", __func__, next_psxvram_x, next_psxvram_y, int(target), target_w_in_bytes, w, h, bpp);
 
 	if (next_psxvram_x == -1)
 	{
 		ffnx_warning("%s: cannot detect VRAM position\n", __func__);
 	}
 	else
 	{
-		texturePacker.registerTiledTex(target, next_psxvram_x, next_psxvram_y, target_w, h, Tim::Bpp(bpp));
+		const TexturePacker::TiledTex &tiledTex = texturePacker.registerTiledTex(target, next_psxvram_x, next_psxvram_y, w, h, Tim::Bpp(bpp));
```
myst6re, line 406: Use "w" instead "target_w" to fix a find texture problem

myst6re:
I still need to test some more, but I think it is ready for review @julianxhokaxhiu

julianxhokaxhiu:
Thanks @myst6re for the amazing work you're driving in this PR. I did look into it and so far looks great, didn't spot anything critical so LGTM for now :) Looking forward!

myst6re:
Tests finished

myst6re:
Still known, I don't think I will fix the second and third, the only solution I see is to make a static list that matches a certain pattern. For example detect the name world/dat/wmset/section38/texture20_texture16_texture0_texture1_texture2_texture17_texture18_28.png (real example) and rename it water.png or whatever. I would like to tend to names that are officially used in the remaster, or in demaster mods if the remaster does not support those textures.

julianxhokaxhiu:
I see, makes sense. Allright, guess it's an accepted issue, but that doesn't require an immediate fix, right?

myst6re:
> I see, makes sense. Allright, guess it's an accepted issue, but that doesn't require an immediate fix, right?

Not immediate no. If, in the future, we change names of our files, that could bother modders, but we also may fallback to the old names.

src/ff8/vram.cpp
myst6re: Nope I left this comment for you, so you see one change hidden in the diff of another change

julianxhokaxhiu: Ah all right, usually Github resolves this automatically when it detects a change but it didn't for this one :) Thanks

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

src/ff8/vram.cpp
myst6re: Not if I make the comment after the change :)

PR #692, opened 2024-05-07
Common: Bilinear filtering + external texture optimizations

myst6re:
## Summary

- External textures: Reuse already loaded textures on fallback to palette 0
- External textures (FF8): Optimize field external textures support (Tonberry Mods layer)
- Rendering: Add bilinear filtering option `enable_bilinear`

#### Reuse already loaded textures on fallback to palette 0

Sometimes for the same texture, there are many palettes. In the graphic driver, we do not upload palettes anymore, and external textures often does not need one texture per palette, but are optimized to one texture for all palettes.
In this case, it is interesting for performance that this "one texture" is opened from disk and uploaded to gpu only once. And if the game try to upload the same texture, but with a different palette, it could detects this case and just reuse the already uploaded texture.

#### Add bilinear filtering option `enable_bilinear`

If was asked by few users. For now I just added the option with a filtering enabled only for 3D models, like we already do in ff7 for external texture.

### ACKs

### DIFF
--- a/misc/FFNx.toml
+++ b/misc/FFNx.toml
@@ -85,6 +85,11 @@ enable_antialiasing = 0
 #~~~~~~~~~~~~~~~~~~~~~~~~~~~
 enable_anisotropic = true
 
+#[BILINEAR]
+# Enable bilinear filtering on 3D textures. For FF7, textures in menu are also filtered.
+#~~~~~~~~~~~~~~~~~~~~~~~~~~~
+enable_bilinear = false
+
 #[LIGHTING]
 # Enable advanced lighting mode with real-time shadows.
 # NOTICE: Parameters such as light direction and color can be edited on the lighting debug window in the FFNx DevTools.
--- a/src/ff8_opengl.cpp
+++ b/src/ff8_opengl.cpp
@@ -1077,11 +1077,6 @@ void ff8_init_hooks(struct game_obj *_game_object)
 		patch_code_byte(ff8_externals.worldmap_alter_uv_sub_553B40 + 0x234, 0);
 	}
 
-	// Fix blue color in battle with fire spells
-	patch_code_byte(ff8_externals.read_vram_palette_sub_467370 + 0x6F, 0);
-	// Fix half alpha for palettes
-	patch_code_byte(ff8_externals.read_vram_palette_sub_467370 + 0x64, 0xFF);
-
 	// #####################
 	// battle toggle
 	// #####################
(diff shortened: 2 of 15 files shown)


src/gl/special_case.cpp
```diff
@@ -55,7 +55,7 @@ uint32_t gl_special_case(uint32_t primitivetype, uint32_t vertextype, struct nve
 	if(current_state.texture_set && VREF(texture_set, ogl.gl_set->force_filter) && VREF(texture_set, ogl.external)) current_state.texture_filter = true;
 
 	// Texture filtering mostly does not work well in FF8
-	if(ff8) current_state.texture_filter = false;
+	if(ff8) current_state.texture_filter = enable_bilinear && vertextype != TLVERTEX && current_state.texture_set;
```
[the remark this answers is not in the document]
myst6re: Last time I tried FF7 crashed, I can check again!

src/ff8_opengl.cpp
```diff
@@ -1077,11 +1077,6 @@ void ff8_init_hooks(struct game_obj *_game_object)
 		patch_code_byte(ff8_externals.worldmap_alter_uv_sub_553B40 + 0x234, 0);
 	}
 
-	// Fix blue color in battle with fire spells
-	patch_code_byte(ff8_externals.read_vram_palette_sub_467370 + 0x6F, 0);
```
myst6re, line 1081: Note: write_palette sub is now completely rewritten in vram.cpp, no need to keep those patches

myst6re:
Ready!

misc/FFNx.toml
```diff
@@ -85,6 +85,11 @@ enable_antialiasing = 0
 #~~~~~~~~~~~~~~~~~~~~~~~~~~~
 enable_anisotropic = true
 
+#[BILINEAR]
+# Enable bilinear filtering. For FF7, this parameter does not change anything
```
julianxhokaxhiu, line 89: > For FF7, this parameter does not change anything

Is this still the case?

myst6re: oops, no, it is not

julianxhokaxhiu:
If you can also please update the description of the PR and short the title I'd appreciate :) So it fits while squashing. Thank you!

myst6re:
> If you can also please update the description of the PR and short the title I'd appreciate :) So it fits while squashing. Thank you!

Done

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #694, opened 2024-05-17
Audio: Ensure Voice handle exists before stopping it

myst6re:
## Summary

Fix sudden stop of musics, for example in Fisherman Horizon Concert, when the dialog starts with Rinoa next to the stage.

### ACKs


julianxhokaxhiu:
Interesting patch, how fixing the voice layer fixes also the music part?

myst6re:
I think doing stopMusic on an invalid handler could stop group handlers, those kind of handler I use only in FH concert to synchronize tracks together

myst6re:
Also group handle is considered as invalid by soloud, and cannot be used to stop a music, so maybe internally soloud does not check the handle when you stop the music and find the voice of the group handle instead of doing nothing

julianxhokaxhiu (approved):

julianxhokaxhiu:
Makes sense. Thanks a lot :)

PR #696, opened 2024-05-20
FF8: Add animation support for per VRAM page texture replacement

myst6re:
## Summary

- Support rect copy animation and force palette id animation
- Optimize texture reload hack
- Fix conflict between texture caches when modding battle and worldmap (which leads to bad colors in non-replaced textures)

### Motivation

Why are you doing this? What use cases does it support? What is the expected outcome?

### ACKs

### DIFF
--- a/src/ff8/vram.cpp
+++ b/src/ff8/vram.cpp
@@ -339,7 +339,7 @@ void set_tex_name(const TexturePacker::TiledTex &tiledTex, ff8_tex_header *tex_h
 
 	std::string filename;
 	int file_count = 0;
-	int startVramId = 0;
+	int start_vram_id = 0;
 
 	for (const TexturePacker::IdentifiedTexture &tex: textures)
 	{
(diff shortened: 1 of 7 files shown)


myst6re:
#701

src/ff8/vram.cpp
```diff
@@ -935,6 +944,18 @@ void ff8_wm_chara_one_upload_palette_2(int16_t *pos_and_size, uint8_t *texture_b
 	next_pal_data = nullptr;
 }
 
+void ff8_wm_update_fence_animation()
+{
+	uint8_t current_disk = uint8_t((*ff8_externals.savemap)[51]);
```
julianxhokaxhiu, line 949: Really minor but wouldn't `ff8_currentdisk` work in this case? it's a global var already available. Or are we looking for something else here?

myst6re: I don't know where you could use this variable, but currently it is set only for the Steam version

julianxhokaxhiu: Right, fair enough we don't have any other native way to detect it on both versions. Ok we may replace this in the future, not a blocker :) Thanks

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #703, opened 2024-06-02
FF7: fix 60FPS battle text when text speed is set to max SLOW

tangtang95:
## Summary

When battle text speed is set to max SLOW (far right), battle text popup (e.g. messages after doing a steal) will behave like it is fast due to an integer overflow when casting int to char. 

This has been fixed by fixing the data type which should have been unsigned char and just for safety, I have added also a minimum to avoid even an unsigned char overflow.

### Motivation

Better experience in 60 FPS

### ACKs


julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #707, opened 2024-06-10
FF8: Fix wrong palette for external textures on battle and wm

myst6re:
## Summary

Two fixes:

- When battle is partially modded, non modded textures could use the wrong palette. This happens because the new cache conflict between the two ways of modding
- mkpath does not work anymore on ff8, because the wrong sub is called to uniformize paths

### ACKs


julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #714, opened 2024-08-09
FF8: Ability to use app_path config for the Steam version

myst6re:
## Summary

- [FF8 2013 Steam version] Force data paths to use app_path as the base path, this will allow to customize it via the FFNx.toml config file
- Remove .gitkeep files from deployment archives
- Fix music not starting on Worldmap with the original MIDIs and ff8_fps_limiter config != 0

### ACKs


src/ff8_opengl.cpp
```diff
@@ -905,21 +905,10 @@ int ff8_limit_fps()
...
-		break;
-	case MODE_GAMEOVER:
-		// Gameover screen has nothing to limit
-		qpc_get_time(&last_gametime);
-		return 0;
-	}
+	// For cross music play (vanilla music only)
+	qpc_get_time(&gametime);
+	double *timer_volume_change_related_dbl_1A788B8 = (double *)0x1A788B8;
+	*timer_volume_change_related_dbl_1A788B8 = (1000.0 / game_object->countspersecond) * qpc_diff_time(&gametime, &last_gametime, nullptr);
```
myst6re, line 911: TODO

src/common.cpp
```diff
@@ -2703,7 +2703,7 @@ uint32_t get_version()
 
 void get_data_lang_path(PCHAR buffer)
 {
-	strcpy(buffer, basedir);
+	strcpy(buffer, ff8 ? ff8_externals.app_path : basedir);
```
julianxhokaxhiu, line 2706: Where is `app_path` configured out of curiosity? And how does it different from `basedir` in this case?

myst6re: app_path is configured via the register key "AppPath". So in the Steam version, it comes from "dotemuRegQueryValueExA" override, which is by default `GetCurrentDirectory() + \`.
Then from this value, the game build paths, like {app_path}\data\field.fs for example.

FFNx does have an option to override app_path, after the RegQueryValueExA, but today it only works for the 2000-version, because we don't always use app_path where we should for game data.

(FF7 have something similar, but relative to "data" dir, but I don't want to mess with 7th Heaven)

`basedir` is `GetCurrentDirectoryA()`, which is similar to the default from the Steam version, but this is not overridable. We use this value for mods, config or data used by FFNx, not for internal game data (I think).

myst6re: With the 2000 version, I'm able to install FFNx in `C:/Users/myst6re/Documents/FF8`, put the FF8.exe and the bink dll, and run the game installed to `C:/Program Files/Square Soft, Inc/FINAL FANTASY VIII` without moving all the data, simply by setting app_path to `C:/Users/myst6re/Documents/FF8`

It has a huge advantage: I don't have to mess with admin rights, and I can configure that faster than what for example Satsuki as to do in its pack for ff7

julianxhokaxhiu (approved):

PR #717, opened 2024-08-18
Gardening: Refactor includes, header guards and related things

tangtang95:
## Summary

Huge cleanup of the codebase:
- Refactor includes by using the finest granularity possible (e.g. if vector is necessary include directly <vector> without using an include that transitively included vector). @julianxhokaxhiu Do we have any guidelines to the order of includes? Sometimes if I change the order, it does fix some redefinition error (that is probably ignored by some cflags or linker flags)
- Add pragma once header guard where missing
- Add inline to function in header files where necessary
- Trimming whitespace at the end

### Motivation

Overall, this sped up the build time from 2 minutes to 1 minutes and 36 seconds on my machine (excluding vcpkg install step, basically I timed only the cmake --build). But the main motivation was to understand all deps in order to try something new with the build system.

### ACKs

### DIFF
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -21,6 +21,10 @@
 //    GNU General Public License for more details.                          //
 /****************************************************************************/
 
+#define _WIN32_WINNT 0x0A00
+
+#include "renderer.h"
+
 #include <windows.h>
 #include <windowsx.h>
 #include <dwmapi.h>
@@ -30,7 +34,6 @@
 #include <hwinfo/hwinfo.h>
 #include <regex>
 
-#include "renderer.h"
 #include "hext.h"
 #include "ff8_data.h"
 
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -21,10 +21,6 @@
 //    GNU General Public License for more details.                          //
 /****************************************************************************/
 
-#define _WIN32_WINNT 0x0600
-
-#include "renderer.h"
-
 #include <windows.h>
 #include <windowsx.h>
 #include <dwmapi.h>
@@ -33,7 +29,14 @@
 #include <steamworkssdk/steam_api.h>
 #include <hwinfo/hwinfo.h>
 #include <regex>
+#include <shlwapi.h>
+#include <shlobj.h>
+#include <psapi.h>
+#include <mmsystem.h>
+#include <malloc.h>
+#include <ddraw.h>
 
+#include "renderer.h"
 #include "hext.h"
 #include "ff8_data.h"
 
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -28,6 +28,7 @@
 #include <sys/timeb.h>
 #include <steamworkssdk/steam_api.h>
 #include <hwinfo/hwinfo.h>
+#include <regex>
 
 #include "renderer.h"
 #include "hext.h"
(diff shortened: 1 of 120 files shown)


tangtang95:
@julianxhokaxhiu 

This is quite a huge change. Feel free to take your time in the review. Also, test it on FF8, because I cannot. Or maybe if you don't really feel like to change all these includes, it's not a problem if this is dropped. In the end, it's just for me to try some stuff with another build system

src/common.cpp
```diff
@@ -21,15 +21,19 @@
 //    GNU General Public License for more details.                          //
 /****************************************************************************/
 
+#define _WIN32_WINNT 0x0A00
```
julianxhokaxhiu, line 24: Do we need to define it? If yes, why?

julianxhokaxhiu:
Wow, thanks a lot for this PR. I honestly really like it, as it's the kind of gardening this codebase deserves and I'm happy it comes from someone who knows it really well. I have only one question on top of all of the changes for now, but I'll also test it later if it builds and runs fine on both FF7 and FF8. Appreciated!

src/common.cpp
myst6re: And 0x0A00 seems pretty high, it means we support only Windows 10+

tangtang95: I added it for this macro SM_CXPADDEDBORDER (https://github.com/julianxhokaxhiu/FFNx/blob/cf2a40f75882e5a4af926f891ce8e0a3ad0c15ee/src/common.cpp#L479). If you go into winuser.h you will find that it is behind and `ifndef` (https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getsystemmetrics). I did put a quite high value, but that is taken from renderer.cpp (https://github.com/julianxhokaxhiu/FFNx/blob/cf2a40f75882e5a4af926f891ce8e0a3ad0c15ee/src/renderer.cpp#L29). But maybe the one in renderer.cpp is not an issue.

I think it became necessary because I moved "renderer.h" at first include and something inside there is defining the _WIN32_WINNT to a low value. 

I will put the minimum value required for the macro.

tangtang95: Actually I will try to find a way to avoid putting this macro so that it will be the same.

tangtang95: I removed that macro and also put renderer.h below as I have changed the include inside the renderer.h file. It makes more sense to import common.h first since it contains common stuff.

julianxhokaxhiu:
Before this patch:
>[driver] Build completed: 00:02:09.445

After this patch:
> [driver] Build completed: 00:01:36.022

From a first quick check I can confirm the drop of nearly 40s in build time which is massive! I'm going now to have a check on both FF7 and FF8 and see how it goes.

src/common.cpp
julianxhokaxhiu: Thanks, I think the define would have had no impact anyway but better not have it around hardcoded. Appreciated!

julianxhokaxhiu (approved):

tangtang95:
> Had a quick run on FF7 and FF8 as well, all seems to work fine as usual, so no downsides detected. Well done! Let me know if the PR is ready to be merged or if you aim to do some more changes.

No more changes from my side!

PR #734, opened 2024-09-20
FF8: Fix app_path configuration for movies

myst6re:
## Summary

- Move app_path back to movies, but only if the redirection failed.
- Document the redirection return values, and force redirect_path_with_override to return 0 only when a redirection happened

### Motivation

file redirection works on relative paths, so @julianxhokaxhiu moved the app_path outside the movies logic to the redirection logic.
But if we do this, redirection will be relative to app_path, which I don't want, because app_path is reserved for game data, not mods.

### ACKs

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -9,7 +9,7 @@
 ## FF8
 
 - Fix crash on JP version ( https://github.com/julianxhokaxhiu/FFNx/pull/733 )
-- Movie: Fix `override_path` support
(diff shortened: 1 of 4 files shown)


julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you, this should hopefully settle the conflict we had with this new layer on top of the redirection.

PR #737, opened 2024-09-28
FF7: Add native support for japanese text rendering

### LINKED ISSUE #39: [ FF7 ] Japanese support

image

CosmosXIII:
## Summary

Adds an option to enable japanese text rendering for the japanese version of the game.
Right now field text only works if the jp flevel is re-exported using Makou Reactor.

Related #39 

### Motivation

To end the pain of the japanese nation. 

### ACKs


Changelog.md
```diff
@@ -13,6 +13,7 @@
 ## FF7
 
 - Core: Add missing Cloud models in the model to eye mapping list
+- Rendering: Added option for japanese text rendering ( https://github.com/julianxhokaxhiu/FFNx/pull/737 )
```
julianxhokaxhiu, line 16: Can you please move this into the Next section?

CosmosXIII: Done, somehow I always put it in the wrong place lol

julianxhokaxhiu:
Tested the PR on top of the native Japanese eStore release and first of all, well done :) I'm really impressed by this achievement!
I mainly noticed some bugs while using it, at least visually, I'll try to list them here so we can have a look together on how to fix it in case.

1. Save slots are not aligned with the hand, here an example:
JA:
image
EN:
image

2. Location text seems to be wrong/broken(?) I can see this in both save menu as well as in-game menu:
image
image

3. Loading text seems also to be broken(?)
image

I'm not sure if the context of those makes sense as I don't know the language, so any help on your side to check that is appreciated :)

Thanks a lot and looking forward to this really promising PR!

P.S.: I changed a bit the logic on how we enable this layer. Your original flag has been removed in favour of `ff7_japanese_edition` as it was already there since the beginning of the code, and we auto-enable it when we detected the japanese version of the game. However I left it also as a flag in the toml in case we might need to enable this manually because the auto-detection fails. Hope this is fine for you, let me know in case. Finally, I updated a bit the changelog message to fix the overall mission of this PR, let me know what you think.

CosmosXIII:
Thanks for testing it.

> 1. Save slots are not aligned with the hand

For the iro I was replacing the exe text with hext so I just added a few spaces to fix this.
By the way the english exe text also has 3 spaces like this: "Save 1   "
I'm not sure why but the japanese one did not have spaces so my guess is they hardcoded the spacing somewhere in code...

> 2. Location text seems to be wrong/broken(?)

I think this is because of the broken flevel. Once I fixed that with Makou the location also worked fine for me.

> 3. Loading text seems also to be broken(?)

The text here actually is correct but it's going outside the box.
So I think the fix here is to resize the box so that it fits the text but no idea where this is done.

canesalato:
Thank you so much for this work Cosmos!
Will this require a repurchase of the game? Japanese version? Or will the steam version be able to be converted to support Japanese language, like with FF9 and memoria?

CosmosXIII:
> Thank you so much for this work Cosmos! Will this require a repurchase of the game Japanese version or will the steam version be able to be converted to support Japanese language, like with FF9 and Memoria?

Unlike FF9 the JP data is not included  in the Steam release but it's technically possible to make it work with the Steam version if you provide the JP data in an IRO mod. Not sure yet if we it will go in that direction though.

zaphod77:
the android version seems to have broken flevel files too.  even the english one. at least the version I got.

the windows store version does not have broken flevel files.

How exactly did you do the makou reactor fix? i tried editig a text so makour rector would save, and it didn't fix a thing.

julianxhokaxhiu:
> How exactly did you do the makou reactor fix? i tried editig a text so makour rector would save, and it didn't fix a thing.

Please do not hijack this PR with offtopic questions. If you need support or help, use Discord support channels. Thank you

zaphod77:
the new ffnx resolves all broken lgp file issues including the one that's messing with this. the jp e-store release jfleve.lgp should read fine now, like the android version.

julianxhokaxhiu:
@zaphod77 that was never the issue blocking this PR. Read above what was pending for.

CosmosXIII:
> @zaphod77 that was never the issue blocking this PR. Read above what was pending for.

I think zap is referring about what I wrote in the PR summary about field text not working with unmodified flevel unless it’s re-exported with MR. I’ll check later to see if it works.

canesalato:
I have been testing this PR with the mobile version of the game.

To start with, I can confirm that, before the latest changes @zaphod77, field text was completely broken. It used to display actual Japanese characters, but making meaningless words with it. I am very pleased to report that everything seems to be working well now. The game plays just fine in Japanese. I checked all the main menus, tried different resolutions, 16:9 vs 4:9, loaded different save files and, so far, it seems to be working very well. Also I did not see major misalignement or issues (with 2 exceptions see points 1 and 4).

Concerning the points (1-3) raised by @julianxhokaxhiu:
1. This still happens, and I found it in the field as well (see attachment). Basically, everytime there is a selection to make, the cursor is slightly misaligned. It seems indeed to be related to some hardcoded spacing value.
screenshot (16)
screenshot (18)
2.  Fixed for me after @zaphod77 's latest commit! (see attachment) 
Screenshot 2024-12-05 122715
Screenshot 2024-12-05 122615
3.  Loading screen also looks fixed now!

Something I found, not previously reported:
4. An issue with text when displaying input keys: eg. playing the piano at Tifa's house (see attachment):
screenshot (17)
screenshot (19)
5. Finally, I have noticed that Japanese text seems to work well even with the japanese edition flag set to "false" in the configuration file. However, setting it to "true", seems to break text when loading the normal english executable.

canesalato:
One more misaligned element:
1. Gold Saucer, Mog video game, at game over
PXL_20241214_052643520 .
The bomb countdown is similarly misaligned.

canesalato:
Some updates that perhaps can be interesting. In my previous post I wrote that there seem to be an issue with displaying input keys (like, tifa's home piano). 

I would like to partially revise that statement. 
It does not seem to be just input keys, it also involves every text that has color. At first I missed this, because the english version seem to be missing much of the text color of the japanese version, so i missed that some words were supposed to be colored **in Japanese**.
You can compare Vanilla English and Vanilla Japanese:
PXL_20250204_142756835
PXL_20250204_133340501

And this is how it appears in the Japanese FFNx from this PR:
screenshot (8)
Not only the font is white but the セ from セーブ is missing. Display and spacing issues seem to be a constant with all colored text, which is generally included within square brackets [ ].  It just happens that input keys are the most common use-case for this style of text. Could the problem be that some colored font textures are missing?

Here is another example:
Vanilla JP:
PXL_20250204_133031230
PR:
screenshot (9)

Tested on top of both the Japanese estore release and mobile version

zaphod77:
Okay, i've done some investigation, and this appears to be square intentionally borking the files and fixing it in their driver.

Evidence.
Square E-store and mobile versions have borked lgp files that i figured out how to unbork.  100% proven
windows store has borked window.bin that ruins spacing for text. 100% proven by examination with hack7.

So it would not surprise me if the lang_ja\kernel\window.bin was also mangled, and unmangled in their driver, and that is the source of all the issues.

Anyway colors are done with $fe and a following byte, which in utf8 signifies a four byte character. The color codes appear to be put in unchanged.  meaning the encoding is probably not actually unicode at all.  

{xdd}âΩá©ÙÙo´o´ÙÙ
_s≠+âΣuõ{PURPLE}{xd7}z{xd0}$QåΠÜ{xd8}{WHITE}ŒŒ{xde}

this is the script for the "fabled save point".
{xdd} properly renders character in position DD from jafont_1, and everything is fine up to the purple.
what's puzzling is that {xD7} render properly, but the z afterwards, which corresponds to position {x5A) doesn't show. {xd0} then renders fine, and things are okay until the next color code, which again fails to insert the color change code, and then makes the character two after vanish. (the second $ae)

all of these hex characters in the script ARE single byte.

Okay i think i've cracked it.
characters normally pull from ja_font1.  ja_font1 has characters not normally existing in us_font_h or usfont_l.
FA=following byte pulls a character from ja_font2
FB=following byte pulls a character from ja_font3
FC=following byte pulls a character from ja_font4
FD=following byte pulls a character from ja_font5
FE=following byte pulls a character from ja_font6, EXCEPT if second byte is > $d1, then it's an argument for FE do normal processing of opcode. 

that said, ja_font1.tex has normal characters in positions that already seem occupied by opcodes. ALso ja_font5 has the same empty spaces as ja_font6

There may still be more to this.  I'd assume that anything that is blank in the bottom section of a ja_fontx.tex file may have a special meaning unless it's the empty space before the right pointing arrow, which had the heart in 1.0 JP.  There might be more jp only escape sequences present beyond these guidelines.

zaphod77:
The jp font textures don't have palettes like the us font textures at all. yet they are somehow colored in the square driver.

So that appears to be the issue. it tries to load palette 2 or 4 or 6, and they just are not there in the .tex file. instead, the square driver does the coloring inside the driver in jp rendering mode.  to do coloring properly, ffnx will have to do the same.

also i had a look at japanese_text, and it seems you are no longer performing the color processing.

the code that does different stuff depending on what the byte after FE starts at line 597, and is commented out. nothing else sets ff7_externals.word_91F028, which is later read to change the color.

so i think code from there needs to be moved up into the case at line 546.

NOTHING after d1 reads from ja_font6, because there is nothing after that in that texture. period. correct thing to do is set kanjidetected to false if >d1, and send stuff out to the opcode parser. 

https://github.com/CosmosXIII/FFNx/blob/e0aa1f8e69e13efbad08b21d3006aa03b0cefadb/src/ff7/japanese_text.cpp#L546

The issue is here.

After advancing and before changing the character sheet, it needs to check *buffer_text, and only if it's less then 0xD2u, do what it's doing.

Else, it should decrement buffer_text to point back at the 0xFEu, and then fall through to default;

Then come the case bit that got commented out.

0xFAu-0xFDu cannot happen, and are correct to remove from the case statement, because they were handled above.  but 0xFEu IS possible.  if it gets here, then it's an opcode, not a JP character. delete this entire if block

 if ( *buffer_text < 0xD2u )

as it can't happen.
 
If I could just build this version, i'd test the code myself, but i get cmake errors every time i try to clone the branch.

zaphod77:
okay, i'm right about the color. :)

My version.

image

matches the switch screenshot exactly.

The other layout issues still happen, though.

it seems the new code screws up references to btl_win in field script, which is what's causing the messups with prompting.  Someone who knows more than i do about debugging needs to trace this.

my commit here.

https://github.com/zaphod77/FFNx/commit/3322ee873410986c75ca5715b066d4e88a78ae2c

zaphod77:
Okay, if you change the /512.0 to /256.0 it fixes the prompt graphics to read from the right spot in btl_win a and b.

If you then delete all the -4.0s from the prompt rendering section, and change 

character_x += 32; 

to

character_x += 16;

that fixes the spacing for the built in prompts to "close enough"  as the graphics for those unused prompts arew left aligned, it will need some tweaking. 

This leaves the font spacing/positioning and the name screen issues.  They injected SOMETHING to significantly change the name screen, and i have no idea how they did it.  But hey, progress!

canesalato:
This is amazing! Thank you so much zaphod77!
________________________________

zaphod77:
more fun.

image

Unfortunately, will probably need to reverse engineer the af3dn.p from jp estore release to work out what the @$@#% they did for the naming screen, and to solve the rest of the issues.  Even if i knew how to decompile a driver, i don't have the JP e-store release, and it can't be bought now.

canesalato:
The 2026 Steam re-release includes Japanese language  support. Is their af3dn.p different from the JP e-store?


________________________________

zaphod77:
there is no af3dn.p  it's all built into their 64 bit FFVII.exe

the jp estore release is the only one that actually did it in the af3dn.p

My commits are here.

https://github.com/julianxhokaxhiu/FFNx/commit/3322ee873410986c75ca5715b066d4e88a78ae2c

https://github.com/julianxhokaxhiu/FFNx/commit/7e3b7743e87b43fbf4775f995dc6af5a3c37ec31

julianxhokaxhiu:
@CosmosXIII would you be open to review the commits above and merge them in your branch if you see them fit? Thank you

zaphod77:
More work.

I now have the new prompts supported, reading from their intended spots in btl_win.

image

Pardon the weird look of them, but that's what the graphics are.

my fork is here.

https://github.com/zaphod77/FFNx

oops, had buttons 6 and 7 swapped. correct now. :)

Please review.

Faults:
1) this ignores what the actual controller mapping is. it should check externals to see which button number is mapped to what, and decide what to show based on that, like the real game does for the four face buttons.
2) the JP smalltext in battle/ pause menu is misaligned. it's too high.
3) the menu text (and the button prompts to go with it) is too small. it looks like it should be amplified by 1.25 or so, and spaced out horizontally to match, with vertical spacing unchanged. this will probably fix all the pointer misalignment issues.  It still won't fix the rename screen, though.  I haven't figured out how to do that yet.
4) rename screen is still broken. this is a complicated code patch which needs to be reversed from square's driver from the jp estore version.

zaphod77:
Confirmed. ALL jp text out of those font text files should be scaled 1.25x before printing, with spacing between adjusted to match.

The auto resizer also needs it's adjustment for width for number of characters adjusted by the same amount.

This fixes all alignment issues in field scripts. the pointer points at the right spot.  I'm not sure of the actually correct way to do this.

And i now have what appear to be perfect field dialogs.
image

 common_submit_draw_char_from_buffer_6F564E_jp needs much more work, as it is used to print text of *different sizes*, and currently makes everything it prints the same size. it somehow aligns before i mess with it, but aligns horribly if i try to fix it. i'm not sure what the heck is going on there.

zaphod77:
Okay z_value is meant to scale the text somehow. but the larger it is, the SMALLER the text is.

This suggests that the text probably should be at 32 instead of at 16, and then scaled down by the z_value somehow.

zaphod77:
Okay patched the string length grabbing function at the bottom in my fork in such a way that the save file select screen matches the pointer.  It's a HORRID hack, but appears to work well enough.  

image

zaphod77:
Made a lot of progress.

Field dialogs seem to be properly sized and glitch free, with the help of a bit of code to shove dialog windows that are dangling off the left edge of the screen back on.

Top of screen text (mostly battle) seems to be all scaled up by 1.25x, same as the field stuff. However, only SOME of the menu text is scaled up.  When text is smaller, the bottom stays put, and the TOP is moved down, instead of the bottom going up.  I've implemented this in my fork now. Currently field dialogs are 20 units, and everything else is 16 units, because i haven't figured out the logic for making SOME menu text 20 and some 16.

Also currently jp text seems to show on top of everything else when other thing should be able to show on top of it, like the square iris for the battle UI.  The code is clearly setting the z value properly, so that should affect what's in front of what?

zaphod77:
Okay. so apparently the Square driver is scaling different texts differently, and applying tweaks based on *which section of the menu texts it comes from, and/or exactly which text it is*.  which i see literally no way to find out inside of common_submit_draw_char_from_buffer_6F564E_jp. All i can read is the z_value, and the color. which is not enough information to decide whether the text is scaled, and whether it is realigned manually. Here's an example or two.

The text in the main selection box in main menu is big.
Character names are big.
Item names are big.
translated "Weapon:"=small
translated "Armor:"=small
translated "Check"=small
translated "Arrange"=small
Description Windows are Big
Materia names are big
The LV/HP/MP text is big, even though most light blue text is small.
The text above the level and limit bars is small.
Time and Gil labels in that box are small.
Sort popup in item menu is small
Buy/sell/exit text in shop menu is big, and Sell and Exit are _manually realigned to match the finger spots_.
"GAME XX" translated text is big, and manually moved over to the left.
All small text is moved down 4 units, even when it shouldn't have been.
The centered top of battle text and the help text is big.  the battle commands are small.

I only have the x_value (which depends on the window being rendered to) and the text color to go on.  This is not a complete list.  Square seems to be checking when the text is copied into the buffer, and then changing the scaling factor based on the source of the copied text.  Which i see no way to actually do.

AAAARRRGGGGHHHHH!!!!

But hey at least dialog text seems correct, and i fixed the alignment for the save screen regardless. :)

zaphod77:
Okay i think i have now got the game playable with no alignment and cutoff issues issues in the field, and a few alignment issues in menus. this i because i still can't figure out how to tell if menu characters are supposed to be big, because the function there isn't passed enough info to deduce this. For the file select screen, everything comes from the same buffer, and everything is big. but in the slot selection screen, there's stuff that6's from the same buffer, but different sizes!

It seems exactly what section the text is sourced from in kernel.bin determines if it's big or small, and for text in the exe itself, they seem to be manually decided on. it seems every function that puts text into a buffer for use with the character drawing function needs to be found, and patched to set a flag saying either small or big somewhere that i can read it later.

canesalato:
I have extensively tested Zaphod latest work (on top of Cosmos) and there was a lot of back and forth with me reporting issues and Zap fixing this and that problem until things looked good.

As of now, beside the renaming screen, the result is a perfectly functional japanese FF7 that can be modded. The remaining alignment issues in menus are minor and the vanilla japanese version is not always perfectly aligned either.

If the naming screen cannot be fixed, I wonder if there is a way to select the default names and skip that screen entirely if lang=japanese. Then this could be merged and make Japan happy. 😁

Congratulations to Cosmos and Zaphod, thank you (and all FFNx contributors) for the work in making this a reality, you guys rock!

zaphod77:
This requires editing the field script, or actually replacing the opcode, because the naming screen is when the character names are assigned. before then, Cloud is named Ex-Soldier (or whatever it is in JP). I was able to properly skip the naming screen without crash, but then Cloud is Ex-Soldier for the rest of the game.

So instead of just skipping the screen, it must check which character is being named, and set the correct default.

This task should still be doable, though, as a temporary measure.

Note that the 2026 release detects other languages default names for characters and areas, and changes them on the fly in memory.  Whether FFNX should do the same is a matter of debate. 

All of this proves that the actual JP support in dotemu's releases is a horrible mismash of hacking, which is why we have to do the same.  My scale ad fudge factors are essentially what they also did, i just don't know how to gather the required information to set them properly yet.

I believe the correct approach is to track down the different parts of the menu modules, and, depending on which text source is used, set a global to signify if the next text is supposed to be big or small, and another global to adjust the x position of the text when needed to fix things.  Checking which buffer is used only tells you which menu module is being used, as the original text is copied to a new buffer. a combination of z-value and buffer address should reveal which window the text is in, btu even that is not enough, because some windows use mixed size. even checking color as well won't do it, and wont tell you when an offset to position is needed.  So it seems we need to hijack parts of the menu and set globals in them.

I have resynced my fork, in case anyone wants to look it over.

zaphod77:
I'm 100% okay with you taking the japanese_text.cpp from my file, modifying it, and improving this pull request.  Only reason it's not here is that I was unable to build your fork when i tried to fork it, ok?

pphouse2789:
A follow-up on the naming screen — the piece canesalato flagged as the remaining one ("beside the renaming screen, the result is a perfectly functional japanese FF7"), and one of the items @zaphod77 asked for help with in #925 (re-implementing the name screen): it's now fixed, restored as the full 3-mode (hiragana / katakana / eisuu) vanilla-style screen by rewiring the name-screen machinery still present in ff7_ja.exe. Details and before/after screenshots: zaphod77/FFNx#2. The one known limitation left on that screen is the ♥ glyph (see zaphod77's note in #925 about the empty slot at jafont_1 hex D8).

Following the direction suggested in both threads — patches flowing into @CosmosXIII's branch so this PR can continue — I've opened CosmosXIII/FFNx#14: a clean fast-forward of `japanese_text` (0 conflicts) that combines Cosmos's original work, zaphod77's fixes (all git authorship preserved), and the naming screen. I've taken @zaphod77's comment in #925 ("I'm 110% fine with Cosmos, or any other dev of FFNX taking what i've done... as long as Cosmos keeps credit") as an OK for doing this — if I've read too much into it, I'm happy to adjust or close that PR. As of today it also merges cleanly into current master, which would resolve this PR's conflicts. Happy to keep it rebased while we wait, and I defer entirely to @CosmosXIII and the maintainers on how and when to use any of it.

zaphod77:
Let me know when it's time to cancel my draft pull request.

As long as cosmos keeps all credit for what he originally did, i'm fine with whatever happens.  Any modifications to my own code to keep things up to standard, and fix issues i've created are 100% fine.  I have the same motive he does. Getting doublebyte characters working so japanese people can also enjoy upscales and interpolation, and other people can make their own multibyte character translations, which can't really be done with the old font system.

julianxhokaxhiu (dismissed):

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thanks a lot @CosmosXIII and well done again :)

TurquoiseHammer:
Mmyellow. Going about testing SAC's JP version and I've noticed the JP in non-field windows is a smaller point across the board. Here are a couple comparisons.

Main menu launching through Steam:
image
Main menu launching through 7H:
image

Steam shop window:
image
7H shop window:
image

I can test more locations if you'd like, but it seems to hold true for anything off the field.

zaphod77:
I have known this for a while. the game uses two sizes, and there's no way to work out which one is supposed to be used with the existing externals.  what seems to matter is the source of the text.

Some texts are small and some texts are big. color and what menu screen it is alone is not enough to guess the proper size.

TurquoiseHammer:
Hmm so are you saying it would just involve a lot of trial and error blindly setting font sizes to see if they increase the right thing?

zaphod77:
The issue is the function we are hooking isn't sent the information we need to make a decision on the text size.

https://github.com/julianxhokaxhiu/FFNx/blob/7717ef454936127b3b83b004b1103e31dd903e4b/src/ff7/japanese_text.cpp#L1114

This function only let's us access the character string, the position, the color, the z order, and the memory location the text is being read from.  *That's not enough information to determine the size*, because the text is **copied into a buffer first before it's sent to this function**, so we have no idea what we are actually trying to print.  We need to intercept the copying of the text to the buffer, and change the scale factor.  We know field dialogs are always big, so that works fine, but the other values have mixes of big and small.  I spent ages trying to be clever, and simply couldn't get anywhere, and had to make everything but some stuff we cold safely determine is big to small.  The original square code also applies alignment changes, and itself isn't even perfect.

I had hoped it was a L vs H difference (and that would have made some SENSE) but implementing checking that did absolutely nothing useful.

TurquoiseHammer:
Okay sounds messy. I'm not a coder, or at least not a skilled one, so forgive me if these are silly questions:
-This function handles all text drawing, right? And I assume from the parameters passed in, you can determine if it's field text, in which case you make it "big."
-So if it's menu text, would a potential stopgap be to see if it matches any of the strings we're expecting in certain menus? I guess this could work for some things like Items, Materia, Save, etc., but would get pretty out of hand if you want to check every single item name, every map name, etc. Also character names would be impossible. Damn, yeah, I guess this wouldn't work.

zaphod77:
NO i can't.  

the field dialogue one doesn't call that, so that's why it's big like it's supposed to be.  But most other battle menu and other menu stuff does call that function.   and given the choice between making everything big vs making everything small, i had to pick small so stuff will fit.

It not the string that determines big/small, square hooks the menu modules, and changes the size and alignment based on what the game is trying to print. whether it's a player name, status text, item name, etc.

For example sort popup and the three terxts at the upper left of use item menu are small, but the rest is big.

But the character stats menu is a MESS, and a mix of texts of different colors and sizes.

TurquoiseHammer:
Thanks, I'm getting a better understanding now. I appreciate you explaining it all. I guess what confused me is you said "This function only let's us access the character string, the position, the color, the z order, and the memory location the text is being read from." Is this information not enough to either a) change the scale factor based on the memory location or b) check if the string matches one of the small strings in the menu, such as つぎのレベルまであと?

zaphod77:
there are way too many strings to do that. it's the class of text, and the module that's printing it that matters.

For example, Item names in the item menu are always big.  

And changing the scale factor based on the memory location doesn't work, as i said, because the text is coped from its original source to a temporary buffer to use with the function. so i can't find out where the text was copied from.

Basically exe texts have to be assigned a size attribute that they don't actually have.  and while battle win and us font DO have two different versions, the jp font does not.

TurquoiseHammer:
Ah, okay I understand the memory location thing now w/r/t the buffer.
Last thought: what if you did it the other way around—check for things that should be small and make the rest big. It seems like there are fewer of those, and they're never variable as far as I know (character names, items, descriptions, etc). (And I'm aware that this is awfully hacky. Was just thinking as a temporary measure.)

zaphod77:
It's very annoying.  You need to check all menus in the jp pc version to see how crazy some of those screens get with mixed text.

TurquoiseHammer:
I forgot that the spell names in the bottom left panel of the materia submenu are small text. Yeah that would be hundreds of strings to manually check against... Well thanks all the same. Let me know if there's anything I can do to help, even if it's something brute-forcey.
I've got a couple other questions about text-related stuff, hopefully more straightforward ones with simple fixes.

First, on certain screens like this one, FFNx doesn't pad the text the same way the official port does.
FFNX: 
Official port: 

Of course, the ideal would be perfectly centered like the PSX, but that will have to come if and when we can implement a wider font


These lines are just manually spaced in the flevel, so I wonder if something is reducing that whitespace here.
image


Next, and possibly the same issue with whitespace getting eaten, lines in FFNx JP aren't flush with the inside of opening quotation marks like they're supposed to be.
FFNX (ignore that heart symbol, that's a different problem): 
image
Official port:
image

zaphod77:
You've actually found a bug somehow. this was only supposed to put in the heart if it was on the first ja_font.

TurquoiseHammer:
Oh haha, I thought it was SAC related.

zaphod77:
made a pull request to fix it

TurquoiseHammer:
Ok I looked into it a little more and the issue is that in FFNx JP, the space (　) character is rendered too narrow. This throws off the spacing on dialogue choices and other flashback dialogue/internal monologue stuff that uses that character to pad text. Fixing that should fix all these problems at once.

This is how a standard four-space-padded dialogue option should look:
image
And this is how it looks in FFNx right now. Not a big issue, but sometimes event planners only use three spaces, and this causes the cursor to overlap the text on FFNx.
image

zaphod77:
that should be simple as going to the width table that starts at line 320, and editing the number for the space.

zaphod77:
Okay, setting space at 31 width is too big for sure. 24 seems closer

also there's not enough padding overall.

You can see that the text strings look longer in the official port in the field dialogs.

TurquoiseHammer:
That's because dialogue window autosizing is turned off in the version I'm testing, so the box is wider. The text lines up perfectly otherwise, aside from the tiny bit of missing space at the beginning of the line.

zaphod77:
No, that's not it.  i'm talking about the actual text. it seems the max width is supposed to be 32, not 31.

TurquoiseHammer:
Oh, do you mean whatever is happening here?
image
(Longer text is official port)

It looks like the character width on っ and た is every so slightly narrower

zaphod77:
It's obvious in the meeting screenshot.

ter's a little more padding between characters.

the impression i get is that all non zero padding values have to have one added to them to fix this.

TurquoiseHammer:
Here's a better comparison of the meeting room text windows:
image

second and third lines, making up for reduced space character:
image

So that's just getting offset by one pixel per character?

zaphod77:
looks like. but fixing that with a patch i made throws off the auto resizer.

so i have to patch that too.

but THAT will break the arabic, which expects the current behavior.

TurquoiseHammer:
Imo it's not a huge issue to leave it as-is. The auto sizing accounts for it, right? And it makes no difference to legibility. Plus, ideally at some point we will introduce a whole new JP font that works more like the original game's.

zaphod77:
Not doing the offset DOES affect legibility.

And not adjusting the autosize to treat characters as one wider than they are if <32 causes it to not move the right side of the box enough.

This is VERY annoying.

Someone with better reverse engineering chops than me needs to work out all the alignment stuff that square is actually using.   But i believe with my latest pull request things are probably "good enough". in the future someone should be able to make a mod that actually works right with FFNX

TurquoiseHammer:
Okay here is what confuses me though. If it were a padding issue between characters, wouldn't the third line here offset at the same rate as the second? And yet all the ドs line up 
image

zaphod77:
you are right.  it seems square is just doing all sorts of weird stuff and hacky alignment fixes.  sigh...

julianxhokaxhiu:
Please move on the conversation on the issue, which I've also left some notes which you might find interesting: https://github.com/julianxhokaxhiu/FFNx/issues/39#issuecomment-5348104829

TurquoiseHammer:
Confirming the 値 character no longer results in a heart!
image

Issue #746, opened 2024-11-03
[ FF8 ] Regession with blending type 2 in external textures


myst6re:
## Describe the bug

Since FFNx 1.20.x, blending type 2 (substractive) does not work on external textures.

## To Reproduce
Use any existing mods using the Tonberry layer for field and go to the BGU elevator.

## Expected behavior
See https://github.com/julianxhokaxhiu/FFNx/issues/693

## Additional context
Before 1.20.x, the Tonberry layer was emulated via the TexturePacker (CPU side), the built texture was with alpha=127.
For some reason the GPU driver needs textures to have alpha=127 for blending type 4 to work.

myst6re:
Two options to fix this bug:
- revert the change made in 1.20.0 and emulate the Tonberry compatibility layer on CPU side. Textures will be slower to load
- find why textures need to have alpha=127, and patch FFNx to remove this prerequisite

julianxhokaxhiu:
As far as I am aware that value comes from the game engine, as we inherit whatever is being sent to use and we compute it against `255.0` so might be something deep in the game to find out.

However this might be a good starting point to investigate in case:
https://github.com/julianxhokaxhiu/FFNx/blob/master/src/common.cpp#L2206-L2224

I'll see if I find more.

myst6re:
Maybe not related, I found the patch you've made long time ago to enable alpha on the elevator: https://github.com/julianxhokaxhiu/FFNx/commit/3315ebe7bf7eda51bc7af4edfb102fa9c1284904

myst6re:
Maybe not related, I found the patch you've made long time ago to enable alpha (not the correct formula) on the elevator: https://github.com/julianxhokaxhiu/FFNx/commit/3315ebe7bf7eda51bc7af4edfb102fa9c1284904

Edit: I have a lot of GEQUAL with alpharef=112 in the elevator field.

julianxhokaxhiu:
I think I might start to understand what is going, great finding.

Modulate Alpha actually does this in the shader code ( https://github.com/julianxhokaxhiu/FFNx/blob/master/misc/FFNx.frag#L274-L279 ) which simply multiplies the color by the pixel color value. Disabling it simply picks Alpha as it is, which might explain why I got it working back then as alpha didn't had to be "multiplied" by the vertex color when drawing the door texture, but simply inherited.

Not sure if this intel helps, but might help us on crack this specific issue.

myst6re:
[...]

PR #749, opened 2024-11-10
FF8: Fix elevator blending for mods #746

myst6re:
## Summary

- Fix elevator blending for mods
- Fix crash when more_debug option is enabled -> replacing OutputDebugStringA was a bad idea and leaves the game in a undefined state

2024-11-10 22_56_52-Final Fantasy VIII _ FPS_ 30,0

### ACKs

### DIFF
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -903,11 +903,6 @@ int common_create_window(HINSTANCE hInstance, struct game_obj* game_object)
 				if (more_debug)
 				{
 					replace_function(common_externals.debug_print2, external_debug_print2);
-
-					if (ff8)
-					{
-						patch_code_dword(ff8_externals.outputdebugstringa, DWORD(external_debug_print));
-					}
 				}
 
 #ifdef NO_EXT_HEAP
--- a/Changelog.md
+++ b/Changelog.md
@@ -9,8 +9,10 @@
 ## FF8
 
 - Core: Fix crash for non-us versions introduced in 1.20.3 ( https://github.com/julianxhokaxhiu/FFNx/pull/741 )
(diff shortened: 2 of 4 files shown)


julianxhokaxhiu (dismissed):

src/common.cpp
```diff
@@ -903,11 +903,6 @@ int common_create_window(HINSTANCE hInstance, struct game_obj* game_object)
 				if (more_debug)
 				{
 					replace_function(common_externals.debug_print2, external_debug_print2);
-
-					if (ff8)
-					{
-						patch_code_dword(ff8_externals.outputdebugstringa, DWORD(external_debug_print));
```
julianxhokaxhiu, line 909: What is the reasoning of this removal other than fixing the bug? Shall we maybe hook the sub or do we lose any functionality here?

myst6re: I don't know how to fix the bug, I though replacing the dynamic address of OutputDebugStringA was a good idea, it works for the first call, and then the program is in a undefined state

julianxhokaxhiu: Ok, let me have a look at this after work and I'll let you know if I find a better way to fix it other than removing the patch itself :) Thank you

myst6re: You can quickly reproduce the bug/crash by launching the game and quickly after change to another application window

julianxhokaxhiu: Allright, tested but I'm not able to find any possible solution atm. Guess we'll merge the patch as it is, to eventually unpack this eventually another day. Thanks for catching this bug :)

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #750, opened 2024-11-12
FF8: Bring back shadows and tinted window

### LINKED ISSUE #693: [ FF8 ] Transparency appears really dark when using custom background textures
**Describe the bug**
When using custom background textures (even ones dumped directly from the game using save_textures = true), the transparency (alpha?) on things like the elevator or transparent screen that Ellone talks to squall through in the intro are much darker than default.

**To Reproduce**

1. Set save_textures = true and start a new games
2. After the doctor talks to Squall and the girl appears on screen take a screenshot
3. Once you have full control go down the elevator outside the classroom and after you step out take another screen shot.
4. Set save_textures = false and start a new game.
5. Repeat 2 + 3
6. Compare both screenshots

**Expected behavior**
Screenshots are identical

**Actual behavior**
Screenshots appear to to be identical except for the transparent surfaces.

**Screenshots**
FF8 no custom textures
ff8 custom textures

**GPU (please complete the following information):**
- NVIDIA
- driver 551.86
- driver 552.44

**Additional context**
- tested with 1.19.1.0
- mcdingus' patreon angelwing textures showed the same issue.
- I tried the latest canary - 1.19.1.34 which showed the same issue with dumped textures (pixel perfect)
- 1.19.1.34 with mcdingus' textures were worse than 1.19.1.0 as they turned pitch black where there should be transparency.

myst6re:
## Summary

Fixes #693 and more

**Bring back shadows and tinted window**
2024-11-11 22_25_03-Final Fantasy VIII _ FPS_ 30,0
_Elevator window is darker_
2024-11-13 00_01_53-Final Fantasy VIII _ FPS_ 15,0
_Shadows are back_

**Battle transition + shadows**
https://github.com/user-attachments/assets/674a7e17-cfd7-48ae-b84d-d32bb228cd5a

**Field transition**
https://github.com/user-attachments/assets/f26643be-1462-439a-bd11-076ec4b18471

### Motivation

Substractive blending is enabled only if the graphic driver is software and use paletted textures.

### Regression

Unfortunately Battle transition (from worldmap at least) suffer from this change

### ACKs

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -13,6 +13,7 @@
 - Ambient: Fix Battle ID detection for random encounters in Field
 - Modding: Allow modding card names hardcoded in exe ( https://github.com/julianxhokaxhiu/FFNx/pull/739 )
 - Modding: Add compatibility to LZ4 compression in FS archives ( https://github.com/julianxhokaxhiu/FFNx/pull/741 https://github.com/julianxhokaxhiu/FFNx/pull/743/files )
+- Rendering: Add support for substractive blending ( https://github.com/julianxhokaxhiu/FFNx/pull/750 )
 - Voice: Add support for Worldmap question dialogs
 - Voice: Add support for Auto text in World messages
 
--- a/src/ff8.h
+++ b/src/ff8.h
@@ -1059,7 +1059,9 @@ struct ff8_externals
 	uint32_t get_disk_number;
 	char* disk_data_path;
 	const char *app_path;
+	uint32_t swirl_enter;
 	uint32_t swirl_main_loop;
+	uint32_t sub_460B60;
 	uint32_t field_main_loop;
 	uint32_t field_main_exit;
 	uint32_t psxvram_texture_pages_free;
@@ -1068,6 +1070,8 @@ struct ff8_externals
 	uint32_t engine_set_init_time;
 	uint32_t sub_4672C0;
 	uint32_t sub_471F70;
+	uint32_t field_fade_transition_sub_472990;
+	uint32_t sub_45CDD0;
 	uint32_t sub_4767B0;
 	uint32_t sub_4789A0;
 	char (*sub_47CA90)();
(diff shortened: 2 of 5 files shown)


src/common.cpp
```diff
@@ -2866,8 +2866,11 @@ __declspec(dllexport) void *new_dll_graphics_driver(void *game_object)
 	// game-specific initialization
 	if(!ff8)
 		ret = ff7_load_driver(game_object);
-	else
+	else {
+		VOBJ(game_obj, game_object, game_object);
+		VRASS(game_object, current_gfx_driver, 1); // Force software rendering game hacks
```
julianxhokaxhiu, line 2871: This bit is set here as well https://github.com/julianxhokaxhiu/FFNx/blob/master/misc/FF8.reg#L24-L40, which is suggested to run before using FFNx. If you change the settings on the FF8Config GUI for example the value is going to be overridden though. For Steam, this is always enforced as we hard code the registry value via the dotemu APIs.

myst6re: I will check that thanks. For me the gfx_driver id cannot be enforced by registry in ff8, FFNx forces to 2 anyway, because that's the easiest way to make the game run the custom driver function here

myst6re: The reason I force 1 after is that I found several places in ff8 where current_gfx_driver is testes with the hardcoded value of 1

myst6re: I checked and ""Graphics"=dword:00100021" does not affect the current gfx driver value at all.

And I confirm that FFNx sets this value to "2" itself here: https://github.com/julianxhokaxhiu/FFNx/blob/master/src/common.cpp#L3471

julianxhokaxhiu: Ah ok maybe I remember wrong, so let's change the line over that function to 1 for FF8 at this point if it brings better results visually. Thanks for going this deep, appreciated

julianxhokaxhiu:
Spectacular PR, well done :) Thank you!

PR #760, opened 2024-12-04
Re-enable loading broken field LGP files once again.

zaphod77:
## Summary

FFNX now works with android data files once again, even though the lgp files are broken.

### Motivation

I followed the android instructions. they didn't work.  I figured out why, and coded a fix.  Since most versions of jfleve.lgp are broken, I figure this is useful for those wishig to support japanese.

### ACKs

- [*] I have updated the [Changelog.md](https://github.com/julianxhokaxhiu/FFNx/blob/master/Changelog.md) file
- [*] I did test my code on FF7
- [ ] I did test my code on FF8 (my pull request is ff7 specific)

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,7 @@
 # Next
-
+## FF7
+- Allow loading of field lgp files that have been mangled by Square to break modding.
+  
 - Full commit list since last stable release: https://github.com/julianxhokaxhiu/FFNx/compare/1.21.0...master
 
 # 1.21.0
--- a/src/ff7/file.cpp
+++ b/src/ff7/file.cpp
@@ -638,7 +638,7 @@ int ff7_read_field_file(char* path)
 
 	// Attempt to override part of the current field file with chunk files
 	char chunk_file[1024]{0};
-	int data_ptr = 0, data_len = 0;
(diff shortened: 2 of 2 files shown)


zaphod77:
All unmodified version of ff7 and square's drivers appear to ignore that particular header, so we need to ignore it as well to gain compatibility with all square supplied files and still support field chunk loading. If the section pointer is wrong, then the file won't work with vanilla, so it's correct to trust the section pointers.

julianxhokaxhiu:
Very good discovery, I double checked that the calculations match also the correct headers for eg. in 1998 so this approach is really preferred as it works across all FF7 editions, including the Android port. Well done and thank you!

julianxhokaxhiu (approved):

zaphod77:
Should probably compare what's there with the truth, and log the malformed lgp file to ffnx.log, but i just wanted to get stuff working. I'm sure you are up to the task?

Yeah, my approach is foolproof in that anything that works with any other existing version of the game will also work with ffnx now, at the expense of allocating padding bytes, which is what obeying the header is supposed to prevent.  it's possible to program a seek back and shrink feature to remove padding bytes, but i wanted to keep it simple. :)

PR #764, opened 2024-12-11
FF7: allow filtering of internal textures that were originally filtered.

### LINKED ISSUE #754: Add field and battle texture filtering options.
## Summary

Original steam release allows you to apply bilinear filtering to the field and battle textures.  This is impossible in FFNX as near as I can tell.

### Basic example

enable_bilinear option exists, but only seems to affect the field models, as per the original pull request (moving trains in the trainyard, for example).  The main low definition field images are still shown using nearest neighbor, and the battle textures are blocky still.

### Motivation

Provide feature parity with the steam version. currently even with no mods, we need to use FFNX to get proper controller support for xinput controllers, which causes loss of the bilinear filter option that the original game has.

zaphod77:
## Summary

This restores the missing filtering from all the 3d models when using the internal textures.  

### Motivation

This restores a missing feature that ffnx removes.  FFNX was refusing to filter any internal textures.  Now it only refuses to filter the ones that aren't safe to filter, as far as I know.

I'm aware this code is ugly and needs gardening, but i've tested it extensively (checked all the minigames), and it now appears to be at the point that it manages to replicate the look of the unmodded PC version when using the original assets.  

### ACKs

- [x] I did test my code on FF8 (does nothing to ff8)

Closes #754 , in my opinion.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,10 @@
 
 - Full commit list since last stable release: https://github.com/julianxhokaxhiu/FFNx/compare/1.21.3...master
 
+## Commmon
+
+- Music: Add the ability to play a custom `main_menu` music file on the new game screen.
+
 # 1.21.3
 
 - Full commit list since last stable release: https://github.com/julianxhokaxhiu/FFNx/compare/1.21.2...1.21.3
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -1246,6 +1246,9 @@ void common_flip(struct game_obj *game_object)
 	// Update day night time cycle
 	if (!ff8 && enable_time_cycle) ff7::time.update();
 
+	// Handle main menu background music
+	handle_mainmenu_playback();
+
 	// FF8 does not clear the screen properly in the card game module
 	if (ff8)
 	{
--- a/src/ff7_data.h
+++ b/src/ff7_data.h
@@ -50,7 +50,7 @@ static struct game_mode ff7_modes[] = {
 	{FF7_MODE_BATTLE_MENU, "MODE_BATTLE_MENU", MODE_MENU,        true  },
 	{FF7_MODE_UNKNOWN18,   "MODE_UNKNOWN18",   MODE_UNKNOWN,     false },
 	{FF7_MODE_EXIT,        "MODE_EXIT",        MODE_EXIT,        true  },
-	{FF7_MODE_MAIN_MENU,   "MODE_MAIN_MENU",   MODE_MENU,        true  },
+	{FF7_MODE_MAIN_MENU,   "MODE_MAIN_MENU",   MODE_MAIN_MENU,   true  },
 	{FF7_MODE_UNKNOWN21,   "MODE_UNKNOWN21",   MODE_UNKNOWN,     false },
 	{FF7_MODE_INTRO,       "MODE_INTRO",       MODE_INTRO,       true  },
 	{FF7_MODE_SWIRL,       "MODE_SWIRL",       MODE_SWIRL,       true  },
@@ -121,6 +121,7 @@ inline void ff7_find_externals(struct ff7_game_obj* game_object)
 	ff7_set_main_loop(MODE_CREDITS, credits_main_loop);
 	menu_main_loop = get_absolute_value(main_loop, 0x62E);
 	ff7_set_main_loop(MODE_MENU, menu_main_loop);
+	ff7_set_main_loop(MODE_MAIN_MENU, menu_main_loop);
 	battle_main_loop = get_absolute_value(main_loop, 0x89A);
 	ff7_set_main_loop(MODE_BATTLE, battle_main_loop);
 	field_main_loop = get_absolute_value(main_loop, 0x8F8);
@@ -299,7 +300,7 @@ inline void ff7_find_externals(struct ff7_game_obj* game_object)
 	ff7_externals.config_menu_sub = ff7_externals.menu_subs_call_table[8];
 	ff7_externals.menu_sub_6FEDB0 = ff7_externals.menu_subs_call_table[10];
 
-  ff7_externals.config_initialize = get_relative_call(main_init_loop, 0x4B0);
+	ff7_externals.config_initialize = get_relative_call(main_init_loop, 0x4B0);
 
 	ff7_externals.menu_tutorial_window_state = (BYTE*)get_absolute_value((uint32_t)ff7_externals.menu_tutorial_sub_6C49FD, 0x9);
 	ff7_externals.menu_tutorial_window_text_ptr = (DWORD*)get_absolute_value((uint32_t)ff7_externals.menu_tutorial_sub_6C49FD, 0x18);
--- a/src/music.cpp
+++ b/src/music.cpp
@@ -45,6 +45,31 @@ uint8_t ff7_last_akao_call_type = 0;
 uint32_t ff7_last_music_id = 0;
 int16_t ff7_next_field_music_relative_id = -1;
 
+void handle_mainmenu_playback()
+{
+	struct game_mode *mode = getmode_cached();
+	static bool is_main_menu = false;
+
+	switch (mode->driver_mode)
+	{
+	case MODE_MAIN_MENU:
+		if (!is_main_menu)
+		{
+			is_main_menu = true;
+
+			nxAudioEngine.playMusic("main_menu", 0xFF, 0);
+		}
+		break;
+	default:
+		if (is_main_menu)
+		{
+			nxAudioEngine.stopMusic();
+			is_main_menu = false;
+		}
+		break;
+	}
+}
+
 void music_flush()
 {
 	nxAudioEngine.flush();
--- a/src/music.h
+++ b/src/music.h
@@ -23,5 +23,6 @@
 
 #include <stdint.h>
 
+void handle_mainmenu_playback();
 void ff7_play_midi(uint32_t music_id);
 void music_init();


Changelog.md
```diff
@@ -1,3 +1,7 @@
+## FF7
+
+- Core: added option to filter what the original game filtered when still using internal textures.
+
```
julianxhokaxhiu, line 4: Please put this section in the right order after the Next. See other PR examples.

src/gl/special_case.cpp
```diff
@@ -55,7 +55,18 @@ uint32_t gl_special_case(uint32_t primitivetype, uint32_t vertextype, struct nve
...
+  // don't filter original textures in submarine minigame
+  if (current_state.texture_set && (mode == MODE_SUBMARINE) && !(VREF(texture_set, ogl.external))) // only ff7 has a submarine mode.
+  {
+    current_state.texture_filter = false; // don't filter original submarine textures, this causes thin lines, and original did not. 
+  }
+  // or coaster minigame
+  if (current_state.texture_set && (mode == MODE_COASTER) && !(VREF(texture_set, ogl.external))) // only ff7 has coaster minigame.
+  {
+    current_state.texture_filter = false; // don't filter original coaster textures
+  }
```
julianxhokaxhiu, line 68: For both conditions even though we know it's only for ff7, I'd appreciate if we could add `!ff8` nevertheless in the condition check. Additionally, please avoid over commenting ( stuff like `// only ff7 has a submarine mode.` is literally noise while looking at the code ), as well as `// don't filter original ... textures,`, just add notes of what it breaks in case it is set to true. No need to state the obvious :)

src/gl/special_case.cpp
```diff
@@ -101,11 +112,37 @@ uint32_t gl_special_case(uint32_t primitivetype, uint32_t vertextype, struct nve
...
-				// avoid filtering window borders
-				if(!_strnicmp(VREF(tex_header, file.pc_name), "menu/btl_win_c_", strlen("menu/btl_win_c_") - 1) && VREF(texture_set, palette_index) == 0) current_state.texture_filter = false;
-			}
+      if ((uint32_t)VREF(tex_header, file.pc_name) > 32)
+      {
+        // avoid filtering window borders
+        if (!_strnicmp(VREF(tex_header, file.pc_name), "menu/btl_win_c_", strlen("menu/btl_win_c_") - 1) && VREF(texture_set, palette_index) == 0) current_state.texture_filter = false;
+
+        // don't filter original menu textures in ff7, because it introduces glitches
+        if ((VREF(texture_set, ogl.external)))
```
julianxhokaxhiu, line 121: Maybe do the opposite check so you avoid having this huge empty block of nothing?

julianxhokaxhiu (changes requested):
Thanks for coming up with PR. I'm not really sure what are we fixing here, if this is just your preference, or we assume that what the Steam release did is the right one. In the meantime we unpack this, please look at the comments I left which we need to work nevertheless on top.

Overall I'm not really a fan of adding flags, or better only if needed, so I'm trying to understand what are we really trying to bring up here and which would be the best way to bring it to players.

Nevertheless, thanks for working on this, looking forward to it :)

Changelog.md
zaphod77: Sorry for taking time getting back to this. I got sick. still am but now i've actually spotted this.

The original 98 release applied filtering to all the 3d textures.  FFNX, well, doesn't. a blanket fix appears to have been put in to stop it just to remove all the graphical glitches when upscaling.

The steam release does too, but lets you disable it. That's why a new flag is needed, and the ability to let users choose it.

src/gl/special_case.cpp
zaphod77: I tried it the other way before and it wasn't cooperating.  It cold have been another issue. i did say it was ugly. will look into this.

src/gl/special_case.cpp
zaphod77: It seems we have different code documentation views.  Not saying you are wrong. as for leaving out the !ff8 check that's a matter of optimization, with an explanation of why i didn't bother to check. But i see the point. I should be able to refactor that code easily enough.

src/gl/special_case.cpp
zaphod77: okay change made.

To fully test this, have anisotropic filtering off, and bilinear filtering ON. This is the setting that revealed the most glitches.   I was unable to detect any more of them in my testing, but that doesn't mean there aren't any.

zaphod77:
both the original pc release, and the steam release under default settings look like this. t was hard to find footage of OG pc release, but here it is, on an emulated voodoo graphics card it was designed for.

https://youtu.be/LSkuTPRanuk?t=472

You can see the filtered buster sword texture, and the filtered battle textures.

In the steam release, you can disable linear filtering, which makes things look like FFNX currently does it.

Because both the steam release and the 98 cd release do this out of the box, i decided to make it a flag that defaults to true, with people who just extracted it into their steam version given the option to change it.

I trust i explained my reasoning?  If the original 98 release hadn't actually done this I wouldn't have bothered.

Also, at some point FF8 may end up being able to filter textures too, and similar work will need to be done there. i was unable to get ff8 to filter any internal textures in my testing, or I would have written similar code for that game as well.

I added the option because in my opinion, "the PSX version didn't do this" isn't a good enough reason to remove a feature.

src/gl/special_case.cpp
```diff
@@ -55,7 +55,16 @@ uint32_t gl_special_case(uint32_t primitivetype, uint32_t vertextype, struct nve
 	if(ff8) current_state.texture_filter = enable_bilinear && vertextype != TLVERTEX && current_state.texture_set;
 	else if (enable_bilinear && (vertextype != TLVERTEX || mode == MODE_MENU || (current_state.texture_set && VREF(texture_set, ogl.gl_set->force_filter)))) current_state.texture_filter = true;
 
-	// some modpath textures have z-sort forced on
+  // since we are no longer disabling all filtering on interal textures, we need to shut it off for the ones that filtering breaks.
+
```
julianxhokaxhiu, line 59: Is this new line intended? If not, please remove it.

src/renderer.cpp
```diff
@@ -705,8 +705,8 @@ void Renderer::bindTextures()
                             if (internalState.bDoMirrorTextureWrap) flags |= BGFX_SAMPLER_U_MIRROR | BGFX_SAMPLER_V_MIRROR | BGFX_SAMPLER_W_CLAMP;
 
                             if (internalState.bIsMovie) flags |= BGFX_SAMPLER_U_CLAMP | BGFX_SAMPLER_V_CLAMP | BGFX_SAMPLER_W_CLAMP;
-
-                            if (!internalState.bDoTextureFiltering || !internalState.bIsExternalTexture) flags |= BGFX_SAMPLER_MIN_POINT | BGFX_SAMPLER_MAG_POINT | BGFX_SAMPLER_MIP_POINT;
+                            
+                          if (!internalState.bDoTextureFiltering || (!internalState.bIsExternalTexture && !filter_internal_textures)) flags |= BGFX_SAMPLER_MIN_POINT | BGFX_SAMPLER_MAG_POINT | BGFX_SAMPLER_MIP_POINT;
```
julianxhokaxhiu, line 709: No external flags will ever be allowed to mess with the `renderer.cpp`. Please find a way to accomplish this using the renderer internal flags. Thanks

julianxhokaxhiu:
Starting from the fact that linear filtering is a visual preference, we don't have to do what the original driver does if we think it is not needed, especially because we allow those textures to be modded in the first place ( unlike the original game and driver which is intended to work only with the original assets ).

Because of this, I was asking if this was on PSX because that's the version we use as a basis to be the "most polished one" as we all know the history behind the PC port and how crap it was, so a lot of things were left over.

I understand you might want this feature in, but this is not an heavy argument for me to forcefully accept the PR. I am still open however to include it, but never I will accept statements like this:

> I added the option because in my opinion, "the PSX version didn't do this" isn't a good enough reason to remove a feature.

Whatever we do here serves the community, not me, not you. So here you have "my PSX argument". Again, I would like to know what are we really trying to solve here, before I merge this code. I hope you do understand.

On top of all of this, I also left some comments as there are still minor things to polish. Thanks!

src/renderer.cpp
zaphod77: Renderer.cpp itself was what was deciding "never filter internal textures, full stop. ever", so that's why i put the code there.  Just deciding that the user doesn't get to chose if those textures are filtered or not is just as bad one way as the other.  The flag being there is exactly what i intended.

That said. I have deleted that check from renderer.cpp, and attempted to move it to the end of specialcase, because s I understand the code right, that should apply it properly.

Will compile a new one and test once more for myself.

src/gl/special_case.cpp
zaphod77: Figured out what you meant. corrected.

src/renderer.cpp
zaphod77: Okay i moved the check, but when i do so it forces filtering on the square enix logo movie, which apparently skips special_case completely somehow..

So if you can tell me how to actually catch that case properly, i'll fix the code, but i have no idea HOW.

src/renderer.cpp
```diff
@@ -705,8 +705,8 @@ void Renderer::bindTextures()
                             if (internalState.bDoMirrorTextureWrap) flags |= BGFX_SAMPLER_U_MIRROR | BGFX_SAMPLER_V_MIRROR | BGFX_SAMPLER_W_CLAMP;
 
                             if (internalState.bIsMovie) flags |= BGFX_SAMPLER_U_CLAMP | BGFX_SAMPLER_V_CLAMP | BGFX_SAMPLER_W_CLAMP;
-
-                            if (!internalState.bDoTextureFiltering || !internalState.bIsExternalTexture) flags |= BGFX_SAMPLER_MIN_POINT | BGFX_SAMPLER_MAG_POINT | BGFX_SAMPLER_MIP_POINT;
+                            
+                          if (!internalState.bDoTextureFiltering) flags |= BGFX_SAMPLER_MIN_POINT | BGFX_SAMPLER_MAG_POINT | BGFX_SAMPLER_MIP_POINT;
```
julianxhokaxhiu, line 709: Please revert this line to what it was before, this PR shouldn't touch the renderer at all. It took us quite an effort to get it working for both FF7 and FF8. Thanks

src/renderer.cpp
julianxhokaxhiu: Left a comment on the same line down below. Closing this thread.

julianxhokaxhiu:
I understand your argumentation and I'm fine adding a new option on top. Let's find a way though to make it work without disrupting core parts as it's just a behavioral change.

About movies, please keep them as they are, it was like this since the days of Aali's driver and users are accustomed to it.

One more thing: I noticed all your indentation is using spaces. Please look careful on this PR change tab and check what is aligned with what was before ( for eg. if the file is using tabs, keep using tabs ). We'll eventually uniform all files to use spaces in a separate commit. Thank you.

src/renderer.cpp
zaphod77: But the renderer itself is what's hard blocking filtering of internal textures.

The pull request depends on me stopping this hard override.  If i don't do this, nothing else in mu pull request has any effect at all.

There's is literally no way to override this check. It was the entire reason turning bilinear filtering on did nothing before unless you were modding.

src/renderer.cpp
zaphod77: yeah, i can't figure out how to even check if the movie is external there or not. the code from special_case doesn't compile.

zaphod77:
> About movies, please keep them as they are, it was like this since the days of Aali's driver and users are accustomed to it.

Tell me how to place an override in for the internal movies, then. i can't figure it out. :)  !(VREF(texture_set, ogl.external)) shows intellisense error when it's inside of gl_draw_movie_quad_common

julianxhokaxhiu:
As stated above, please leave the `renderer.cpp` untouched. Find any other way to accomplish what you want to accomplish using texture flags. Then check where those flags are used to run the internal renderer class methods that toggle the filtering behavior. Thanks

zaphod77:
I didn't think lying by changing bIsExternalTexture was an option. how would i even do that?

zaphod77:
so seems this is dead code, or nearly so.

https://github.com/julianxhokaxhiu/FFNx/blob/7f3739e2ae873bec221adb1c565afbd3b09502da/src/gl/gl.cpp#L132

as ogl.external is only ever set true by load_external_texture, which doesn't appear to be called for movies as near as I can tell. do tell me if i'm wrong here. 

Which had me wondering if i had revealed a bug after all, since the source was saying that movies were getting filtered, and yet they clearly were not, with this very difficult to verify for anything but movies that are low enough resolution that you can actually tell.

zaphod77:
Okay. I figured out what's going on with the movies.
1) ffnx has always requested filtering for movies.
2) yet they have NEVER been filtered, because no movie is considered an external texture, and every texture not external was forced unfiltered.. 

By removing that check, all movies get filtered.

So if the blanket low level force filtering off for anything not an external texture is disabled for ff7, then the code that did nothing in draw movie quad now does something, and to restore previous behavior, it needs to set all movies to unfiltered.

I have checked aali's source and indeed it forces movies to not be filtered. so line133 in gl.cpp has been wrong, and always has been. This was chosen because the fields were also unfiltered, and the movies were low resolution. this way things wouldn't jump from blurry to not.

Problem is, square upscaled them to 1280x960, and that made them look blurrier. so we now have the blury to non blurry jumping again. either way, the filtering does pretty much nothing nowadays. but for people who still have their cd version, with the low quality duck truemotion movies, the behavior of having them unfiltered is correct.

To allow filtering internal textures, either the line blocking that in renderer.cpp needs changing or very "okay to filter" internal texture will need is "isExternalTexture" flag set. which is a lie. :)  removing the test, and then blacklisting internal textures that do not filter properly, as i have done for ff7, is 100% the correct approach for enabling the option to do what the early releases did.

The original releases did filter okay, but ONLY in the native resolution of 640x480.  And the filtering is invisible at that resolution for the troublesome textures.

So someone not me needs to remove the check, play through ff8, and blacklist the offending internal textures.

PR #767, opened 2024-12-21
FF7: Add option to remove top/bottom black bars on the screen

CosmosXIII:
## Summary

Added uncrop option to remove black bars from top/bottom of the screen. This can be used to uncrop the screen even when using 4:3 aspect ratio or for not widescreen extended screens in 16:9 mode.

### Motivation

So that people that prefer almost squared stuff can be a little more happy.

### ACKs

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -6,6 +6,10 @@
 
 - Music: Add the ability to play a custom `main_menu` music file on the new game screen.
 
+## FF7
+
+- Rendering: Add uncrop option to remove black bars from top/bottom of the screen ( https://github.com/julianxhokaxhiu/FFNx/pull/767 )
+
 ## FF8
 
 - SFX: Fix incorrect volume assignment when playing sfx effects using the external layer
--- a/misc/FFNx.toml
+++ b/misc/FFNx.toml
@@ -58,6 +58,11 @@ internal_resolution_scale = 0
 #~~~~~~~~~~~~~~~~~~~~~~~~~~~
 aspect_ratio = 0
 
+#[UNCROP]
+# Uncrops the image to remove the black bars on the top and bottom of the screen.
+#~~~~~~~~~~~~~~~~~~~~~~~~~~~
+enable_uncrop = false
+
 #[ENABLE VSYNC]
 #~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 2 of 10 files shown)


julianxhokaxhiu:
Thanks for the PR Cosmos, however is there any plan to bring this to FF8 as well? If not, I'd prefix it with `ff7_` and put it in the relative category down below in the toml.

Additionally, is there any other way we can think of enabling it without an additional flag? I know they are cheap but the less they are the better is the maintainability of the code. Thank you in advance.

CosmosXIII:
> is there any plan to bring this to FF8 as well? 

It should work in FF8 too mostly. The parts that are now FF7 only are there to fix some some backgrounds where the black bar reappear unless the camera range is corrected through the widescreen config. Now that we started FF8 widescreen implementation I think it would be better to keep the flag common or else we will end up with two separated flags for the same thing.

> Additionally, is there any other way we can think of enabling it without an additional flag? I know they are cheap but the less they are the better is the maintainability of the code. Thank you in advance.

At first I thought putting it into the aspect_ratio flag but then it would get more confusing because I want it to also work when using 16:9 aspect ratio without the extended backgrounds in Limit Break.

julianxhokaxhiu:
> At first I thought putting it into the aspect_ratio flag

I was wondering why can't we just add this functionality already in the existing 4:3 check instead, as vertical bars were added only on PC if I am not mistaken, while on PSX they were never there. Since the mission here is to bring most of the PSX look and feel to PC, I think we can safely assume this mode is what everyone wants really, without adding a further flag.

Since you can also govern which FFNx flag needs to be enabled via the mod you can enable or not the 16:9 based on the case if needed. Otherwise it will just work with whichever way users might pick in the Game Driver window.

What do you think?

CosmosXIII:
>as vertical bars were added only on PC if I am not mistaken, while on PSX they were never there

Actually, the black bars are there too on PSX. See video below:
https://youtu.be/h0mqRkhOvtY?t=328

In fact, while most backgrounds have image data there, a few of them do not. In those cases I had to extend those to fill the top/bottom bars in Limit Break.

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thanks a lot :)

PR #798, opened 2025-05-11
FF8: Initial implementation of some steam achievements

tangtang95:
## Summary

Partial Implementation of some FF8 steam achievements (More than 20s). Even if it is partial, better to split it into some pull requests and avoid huge changes. This functionality is under a debug flag so, users will not have a partial implementation

### Motivation

To be able to achieve steam achievements of FF8 using FFNx

### ACKs

- [x] I did test my code on FF8


.gitignore
```diff
@@ -398,3 +398,5 @@ cmake-build*/
 # vcpkg
 !make.exe
 vcpkg_installed/
+
+compile_commands.json
```
julianxhokaxhiu, line 402: Is this block needed? I can't see any file committed with this name. Thanks :)

src/ff8_opengl.cpp
```diff
@@ -1267,6 +1481,49 @@ void ff8_init_hooks(struct game_obj *_game_object)
 	// All possible message and ask windows
 	ff8_opcode_old_battle = (int (*)(int))ff8_externals.opcode_battle;
 	patch_code_dword((uint32_t)&common_externals.execute_opcode_table[0x69], (DWORD)&ff8_opcode_battle);
+
+	//###############################
+	// steam achievement unlock calls
+	//###############################
+	if(steam_achievements_debug_mode)
```
julianxhokaxhiu, line 1488: Is this flag name intentional or a left over?

julianxhokaxhiu, line 1488: Is this flag the right one being used here? Or did you mean `enable_steam_achievements`?

julianxhokaxhiu:
> This functionality is under a debug flag so, users will not have a partial implementation

Since we're still on a canary and a stable will take months before being released, I see nothing wrong in getting a partial implementation enabled by default with more to come as we go. FFNx is still WIP even if we do stable releases, so if you want to enable it ( given it won't provide an unstable gameplay ) I'm fine with it :) LMK

.gitignore
tangtang95: It's in gitignore to avoid being added into the git index. I have it locally because I used neovim

tangtang95:
> > This functionality is under a debug flag so, users will not have a partial implementation
> 
> Since we're still on a canary and a stable will take months before being released, I see nothing wrong in getting a partial implementation enabled by default with more to come as we go. FFNx is still WIP even if we do stable releases, so if you want to enable it ( given it won't provide an unstable gameplay ) I'm fine with it :) LMK

I am not enabling it to avoid users reporting some steam achievements not working since they might see some working. I would like to enable it once everything is implemented and then proceed for bug fixing based on user reporting

.gitignore
julianxhokaxhiu: Ah ok I see, let me add one more patch on top then, we're aligned with https://github.com/github/gitignore as this file is mainly a combination of different ignore files.

julianxhokaxhiu: Done, your line is now part of the default CMake ignore file :) I used this chance to update the others as well as adding the Jetbrain one to ensure wider IDEs support. Let me know if you're fine with this change. Thanks!

tangtang95: LGTM. Thanks!

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thanks a lot for this amazing contribution!

PR #813, opened 2025-07-14
Revise LUT shader code to avoid 1.0 coordinates that some graphics drivers may sample incorrectly.

ChthonVII:
## Summary

Implement the revisions to LUT shader code discussed in #777 to avoid 1.0 coordinates that some graphics drivers may sample incorrectly.

### Motivation

While #777 turned out to be unrelated to LUT sampling, this change may nevertheless be a good prophylactic against bad graphics drivers that treat 1.0 coord as out-of-bounds and/or don't honor clamp-to-edge.

### ACKs

- [no] I did test my code on FF8 - I don't have FF8 installed, but the behavior should be consistent with FF7.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -26,7 +26,6 @@
 
 - Core: Users can now choose on which screen to run the game, using the new `display_index` option.
 - Music: Add the ability to play a custom `main_menu` music file on the new game screen.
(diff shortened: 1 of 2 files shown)


julianxhokaxhiu (approved):

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #814, opened 2025-07-14
Replace unprincipled "toeless sRGB" gamma function for movies with BT1886 Appendix 1 function

ChthonVII:
## Summary

Replaces the default gamma function for movies (that lack metadata specifying a particular gamma function). Removes the unprincipled "toeless sRGB" gamma function. Replaces it with BT1886 Appendix 1 function with constants selected to match a properly calibrated mid-90s Sony Trinitron CRT.

(Note that BT1886 specifies _two_ EOTF functions: The standard "Annex 1" function, which is crap; and the more accurate "Appendix 1" function, which is used here.)

### Motivation

"Toeless sRGB" was always an unprincipled hack, motivated largely by the unsuitablity of every alternative I knew of at the time.

BT1886 Appendix 1 is correct in principle, because these videos were intended to be played back on a CRT television. (And "mid 90's Sony Trinitron" is a reasonable guess for which CRT television the original developers had on their desks.) In addition to being right in principle, it does everything "toeless sRGB" was meant to -- no banding near black around the stars in the FF7 opening movie, shadowy areas aren't totally lost to blackness, brights aren't blown out.

### ACKs

- [no] I did test my code on FF8 - I don't have FF8 installed, but the behavior should be consistent with FF7.


julianxhokaxhiu:
Can you please rebase this MR on top of master?

src/renderer.h
julianxhokaxhiu, line 1: If you did something on this file, please make sure you did not change indentation as it's not clear what happened. Thank you.

misc/FFNx.frag
julianxhokaxhiu, line 1: Same goes for this file.

ChthonVII:
I'm going to close this because I resubmitted it with the EOLs unchanged. Please see pull 816.

Issue #757, opened 2024-12-01
[ FF7 ] NTSC-J Gamut mode dithers the entire screen.


zaphod77:
**Describe the bug**
Enabling NTSC-J gamut applies a fullscreen quasirandom dither, when things would most likely look better without it.

**To Reproduce**

1. set  enable_ntscj_gamut_mode to true in ffnx.toml
2. install some mods with upscales.
3. Start game
4. Start a new game.
5. Skip the FMV to witness the problem.

**Expected behavior**
The game to look like it should on a Japanese NTSC tv with yellow boost, with no added dithering.

**Screenshots**
These are with syw unified upscaled field ninostyle chibi.

![image](https://github.com/user-attachments/assets/c88d3a1f-ef0a-4d90-ba35-aefb07c604f2)
![image](https://github.com/user-attachments/assets/bd33be26-a937-4d75-acaa-370e2bf31836)
The dithering is highly visible at lower DPI on a monitor.
**GPU (please complete the following information):**

- Nvidia GTX 1060 3GB
- Driver Version 561.03 (current)
But this isn't a driver issue, as i can see the code in the source that dithers the entire screen whenever a gamut conversion is done.
**Additional context**
Dithering is only needed to hide visible quantization noise, and shouldn't be needed for gamut conversions at 8 bit per channel of stuff with no banding.  I have no trouble changing gamut in photoshop 2 without introducing dithering or banding.

I suggest making dither_gamut_conversions an option, so the user can decide if it increases or decreases quality.

zaphod77:
i've done some testing and it is the dithering in the ffnx.post.frag that seems to cause this issue, and disabling it there seems to improve the image quality.

zaphod77:
a bit more testing reveals that the gamut conversion does introduce banding in the menu gradients, which would indeed need the dithering to fix.

The gamut converter thingie applies the same dither mask, whether it's needed or not.  While such dithering will indeed fix banding, it's still most likely wrong to apply it in post processing. it should be applied only to the original texture itself, and the gamut conversion should be done there, and nowhere else. This is particularly obvious when you use non upscaled assets, and you can see a pattern inside of the clearly visible pixels. I also believe it wrong to apply it to the gradient filled dialog and menu boxes, because those are generated in srgb.  the correct thing to do should be to gamut convert the default colors, and then not modify the generated gradient.

[...]

PR #816, opened 2025-07-15
Replace unprincipled "toeless sRGB" gamma function for PS1-derived movies with BT1886 Appendix 1 function with constants selected to match a properly calibrated mid-90s Sony Trinitron CRT.

ChthonVII:
## Summary

(This is a resubmit of pull 814 to fix the damn inconsistent-EOL issues (or, rather, to _not_ change the EOL marks in changed files despite the inconsistency) while keeping a clean commit history.)

Replaces the default gamma function for movies (that lack metadata specifying a particular gamma function). Removes the unprincipled "toeless sRGB" gamma function. Replaces it with BT1886 Appendix 1 function with constants selected to match a properly calibrated mid-90s Sony Trinitron CRT.

(Note that BT1886 specifies _two_ EOTF functions: The standard "Annex 1" function, which is rather useless; and the more accurate "Appendix 1" function, which is used here.)

### Motivation

"Toeless sRGB" was always an unprincipled hack, motivated largely by the unsuitablity of every alternative I knew of at the time.

BT1886 Appendix 1 is correct in principle, because these videos were intended to be played back on a CRT television. (And "mid 90's Sony Trinitron" is a reasonable guess for which CRT television the original developers had on their desks.) In addition to being right in principle, it does everything "toeless sRGB" was meant to -- no banding near black around the stars in the FF7 opening movie, shadowy areas aren't totally lost to blackness, brights aren't blown out.

### ACKs

- [ ] I did test my code on FF8 -  I don't have FF8 installed, but the behavior should be consistent with FF7.

### DIFF
--- a/misc/FFNx.common.sh
+++ b/misc/FFNx.common.sh
@@ -14,15 +14,16 @@
 /****************************************************************************/
 
 // Constants for BT1886 Appendix 1 EOTF Function
-// These constants correspond to a mid-90s Sony Trinitron CRT with the brightness turned pretty far up.
+// These constants correspond to a mid-90s Sony Trinitron CRT with the brightness turned pretty far up and the contrast turned pretty far down.
 // Do not change them blindly. If you change black or white level, then B, K, S, and I need to be recalculated.
+// And LUTs will need to be recalculated.
 // (https://github.com/ChthonVII/gamutthingy can calculate them for you.)
-#define crtBlackLevel 0.0015
-#define crtWhiteLevel 1.71
-#define crtConstantB 0.0927318537550299
-#define crtConstantK 1.3578779688794360
-#define crtConstantS 1.3852995465983342
-#define crtConstantI 0.1632400591934701
+#define crtBlackLevel 0.0018
+#define crtWhiteLevel 1.5
+#define crtConstantB 0.1042361441620798
+#define crtConstantK 1.1591247176412838
+#define crtConstantS 1.3711574400867077
+#define crtConstantI 0.1489575626482313
 
 // Gamut LUT
 SAMPLER2D(tex_10, 10);
@@ -110,6 +111,8 @@ vec3 toLinear2pt8(vec3 _rgb)
 
 // EOTF Function from BT1886 Appendix 1 (https://www.itu.int/dms_pubrec/itu-r/rec/bt/R-REC-BT.1886-0-201103-I!!PDF-E.pdf)
 // Approximates the gamma behavior of a CRT (more accurately than the crummy Annex 1 function)
+// Constants have been selected to match a mid-90s Sony Trinitron CRT with the brightness turned pretty far up and the contrast turned pretty far down.
+// Assumes input in range 0-1, output in range 0-1.
 vec3 toLinearBT1886Appx1Fast(vec3 _rgb)
 {
 	// add B


Changelog.md
```diff
@@ -4,6 +4,10 @@
 
 ## Common
 
+- Movie: Replace unprincipled "toeless sRGB" gamma function for PS1-derived movies with BT1886 Appendix 1 function with constants selected to match a properly calibrated mid-90s Sony Trinitron CRT.
+
+## Common
+
```
julianxhokaxhiu, line 10: Please compact this section and add the single new change line after the already existing one in the changelog ( top to bottom ). Thank you

ChthonVII:
Please don't merge this just yet. A properly calibrated Trinitron is too dark for FF7. I'm working on finding the best looking parameters.

ChthonVII:
OK, I think I'm happy with the parameters now.

---

I'd like to ask your thoughts on how to proceed with my upcoming contributions. I have two more things in mind. Both of them depend on the commits in this pull, and the second one depends on the first. Would you rather I space them out into this pull plus two more, or lump them altogether into one big pull? (I did one big pull last time, and you didn't see very happy about it.)

The next planned piece is to use BT1886 Appendix 1 gamma for _everything_ except movies that specify a gamma function in their metadata. (The rationale here is that all of the original art was created and designed to be played back on a CRT. Even the 1998 PC release would have been played on a CRT monitor. So we should use a gamma function that approximates a CRT.) I believe I have [this working](https://github.com/ChthonVII/FFNx/commit/985c295f967af45f00af2fdde546bac1f01cb144). Though I do need to ask for someone to test FF8 to make sure I didn't accidentally break it.

The final planned piece is to rework the NTSC-J mode. The current NTSC-J mode does a conversion from the NTSC-J specification to the sRGB specification. The problem with that is that real televisions didn't conform to the NTSC-J spec because industry chemists never figured out how to make the necessary phosphors. Instead they used "P22" phosphors that were pretty far out of spec in tandem with a "color correction circuit" to (very) crudely compensate for the discrepancy. So I can more closely match the original/intended experience by simulating the color correction circuit and then converting from the P22 gamut (or, rather, the subset of the P22 gamut that was reachable given valid inputs to the color correction circuit) to the sRGB spec. The central bit of this is porting [code I've already written for retroarch](https://github.com/ChthonVII/chthons_color_correction/tree/main). The tricky part will be correctly handling all the two-step-conversion possibilities with non-PS1-videos.

I guess I do have another question about color modes: Do we want more, or possibly fewer? Presently, we have two modes, uncorrected and NTSC-J. NTSC-J mode tries to reproduce the "intended experience" of what the original developers saw on the TVs on theirs desks, and the original target audience of Japanese consumers saw. Do we want to make more modes to reproduce what European or American audiences would have seen on their TVs? Or what people would have seen on late-90s CRT computer monitors? Alternatively, do we want to just get rid of uncorrected mode? I can think of three reasons for keeping uncorrected mode, and this planned revision is going to weaken them. (1) The current NTSC-J mode is very saturated, and some people don't like that. The revision will be less saturated. (2) Uncorrected mode is easier for modders, since what you see photoshop or whatever is what you get in FFNx. The revised NTSC-J mode will be less different from uncorrected, so the difference might be tolerable now in many cases. Also, gamutthingy now has a really good inverse mode, so you can convert and get exactly what you saw in photoshop in NTSC-J mode. (3) Previously I wrote some stubs for a potential future system that switched gamuts on a per-asset basis depending on metadata in the mod file. That would require (parts of) uncorrected mode to stick around. But no one seems to have interest in adding that metadata functionality to 7th Heaven. Overall, more modes means more complexity and worse performance, but also, hopefully, more users who have a mode they're happy with. My plan for now is to stick with the two existing modes. If this were my project, I'd probably eliminate uncorrected mode. If someone is really excited about adding other color modes, that's something I'd want to start designing around now.

julianxhokaxhiu:
Thanks for raising this concern. I want to be very honest on the questions you asked but let me try to answer them separately.

>  Would you rather I space them out into this pull plus two more, or lump them altogether into one big pull?

For the sake of the review, the smallest the code patch, the easier is to review it. Huge bloated PRs tend to usually sit there for weeks until I find the proper time to go through them. I personally prefer to have them packed as a single unit feature, hopefully involving less code patches as possible. If you could follow the same approach I'd appreciate.

> Do we want to make more modes to reproduce what European or American audiences would have seen on their TVs? Or what people would have seen on late-90s CRT computer monitors? Alternatively, do we want to just get rid of uncorrected mode?

So to be very frank, one of the reasons why I'm not able to introduce the hardware acceleration for 4K movies in FFNx is exactly because of this layer that requires massive rework to be included while using the ffmpeg HW acceleration layer ( which happenns all in the GPU ). Instead I'd like to ask you, would it be possible so to say to move all this code and logic into a separate Reshade shader for instance? Doing so would allow me to bring back sanity into the overall codebase, including the movie decoding. I've seen other games doing wonder with it on top, and FFNx supports that, would you be open in case to develop a separate solution based on ReShade? I'll be more than happy then to find a way to integrate it maybe as a custom mod in 7th.

Would you be open for it?

ChthonVII:
> [W]ould it be possible so to say to move all this code and logic into a separate Reshade shader?

No, I don't believe that would be possible. The sticking point is that the game engine draws other stuff on top of movies. This means that we have to deal with the movie's color properties right then and there, getting it into the same gamut/color space/gamma as the stuff that's about to get drawn on top of it. I don't not see any way around this.

Also, you may recall that we had to take responsibility for movie tv->pc range expansion away from ffmpeg and do it in FFNx's shader because ffmpeg had a bug causing it to apply the range expansion twice in some cases (including SYW's upscales). So far as I know, that bug still isn't fixed.

Another issue is that, for our most common use case for FF7, PS1-derived videos, we want a gamma function that ffmpeg just doesn't have. (Which I am now replacing with a better gamma function that ffmpeg doesn't have.)

Finally, I do not care for reshade one bit and would not be interested in working on in.

So, at any rate, I guess I will take this to mean that you definitely aren't interested in adding more color modes.

As far as simplifying the existing shader code, I can think of two things that would trim some code:
1. Remove the stubs for per-asset color gamut. No one seems interested in getting the metadata handling into 7th Heaven. So it's just dead code at this point.
2. Remove uncorrected mode. I guess we should see how NTSC-J looks after I'm done overhauling it.

julianxhokaxhiu:
Yep, I agree with both points then. If we can simplify the overall code I'd appreciate. Let's not overcomplicate it for the sake of it, if users are going to ask for more we can eventually expand, but I think the NTSC-J as it is might be the sweet spot everyone would be curious to play with.

> Also, you may recall that we had to take responsibility for movie tv->pc range expansion away from ffmpeg and do it in FFNx's shader

If that's the case then I'm fine with it, as with ffmpeg I can try to pass the GPU texture that already exists created by FFMpeg into bgfx without having to download it on the CPU and reupload it back to the GPU. By doing so we get what we all know as Hardware Acceleration, which means we decode the video data on the GPU and we keep it there. If the colors are fixed on the shader, they should apply as usual through the shader pipeline you already built.

Let me know if you want to handle all these changes into this PR or if you prefer to open a new one ( like if the code we're going to merge on this one is going to be deleted on the next one, there's no point in merging ).

Cheers :)

ChthonVII:
I apologize for not noticing your reply sooner.

I'm trying hard to keep each PR limited to stuff that (a) works on its own, and (b) is not going to change in the next PR.

However, I think I want to keep messing with the black level a bit. Now that I have the whole kit and kaboodle working (I think), I can consider more of the game when picking the black level. Also, I'd mucked up git when I was testing before, so I'm not sure the tests that led to the current choice were valid.

Hopefully I should have everything sorted in a few days.

julianxhokaxhiu:
Thanks a lot and don't worry, take your time. Appreciated the work you're doing on this nevertheless.

ChthonVII:
Rebased for the past few upstream commits.

The next two pull requests are ready to go, as soon as I have time to write out the "summary," "motivation," etc. If you want a sneak peek, they are [here](https://github.com/ChthonVII/FFNx/tree/CRTgammaforeveryone) and [here](https://github.com/ChthonVII/FFNx/tree/reviseNTSCJmode). I hope to write those up tomorrow. And then do one more pull removing the per-asset gamut override stubs. 

After that, it might be time to think about the possibility of removing uncorrected mode in the name of code simplification.

julianxhokaxhiu (changes requested):
CI checks do not pass. I'll be happy to review once the CI goes green. Thank you.

//EDIT: Re-running them just in case but I'd appreciate if you could ping me and let me know if the CI fails because of reasons outside of your own control.

ChthonVII:
> CI checks do not pass. I'll be happy to review once the CI goes green. Thank you.
> 
> //EDIT: Re-running them just in case but I'd appreciate if you could ping me and let me know if the CI fails because of reasons outside of your own control.

All the failed checks yesterday were due to vcpkg failing to download stuff. I did note this over on pull #825.

That said, I'm considering messing with the gamma constants some more. So please hold off from merging for a bit.

julianxhokaxhiu:
Ok, CI seems to pass now, it was just Github runners taking too long to build. Vcpkg is just fine :) I'll hold on then, lmk when ready for review. Thank you

ChthonVII:
I think all my pull requests are ready to go at this point. However, I need to make significant revisions to the pull requests' text/notes. Since it's nearing 1AM here, I'm going to put that off until tomorrow.

ChthonVII:
I think I'm just going to do one big post here, and then link to it in the other pull requests. I think everything's ready to go at this point. I have a few major thoughts I want to throw out here:

1. Starting with the embarrassing one: I've learned the hard way that when Visual Studio/cmake copies files into `.build/bin/` it does ***NOT*** overwrite existing files. I screwed myself up so many times when I tested with the wrong binary or wrong LUT files and ended up thinking something wasn't working when in fact it was. In two cases I reverted something and then put it back again; this is why. (This is probably also how I botched the diagnosis of #818. Twice.)
2. I wrote up [some pretty substantial documentation](https://github.com/ChthonVII/FFNx/blob/reviseNTSCJmode/docs/gamma_and_gamut.md) for the gamma and gamut stuff. I hope this proves worth reading and gives a useful explanation of what this code is doing, and what's baked into those mystery LUTs, and why.
3. I'm trying to break up the pull requests into digestible chunks at points where FFNx should be in a correct working state after each merge. But I think maybe I've hit the limit of how far I can run ahead of stuff getting merged while keeping things straight and orderly.
4. It turns out that I was able to prevent BGFX from doing sRGB linearization in the first place. (This is one of those instances where my changes worked the first time, but I tested with the wrong binary.) It looks to me like it's working with no unintended side effects. But this is a pretty big change, so I'd appreciate a second opinion.
5. Likewise, I can't test HDR or FF8, but both really need tested. (In particular the revised NTSC-J mode needs tested in HDR.)
6. Just FYI: I wanted to do 128x128x128 LUTs, but BGFX crapped out on them. So we're back to 64x64x64. I did try to stretch their bandwidth by storing values in gamma space.
7. What's next? Three things come to mind:
     - Maybe remove uncorrected mode and make everything NTSC-J mode all the time? 
          - The main reason to consider this is code simplification. And possibly slight performance gain.
          - The main reason not to is that modding is easier in uncorrected mode, where what you see in Photoshop, etc. is what you get in FFNx. But...
               - I already broke "what you see is what you get" a little bit when I changed the gamma.
               - The revised NTSC-J mode is closer to uncorrected mode. So usually stuff created in sRGB looks "good enough."
               - It's easier now to make mods for NTSC-J mode using sRGB tools like Photoshop, etc. There is now a Windows binary for gamutthingy, and [instructions](https://github.com/ChthonVII/FFNx/blob/reviseNTSCJmode/docs/mods/preparing_textures_for_ntscj_mode.md) for what to do with it to make mods for NTSC-J mode.
          - I'm pretty biased here. So I'd like to know what other people think after trying out the revised NTSC-J mode. Would they ever go back to uncorrected mode?
     -  The issue in that "substantial documentation" captioned "A Minor Shortcoming" bothers me way more than it should. I'd like to fix it, and I think I see how. But... it would be an unholy mess of conditionals, unless uncorrected mode was removed first.
     - [zaphod77](https://github.com/zaphod77) wanted a feature to disable dithering. (#757) So... On the one hand, the dithering is *correct*, and zaphod's problem is due to viewing a low-DPI screen closer than its minimum recommended viewing distance, and adding a feature means complicating the code. On the other hand, it should be fairly easy to implement, and would make the user happy. Thoughts?

ChthonVII:
Sorry to hold up things up yet again, but I believe I've dreamed up a refactor that could simplify some things while making lighting more gamma-correct in more cases. I'll try to implement it ASAP.

julianxhokaxhiu:
Take your time, to be honest the more I'm looking at those 4 PRs, the more I think it's time for you to look at ReShade and reimplement your shaders over there. I've been working during this time to integrate it on 7th/J8 so you can also release it as a mod in the catalog in case ( or just add it to the official ReShade exe for manual installation ).

I'm actually looking at bringing FFNx code base into a simpler manageable state. And unfortunately NTSC-J is doing the opposite and this maintenance falls on me, which I'm not keen to do. At the same time, transforming your mod into a ReShade shader would allow you to update it anytime with your own release schedule.

i hope you do understand. Thanks nevertheless for working on it, I hope to see this soon released for ReShade.

ChthonVII:
> Take your time, to be honest the more I'm looking at those 4 PRs, the more I think it's time for you to look at ReShade and reimplement your shaders over there.

Unfortunately, that is not possible. Post-processing like reshade is simply too late.

Let me try to illustrate why. Consider the behavior of an actual PS1:
- T1: The PS1 CPU/GPU generate a frame using logic entirely confined to gamma-space R'G'B'.
- T2: The PS1 sends the frame to its DAC/modulator unit which converts it to a composite signal and sends it down the cable.
- T3: The television's demodulator/color correction unit converts the composite signal back to gamma-space R'G'B', applying a color correction matrix in the process.
- T4: The television uses the corrected R'G'B' signals to control the 3 electron guns. The electron guns have non-linear relationship of input voltage to output electron beam intensity. This is CRT gamma. The output electron beam intensity is linear RGB. The gamma function implicit in the electron gun physics is different from the gamma function of modern displays.
- T5: The electron beam excites the phosphors and they glow. The color of the light emitted is different from the color of light emitted by a modern display given the same input.

In the case of a "pure" emulator, like a retroarch core, the emulator simulates T1 (and maybe T2) only. In that scenario, it's possible to hook up some form of post-processing to the emulator output to simulate T3, T4, and T5 correcting for the differences between a CRT and modern displays. There is a neat and clean division.

However, FFNx is **_NOT_** a pure emulator. It does two things that make it impossible to neatly divide matters into "T1&T2" versus "T3&T4&T5" in a way that allows "T3&T4&T5" to be shoved into post-processing:
1. Movies are not guaranteed to be in the same color space as the things that are going to be drawn on top of them. Even worse, in the case of Steam FF8, it looks like they've already undergone some conversion, so they're guaranteed **_NOT_** to be in the same color space. The movies need their colorimetry corrected *immediately*, before things get drawn over the top of them. There's simply no way to go back and fix mixed color spaces in post-processing. (This is not even really a "NTSC-J issue" so much as an overall "colorspace correctness issue." Even when NTSC-J mode is off, it's still a problem.)
2. FFNx does things -- mostly lighting -- in linear RGB. This comes at T4 or T5. To get lighting fully correct, color correction simulation at T3 needs to be handled *before* lighting.* Again, there's no way to go back and fix this in post-processing. In fact, my current refactor is aimed at fixing lighting by moving colorimetry in front of it.

(* It's even a bit worse than that: The output from color correction simulation at T3 cannot be usefully normalized to 0-1 as the lighting code requires, except via the gamut compression that's done at T5. So "T3&T4&T5" before lighting is the only (correct) implementation that's possible. Fortunately, the colorimetry corrections should be distributive over the lighting operations, so the result should be correct. (Except that gamut compression (pulling colors outside sRGB inside without clipping) is only arguably distributive over anything. However, gamut compression is inherently an approximation, so a little (arguable) inaccuracy here is acceptable. And this is the best implementation that's possible, so we just have to live with it anyway.)

ChthonVII:
My refactor efforts are running into a problem with eye textures and battle background floors. If both advanced lighting and NTSC-J mode are enabled, these textures turn black. The problem goes away if I stop color correcting textures. Is there anything special about these textures? Like are they maps or something?

ChthonVII:
#832 incorporates these changes and supersedes my other pull requests. I'm leaving this one outstanding because it could stand on its own.

julianxhokaxhiu:
Just to clarify, does #832 superseed this one? If yes, please close this one and rebase the other one so we can get it merged it cleanly. Thank you!

ChthonVII:
> Just to clarify, does #832 superseed this one? If yes, please close this one and rebase the other one so we can get it merged it cleanly. Thank you!

#832 includes this one (#816). This one (#816) exists as an an independent pull request because you **_might_** hypothetically want to just fix the movie gamma independently before undertaking review of the whole NTSCJ overhaul. If you're not inclined to do that, you certainly can just close this one (#816) and merge #832.

I'll try to find time to rebase #832 tomorrow.

julianxhokaxhiu:
Thanks I looked into #832 and I didn't saw big changes but I'd rather tackle it all at once rather than splitting it. So yeah, let's work over there and close this one instead. Cheers!

Issue #841, opened 2025-09-24
[ FF8 ] Swizzle as one image using wrong stride.


Sebanisu:
**Describe the bug**
https://github.com/julianxhokaxhiu/FFNx/blob/a07fcc06213595a44bd83b39ff3d3217159610ec/src/ff8/mod.cpp#L601C1-L606C6
`imgWidth / imgScale` should be `imgWidth` for when Swizzle is just one image. I haven't yet tested this with textures using Tonberry's format it might break those. I'm gonna prepare some test images for Tonberry method and will let you know if it still works okay.

Problem is the stride is wrong we copy pixels from the wrong parts of the image.

**To Reproduce**
[testpattern.py](https://github.com/user-attachments/files/22521069/testpattern.py)

1. I placed this gpt generated python script in D:\SteamLibrary\steamapps\common\FINAL FANTASY VIII\mods\Textures\field\mapdata\bchtr_1
2. I used this to generate a test image. `python testpattern.py 608 256 4 bchtr_1.png` `python testpattern.py [width] [height] [scale] [filename]`
3. I ran the game and loaded my save which is in `bchtr_1`. The hotel room in Balamb city https://finalfantasy.fandom.com/wiki/Hotel_(Final_Fantasy_VIII)
4. Current canary shows a distorted image.

**Expected behavior**
The 4x image should be clear and look like 1x but better.

**Screenshots**
1x with correct stride
<img width="642" height="512" alt="Image" src="https://github.com/user-attachments/assets/4bf07174-98f2-48b5-bbd7-628508f8be51" />

4x with incorrect stride
<img width="642" height="512" alt="Image" src="https://github.com/user-attachments/assets/eca2413a-0cd0-47fa-a035-08ec7398eac5" />

4x with correct stride
<img width="642" height="512" alt="Image" src="https://github.com/user-attachments/assets/a4ae53a8-20a0-4d1b-b760-ca0cbf2ab3a9" />

**GPU (please complete the following information):**
```log
[00000000] INFO: --- PC SPECS ---
[00000000] INFO:    CPU: Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz
[00000000] INFO:    GPU: NVIDIA GeForce GTX 980 Ti (4095MB) - Driver: 32.0.15.6636 - Backend: Direct3D11
[...]

PR #842, opened 2025-09-24
[ FF8 ] Fix stride when using Swizzle as one image.

Sebanisu:
fixes https://github.com/julianxhokaxhiu/FFNx/issues/841

## Summary

Fix stride when using Swizzle as one image.

### Motivation

Remove ` / ImgScale ` as it was causing the stride to be messed up when going beyond 1x scaling.

### ACKs

- [x] I did test my code on FF8

I tested Tonberry style in Balamb Hotel I think it's still working.
I also converted Angelwing to the new format and ran around Balamb city. I didn't see any issues.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -25,7 +25,7 @@
 - Widescreen: Initial support for the entire game
 - Widescreen: Fix menu rendering bringing it back to its original aspect ratio
 - Widescreen: Enable some full 3D areas to be seen entirely instead of vertically cropped
-- Swizzle as one image: Fix stride, was `imgWidth / imgScale` instead of `imgWidth`. (https://github.com/julianxhokaxhiu/FFNx/pull/842)
+- External textures: Swizzle as one image: Fix stride, was `imgWidth / imgScale` instead of `imgWidth`. (https://github.com/julianxhokaxhiu/FFNx/pull/842)
 
 ## FF8 Steam
 


src/ff8/mod.cpp
```diff
@@ -599,7 +599,7 @@ TexturePacker::TextureTypes TextureBackground::drawToImage(
 			const int col = tileId % _colsCount, row = tileId / _colsCount;
 
 			drawImage(
-				imgData, imgWidth / imgScale, imgScale,
+				imgData, imgWidth, imgScale,
```
julianxhokaxhiu, line 602: @myst6re can you please let me know if this might cause any side effect?

julianxhokaxhiu: So does this mean it's safe to merge? Sounds like a yes, but I'd like to be aligned with you.

myst6re: It is safe, I checked further and the issue is that imgWidth is already divided by imgScale above

Thanks @Sebanisu  for spotting it!

julianxhokaxhiu: Thanks a lot to both :) Allright, I'll wait for @Sebanisu to mark it as ready and then I'll be happy to review and merge. Cheers!

Sebanisu: I tested in the Balamb Hotel with tonberry style and it looks like it's working fine.

julianxhokaxhiu:
Thanks a lot!

PR #852, opened 2025-11-17
FF8: More UV fixes for 3D models (battle, field, worldmap)

myst6re:
## Summary

UV adjustments.  In ff8 data, UVs are (x, y) coordinates with x and y = range 0 -> 255
FF8 PC uses a mapping between this and floating values from 0.0 to 1.0.
For some reason, this mapping is built like this:

```
mapping[n] = n / 256.0f
mapping[0] = mapping[1];
mapping[255] = mapping[254];
```

[0] and [255] are altered to fix some potential issues we seems to don't have in FFNx.
Worst, the division by 256 seems wrong, because it results to values between 0.0f and 0.992188f (255.0f / 256.0f), and not between 0 to 1.

By changing this mapping to:

```
mapping[n] = n / 255.0f
```

Worldmap seems now completely fixed, battle backgrounds repeat seems more consistent, 3D models looks better I think (it is hard to tell), etc...

Seifer uv change (left original game, right FFNx)<br>
Seifer uv change (left original game, right FFNx with this PR)<br>

original seifer texture high res, proof that the red symbol is thin<br>
original seifer texture high res, proof that the red symbol should be thin

uv-worldmap<br>
The difference is really subtle in worldmap, because most of the fixes was already done

uv-worldmap2<br>
Some glitches are gone inside the tunnel. Also, the textures moved a bit.

uv-battle<br>
It is again hard to see, but image on the bottom has a wider mountain texture and sky

### ACKs

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -7,6 +7,7 @@
 - Core: Fix crashes happening in Non-US versions ( https://github.com/julianxhokaxhiu/FFNx/pull/848 )
 - External textures: Fix glitches in field module ( https://github.com/julianxhokaxhiu/FFNx/pull/848 https://github.com/julianxhokaxhiu/FFNx/pull/851 )
 - External textures: Fix Tonberry format when dumping PNGs using `save_textures_legacy` flag ( https://github.com/julianxhokaxhiu/FFNx/pull/848 )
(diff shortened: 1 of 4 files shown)


julianxhokaxhiu (approved):

julianxhokaxhiu:
Super cool, I love this patch. Thanks a lot!

PR #854, opened 2025-11-21
FF8: FFMPEG bink

myst6re:
## Summary

- Use ffmpeg to read bink movies (from ff8-2000)
- Enable external_movies by default for all versions

### Motivation

- Modders can mod the same way on both ff8-2000 and ff8 steam
- we don't have to maintain a special case for ff8-2000

### ACKs


julianxhokaxhiu:
Thanks a lot, masterpiece PR!

PR #857, opened 2025-11-30
FF8: Minimize field background width to prevent overrides

myst6re:
## Summary

bgroad_6 cannot be modded (using the FFNx method) because the mod is overridden by another texture upload on the right of the texture.

Background textures (mim files) are always 3328 bytes wide, but not everything is used by the tiles (map files)
The solution is to check in the tiles what is really used and limit the background texture width.

### ACKs

### DIFF
--- a/src/renderer.cpp
+++ b/src/renderer.cpp
@@ -1726,7 +1726,7 @@ void Renderer::setScissor(uint16_t x, uint16_t y, uint16_t width, uint16_t heigh
     scissorHeight = getInternalCoordY(height);
 
     // This removes the black bars on the top and bottom of the screen
-    if (enable_uncrop)
+    if (enable_uncrop && !ff8)
     {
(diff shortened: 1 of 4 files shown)


src/renderer.cpp
```diff
@@ -1726,7 +1726,7 @@ void Renderer::setScissor(uint16_t x, uint16_t y, uint16_t width, uint16_t heigh
     scissorHeight = getInternalCoordY(height);
 
     // This removes the black bars on the top and bottom of the screen
-    if (enable_uncrop)
+    if (enable_uncrop && !ff8)
```
julianxhokaxhiu: Out of curiosity, is this a left over or done on purpose?

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #860, opened 2025-12-07
FF8: Allow any field map to contains up to 16 visible models

myst6re:
## Summary

- Prevent crash when the sum of mch (3D model) file size is high.
- Reorganize dynamically 3D model textures to prevent overrides between 3D models and background and particles effects

### Motivation

Modders are struggle with the game limitations:
- Create a 3D model with more vertices uses more memory and can crash the game
  - I added 16 MB of memory, the same way the Remastered did.
- When you replace every party members models by the high res version (the game already has several versions for the same model), on some maps the game can override the background texture in the emulated VRAM
  - The PC game does not use all the emulated VRAM, because menus and the screen buffers are outside the emulated VRAM. So I added new places for 3D models there, and created a dynamic algorithm to avoid overrides with background or particles effect textures that still are in VRAM

Please note that despite my efforts, the game is still limited to 16 3D models displayed on screen max, like FF7.

This place is crowded
Even with lot of high res models, it works
2025-12-02 22_45_28-SSIGPU - VRAM Display

- In green: 3D model textures
- In purple: main locations of particles textures
- In blue: background textures
- In turquoise: 3D model textures in the area of background textures (works only on fields with a simple background)
- In black: most of locations are unused and can be allocated for 3D models

### ACKs

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -5,6 +5,7 @@
 ## FF8
 
 - Core: Fix crashes happening in Non-US versions ( https://github.com/julianxhokaxhiu/FFNx/pull/848 )
+- Core: Fix the game engine to show up to 16 high res field models without crashes or texture glitches ( https://github.com/julianxhokaxhiu/FFNx/pull/860 )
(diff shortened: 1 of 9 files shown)


julianxhokaxhiu:
Thanks a lot! This is MASSIVE 🫶

PR #887, opened 2026-02-07
FF8: Add initial Remastered compatibility

myst6re:
## Summary

Bring Final Fantasy VIII Remastered compatibility with FFNx

### Motivation

Because I can

### How to test

Read [how_to_install.md](https://github.com/myst6re/FFNx/blob/8bf0935c23a4ee754edfe7a4c21e7577c7a43d61/docs/how_to_install.md#early-alpha-remastered-release)
Battle will crash, please wait for the remastered mod on junction VIII to fix battle. (https://github.com/tsunamods-codes/Junction-VIII-Catalogs/pull/111)

### Already in master
- [X] Add FS archive support for LZ4 compression
- [X] Add more memory for field models

### Current PR
- [X] Remastered detection (if FFVIII.exe exists, then remastered flag is enabled)
- [X] Load data from main.zzz archive
- [X] Read sounds from other.zzz archive
- [X] Read movies from other.zzz using ffmpeg
- [X] Read music from other.zzz using VGMSTREAM
- [X] Read wav from other.zzz using VGMSTREAM
- [X] Read external textures from main.zzz using pnglib
- [X] Fix FS archive lang paths for field
- [X] Link game to Steam and redirect created files to Remaster paths
- [X] Make steam dll linking external and optionnal
- [X] Ensure the official launcher can start the game with FFNx
- [X] Reduce scale size of remaster models
- [X] Add more memory for battle effects
- [X] Add minimal support for remaster model textures field
- [X] Add full support for remaster textures field
- [X] Add support for remaster textures world (textures/field.fs/field_hd_new/wmset_014_0.png)
- [X] Add minimal support for remaster model textures battle
- [X] Update installation steps
- [X] Add support for high res menu font and icons
- [X] Add support for high res triple triad
- [X] Update ff8_high_res_font parameter
- [X] Add a new parameter to enable/disable remastered textures

### Part 2
- [X] Fix hang for battle models with more vertices (Main characters + Weapons + G-Forces + Some monsters)
- [X] Add native multilingual support
- [ ] Add support for new texture animation system in field
- [ ] Add support for new texture animation system in battle
- [ ] Add full support for remaster textures battle
- [ ] Add chocobo world item drop to Angelo limit break
- [ ] dinput, xinput, keyboard icons auto selection support (hires\iconfl01.tex\en\4_xinput.png)
- [ ] Fix field light on field textures

### Part 3
- [ ] Add ability to configure the field model scale value

2026-02-20 13_52_49-Final Fantasy VIII (Direct3D11) devel
2026-02-20 13_53_08-Final Fantasy VIII (Direct3D11) devel
2026-02-20 13_53_24-Final Fantasy VIII (Direct3D11) devel
2026-02-20 13_54_44-Final Fantasy VIII (Direct3D11) devel
2026-02-20 13_55_20-Final Fantasy VIII (Direct3D11) devel
2026-02-20 13_57_17-Final Fantasy VIII (Direct3D11) devel

### ACKs

- [X] I did test my code on FF8 Remastered
- [ ] I did test my code on FF8 Steam


julianxhokaxhiu:
What is the reasoning behind changing the entire Steam API dll linking out of curiosity?

myst6re:
> What is the reasoning behind changing the entire Steam API dll linking out of curiosity?

FFNx overrides the steam api dll of the game. On 2013 version it works because we have basically the same dll, but with the remastered, the launcher and FFVIII.exe links to a ~~64-bit steam dll~~ untrue, a more recent steam dll which is not compatible with ours. Using our dll prevent the launcher from starting, that's all :)

julianxhokaxhiu:
Appreciated, let's discuss this internally then on how we could tackle it better, but it's not a blocker for now so feel free to keep working with this approach, there are other major things to fix before anyway and this is a minor ( but nevertheless important ) nuance. Thanks!

myst6re:
Added support for menu texture + some screenshots

zaphod77:
If you do get the auto icon thing working, please add it to ff7 as well?  The keyboard and contorller support in this version ,as well as ff7 from windows store is very good.

UnblessedSouls:
@myst6re for the hangs on the remastered version I'm not sure if this helps but I was learning how to research stuff in the remastered version and came across this information if its already known then feel free to ignore 

FF8 Remastered: Found Hardcoded 12-bit Vertex Mask in FFVIII_EFIGS.dll

Description
While investigating the "battle hang" issue when loading high-poly models in the Remastered version, I've isolated the specific hardware mask responsible for the vertex limit. The logic is confirmed to live inside the legacy engine wrapper (FFVIII_EFIGS.dll).

Technical Breakdown
The engine uses a bitwise AND to truncate the vertex count at 4,095 (0xFFF) before memory allocation. Attempting to patch this mask to 0xFFFF via debugger results in a "Vertex Explosion," suggesting that the subsequent rendering pipeline expects a specific 12-bit data structure or is reading adjacent UV/Texture data as geometry.

Location Data

    Target Module: FFVIII_EFIGS.dll

    The Mask Instruction: AND EBX, 00000FFF

    AoB Signature: 8B D8 C1 E8 0C 81 E3 FF 0F 00 00
    (Note: This translates to mov ebx,eax, shr eax,0C, and ebx,0x0FFF)

Call Stack for Hooking:

    Read Vertex Count: FFVIII_EFIGS.dll + 159B273 (movzx eax,ax from .dat buffer)

    Part Pointer Loop: FFVIII_EFIGS.dll + 81171B

    Main Builder Function: FFVIII_EFIGS.dll + 7F967A

Observations
Manual modification of the mask in the active process confirmed it is the master gatekeeper for all battle models (Characters, Weapons, GFs). A proper fix will likely require a detour hook to re-allocate or manage the buffer offset to prevent the vertex explosion seen here:.
Screenshot 2026-02-24 204218

myst6re:
> @myst6re for the hangs on the remastered version I'm not sure if this helps but I was learning how to research stuff in the remastered version and came across this information if its already known then feel free to ignore

Firstly, thanks for the investigation! I needed help, and here you are! The sub you talk about is a remastered hook that's rewrite the original sub entirely, maybe just to add the mask, or maybe for more. I guess I'll eventually rewrite the entire sub too. I done the same for field models, so that's not surprising.

myst6re:
@julianxhokaxhiu I'm ready for the review, battle still needs [this mod](https://github.com/tsunamods-codes/Junction-VIII-Catalogs/pull/111) to work.

docs/how_to_install.md
```diff
@@ -84,7 +84,7 @@ external_music_ext = "akb"
 #### [2000 Squaresoft Release](https://www.mobygames.com/game/windows/final-fantasy-viii)
 
 0. Install the game on this path: `C:\Games\Final Fantasy VIII`
-1. Update your game to v1.2 ( you can find a collection here: https://www.ff8.fr/telechargements/programmes )
+1. Update your game to v1.2 ( you can find a collection here: [FF8.fr download page](https://www-ff8-fr.translate.goog/telechargements/programmes?_x_tr_sl=fr&_x_tr_tl=en&_x_tr_hl=fr&_x_tr_pto=wapp#archives_conserve_pour_la_posterite) )
```
julianxhokaxhiu: Can you please leave the original link? I'd rather let users decide which translator to use instead of enforcing them through Google :) Thanks

docs/how_to_install.md
```diff
@@ -100,6 +100,21 @@ external_music_ext = "akb"
...
 6. Enjoy!
 
+#### Early alpha [Remastered Release](https://store.steampowered.com/app/1026680/FINAL_FANTASY_VIII__REMASTERED/)
+
+**Warning:** The compatibilty with the Remastered version is currently uncomplete. You will most likely encounters crashes
+in battle and model rendering issues.
+
+0. Install the game using Steam
+1. Open the installation directory of the game ( see [How to access game files](https://steamcommunity.com/sharedfiles/filedetails/?id=760447682) )
+2. Download the FF8 patch v1.2 ( you can find a collection here: [FF8.fr download page](https://www-ff8-fr.translate.goog/telechargements/programmes?_x_tr_sl=fr&_x_tr_tl=en&_x_tr_hl=fr&_x_tr_pto=wapp#archives_conserve_pour_la_posterite) )
```
julianxhokaxhiu: Same as above

docs/how_to_install.md
```diff
@@ -100,6 +100,21 @@ external_music_ext = "akb"
 5. Replace all files when asked.
 6. Enjoy!
 
+#### Early alpha [Remastered Release](https://store.steampowered.com/app/1026680/FINAL_FANTASY_VIII__REMASTERED/)
+
+**Warning:** The compatibilty with the Remastered version is currently uncomplete. You will most likely encounters crashes
+in battle and model rendering issues.
```
julianxhokaxhiu: Feel free to put this sentence in a single line, no need to truncate at 80 char width

Changelog.md
```diff
@@ -14,6 +14,8 @@
 - Core: Add `ff7_multibyte_font` mode: multibyte text support for non-Japanese translations on the English executable ( https://github.com/julianxhokaxhiu/FFNx/pull/948 )
 
 ## FF8
+
+- Core: Add support for remastered version, see manual (https://github.com/julianxhokaxhiu/FFNx/blob/master/docs/how_to_install.md#early-alpha-remastered-release) for more informations ( https://github.com/julianxhokaxhiu/FFNx/pull/887 )
```
julianxhokaxhiu: Feel free to add the manual link as [see manual](#)

docs/how_to_install.md
```diff
@@ -100,6 +100,20 @@ external_music_ext = "akb"
 5. Replace all files when asked.
 6. Enjoy!
 
+#### Early alpha [Remastered Release](https://store.steampowered.com/app/1026680/FINAL_FANTASY_VIII__REMASTERED/)
```
julianxhokaxhiu:
```suggestion
#### [Remastered Release](https://store.steampowered.com/app/1026680/FINAL_FANTASY_VIII__REMASTERED/)
```

I know it's early alpha but the warning below already highlights that, as we did back than for JP on FF7 :) So fine with me to remove that nomenclature. Feel free to update the link as well to this section. Thank you

src/steam.cpp
```diff
@@ -0,0 +1,217 @@
...
+	char lib_path[MAX_PATH] = {};
+
+	strncpy(lib_path, "ffnx_steam_api.dll", sizeof(lib_path));
+
+	if (!fileExists(lib_path)) {
+		strncpy(lib_path, "steam_api.dll", sizeof(lib_path));
+	}
+
+	bool is_genuine_steam_api = isFileSigned(lib_path);
+	if (!is_genuine_steam_api) is_genuine_steam_api = sha1_file(lib_path) == "03bd9f3e352553a0af41f5fe006f6249a168c243";
```
julianxhokaxhiu, line 35: This condition will likely match the version we use now as dependency, but let's say we bump it and we load the stock `steam_api.dll` the hash might be different. We shall probably consider using a different hash based on which dll we match. What do you think?

//EDIT: Now that I think it again, we need to add the game edition dimension into the picture. 2013 comes with a specific version, Remastered might come with another, GOG with another ( if you use GOG Galaxy for eg. they ship their own shim ) and so on...not sure how to manage this case now, before it was simplified as we always shipped with the lower common denominator but now this becomes exponentially infinite combinations.

myst6re: Even without my changes, FFNx relies on the steam api DLL provided by us. It means that even if the game uses a different version (I assume it is already the case, FF7 and FF8 Steam Editions does not share the same version I think), FFNx provides its own version for its own usage.

What I really changed, is that we don't override steam_api.dll anymore, so if the game or specifically the launcher does not support our steam_api.dll, it is OK. FFNx will now use ffnx_steam_api.dll and not the original game's steam_api.dll, allowing the launcher to still work

julianxhokaxhiu: Yeah I get  it but the validation routine we do pick either ours or the default dll. IF we allow that then we need to consider the validation for all cases, so include all sha1 for every edition. IF NOT then we can drop loading the default one and just error saying we couldn't find the dll we were looking for and exit. Either case are fine by me, but if I have to chose by complexity, I'd go for the latter option.

But I'll let you decide :)

myst6re: >  IF NOT then we can drop loading the default one and just error saying we couldn't find the dll we were looking for and exit.

Actually FFNx does not load steam_api.dll anymore with this PR, only ffnx_steam_api.dll. I unlinked steam_api.dll in the CMake file

Edit: ah you mean don't try to fallback to steam_api.dll if ffnx_steam_api.dll is not found? Yes I can change that

myst6re: Done

src/steam.cpp
```diff
@@ -0,0 +1,213 @@
...
+typedef ISteamUtils*(*LPUTILS)();
+typedef ISteamUserStats*(*LPUSERSTATS)();
+
+bool load_library()
+{
+	if (hDll != nullptr) {
+		return true;
+	}
+
+	char lib_path[MAX_PATH] = {};
```
julianxhokaxhiu: Can we `#define` this string instead of allocating a buffer for it?

myst6re: Sure :)

.github/workflows/build.ps1
```diff
@@ -84,23 +84,38 @@ cmake --build --preset "${env:_RELEASE_CONFIGURATION}"
...
 mkdir .dist\pkg\FFNx_Steam | Out-Null
 Copy-Item -R "$releasePath\bin\*" .dist\pkg\FF7_1998
 Copy-Item -R "$releasePath\bin\*" .dist\pkg\FF8_2000
+Copy-Item -R "$releasePath\bin\*" .dist\pkg\FF8_Remastered
 Copy-Item -R "$releasePath\bin\*" .dist\pkg\FFNx_Steam
 Remove-Item .dist\pkg\FF7_1998\FF8.reg
+Remove-Item .dist\pkg\FF7_1998\ffnx_steam_api.dll
 Remove-Item .dist\pkg\FF8_2000\FF7.reg
+Remove-Item .dist\pkg\FF8_2000\ffnx_steam_api.dll
+Remove-Item .dist\pkg\FF8_Remastered\FF7.reg
```
julianxhokaxhiu: Is it possible to compact all the `Remove-Item .dist\pkg\FF8_Remastered` in a single block? So it's easier to visualize it even later when we'll edit the file :)

myst6re: I followed the existing pattern, but yes I can

CMakeLists.txt
```diff
@@ -309,10 +332,21 @@ add_custom_command(
...
+          ${CMAKE_BINARY_DIR}/bin/${REMASTERED_RELEASE_NAME}.dll
+          ${CMAKE_BINARY_DIR}/bin/FFVIII_EFIGS.dll
+  COMMAND ${CMAKE_COMMAND} -E copy
+          ${CMAKE_BINARY_DIR}/bin/FFVIII_EFIGS.dll
+          ${CMAKE_BINARY_DIR}/bin/FFVIII_JP.dll
   # steam_api.dll
   COMMAND ${CMAKE_COMMAND} -E copy
           ${CMAKE_BINARY_DIR}/vcpkg_installed/x86-windows-static/tools/steamworkssdk/steam_api.dll
-          ${CMAKE_BINARY_DIR}/bin
+          ${CMAKE_BINARY_DIR}/bin/ffnx_steam_api.dll
```
julianxhokaxhiu: Stupid thing but...can we name the dll as `FFNx_steam_api.dll`? Ideally update also any ref using this name in source code ( I think should be only the define but I'm not sure ). Thanks

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thanks a lot and well done :)

PR #890, opened 2026-02-14
FF7: Fix 60FPS swirl fading effect

tangtang95:
## Summary

Fix 60FPS swirl fading effect: the swirl duration is less than original because there is no easy way to patch it as it will overflow a char value, so the timing has been revised for 60fps. For this reason, the fading effect will start earlier

### Motivation

Related to this issue: https://github.com/tangtang95/ff7-60fps-mod/issues/5

### ACKs

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -12,6 +12,7 @@
 
 - Core: Fix KAWAI opcode behavior ( https://github.com/julianxhokaxhiu/FFNx/issues/809 + https://github.com/julianxhokaxhiu/FFNx/issues/874 )
 - Graphics: Implemented smooth skinning for gltf 3d models ( https://github.com/julianxhokaxhiu/FFNx/pull/882 )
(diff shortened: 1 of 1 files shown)


julianxhokaxhiu:
Would you mind updating also the Changelog please? 🙏

tangtang95:
> Would you mind updating also the Changelog please? 🙏

Done!

tangtang95:
I think I have the correct solution. I need to change the next instruction that is used for the jump

tangtang95:
Nevermind, failed attempt :(

julianxhokaxhiu:
Take it easy, we can always improve it through time :) Thanks!

tangtang95:
There is always room for improvement: we can rewrite the entire swirl function 😂, but I think now the swirl effect is game changing (it needs a new run of the game 😂)

julianxhokaxhiu:
Who am I to say no to that :P Be my guest :D

Issue #654, opened 2024-02-11
[ FF7 ] Ground beneath Ifrit: color inverted


KuraudoXS:
**BUG DESCRIPTION**
With vanilla PC version the Ifrit Summon effect has an issue. The hole that appears beneath Ifrit when you summon him has the color inverted on PC. 

**TO REPRODUCE**
Summon Ifrit.

**EXPECTED BEHAVIOR**
With vanilla PC version the color will appear fully white instead of black. See DF Retro at 30:11: https://youtu.be/-EOROam1GDs?si=w9BoxpBtedKcyHKJ&t=1811

**SCREENSHOT**
![DF Retro_ Final Fantasy 7 - A JPRG Epic Analysed Across The Generations! - YouTube - Brave 2_11_2024 8_08_49 AM](https://github.com/julianxhokaxhiu/FFNx/assets/9687549/dfd8fec5-718d-491b-abf7-eaf7a98fdebb)

KuraudoXS:
If that's fine, I'll just add the other points from the same DF Retro FF7 Analysis video for context, which are all issues found in the battle module (apart from one, which is a field module one.) https://youtu.be/-EOROam1GDs?si=25rRN2YghNX_TF95&t=1737 

30:08 Battle "Ifrit Summon problem: the color is inverted" 
29:19 Battle "It comes down how the game is being rendered using 3d acceleration, the mix of hi-resolution output with bilinear texture filtering causes problems with certain visual effects"
28:57 Field "The background artwork retains the same low 320 by 240 pixel count of the original game, this is likely due to both the difficulty in implementing higher resolution assets at least in terms of alignment and lack of access to the SGI workstations for rendering new versions of the background... the reason this is an issue is that the real-time 3D elements no longer matches the resolution of the backgrounds which creates a sort of mismatch between the two"
 ~~29:45 Battle "The distant building contrasts incorrectly against the background"~~ Fixed with: "Shinra Archaeology Cut" mod available on 7H catalog
 ~~29:34 Battle "The shadows below the character's exhibit a more defined outline"~~ Fixed with: "PSX Battle Shadows and Vertex Snapping" mod available on 7H catalog 
 ~~29:56 Battle "Floating rocks all exhibit black outlines as a result of the texture filtering"~~ Not an issue in FFNx
 ~~30:13 Battle "Leviathan Summon problem: textures on models exhibit visible black lines that aren't present on Playstation"~~ Not an issue in FFNx

[...]

PR #893, opened 2026-02-20
FF7: Fix Ifrit summon ground hole rendering

jakobdowdle:
## Summary

Fixes Ifrit summon ground hole rendering by forcing subtractive blending on the two draw calls that make up the hole.

### Motivation

The PC port lost the original PSX subtractive blending on Ifrit's ground hole, so it shows up as white instead of black. This fixes issue #654 and brings the effect closer to the original PSX look.

### ACKs

### DIFF
--- a/src/gl/gl.cpp
+++ b/src/gl/gl.cpp
@@ -32,6 +32,7 @@
 #include "../matrix.h"
 
 #include "../ff7/widescreen.h"
+#include "../globals.h"
 #include "external_mesh.h"
 
 struct matrix d3dviewport_matrix = {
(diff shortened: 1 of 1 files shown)


julianxhokaxhiu:
Hey Jakob, thank you very much for your first contribution. While the patch might do the job well, unfortunately it seems too generic and patching a GPU pipeline instead of the root cause. We are aware in general FF7 and FF8 code has a blending issue spread around and we are tackling it differently, fixing the problem at the root which of course requires time and knowledge.

Have you got experience reverse engineering? If yes, feel free to join our Discord and we'll be happy to have a chat together. While I really appreciate the PR, unfortunately I can't merge it in its current state. Hope you do understand.

Looking forward nevertheless, thanks!

zaphod77:
The root cause is 100% "game specified wrong blend mode."  It's an issue with fades, it's an issue with this, and it's likely the reason shadow flare is wrong.

Certain effects require subtractive blend, which was not supported by the video cards of the time period. that's the entire reason for all of this trouble. They also only allowed up to 256X256 textures, which is why the higher resolution interface textures were broken up into two or four.

The correct approach is to AB compare the code between PC and PSX, identifying all code paths that are supposed to be subtractive blend, and patching them in the driver.

jakobdowdle:
I really appreciate the feedback! I totally understand and I figured this approach might be too broad, but it was a good learning experience tracking down the issue with RenderDoc. 

> Have you got experience reverse engineering? If yes, feel free to join our Discord and we'll be happy to have a chat together. While I really appreciate the PR, unfortunately I can't merge it in its current state. Hope you do understand.

I don't have any reverse engineering experience, but I'd love to learn and help tackle the root cause properly. I'm already in the Qhimm Discord as JakeIsYoshi, so I'll send a message over there as well.

> The correct approach is to AB compare the code between PC and PSX, identifying all code paths that are supposed to be subtractive blend, and patching them in the driver.

This would be new territory for me, but I would be super interested in trying it out.

julianxhokaxhiu:
I'm closing this one as I would prefer a fix that lives in the game engine code rather than a workaround by catching shader logics and whatnot. But please feel free to open a new PR ( or update the branch and let me know ) in case you find a real fix. Thanks a lot!

zaphod77:
The issue is literally the game setting the wrong blend mode.

Either the blend mode for the hole is incorrect in the battle model itself, the engine is not correctly parsing the blend mode from the battle model file, or there is code somewhere that should request a subtractive blend but doesn't because they didn't exist in the early direct3d.  

I'm suspecting number 3.

PR #898, opened 2026-02-22
FF8: Add chunk feature for battle ennemies c0mXXX.dat files

myst6re:
## Summary

Add chunk feature for battle ennemies c0mXXX.dat files

### Motivation

Allow modders to publish only a modded section of the file (for example the section 7, aka monster stats)

### ACKs

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -14,6 +14,9 @@
 
 ## FF8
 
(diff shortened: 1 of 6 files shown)


HobbitDur:
I can see that to be useful, specially to separate the stat from the AI.
But that would noot allow to change the monster name, as it's the only think in the stat section.
Not a big deal, but wanted to point it out.

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you!

PR #905, opened 2026-02-28
FF7: Implement achievements for 2026 Re-Release edition

tangtang95:
## Summary

Implement all achievements for 2026 Re-Release edition. Some new achievements are still not tested though, but we can test it afterward.

### Motivation

Support new FF7 2026 edition

### ACKs


julianxhokaxhiu:
Boy if you were fast :) Thanks a lot!

tangtang95:
Yeah took only a day. I decided to share the achievement class so that the variable is the same and it's not possible to have any null pointer for the global achievement pointer. Also because most of the achievements are shared

PR #913, opened 2026-03-14
FF8: clear texture packer on module loading

myst6re:
## Summary

Clear texture packer on module loading

### Motivation

When you reset the game, sometimes FFNx try to load external textures with irrelevant names. Those are artifacts from the previous module, field for example.

To ensure this do not happen, we can clear the textures we identified when we change modules. This is already the case for battle.

Field and worldmap modules cannot be cleared like we do for battle, because we can switch to the menu and come back without reloading the textures. In this case, when we come from the menu, we don't want to clear the textures. So in field and worldmap, we must clear the textures just before the game loads the main textures.

This change could help with the remaining texture bleeding issues.

### ACKs

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -13,6 +13,7 @@
 ## FF8
 
 - Core: Fix random crash when battles start ( https://github.com/julianxhokaxhiu/FFNx/pull/912 )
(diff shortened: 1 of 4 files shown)


julianxhokaxhiu:
Thank you!

PR #914, opened 2026-03-20
Common: Fix bilinear filtering

myst6re:
## Summary

`enable_bilinear` enables linear filtering when the game asks to for internal textures

### Motivation

Official Steam versions of ff7 and ff8 have a linear filtering option

### ACKs

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,10 @@
 
 - Full commit list since last stable release: https://github.com/julianxhokaxhiu/FFNx/compare/1.24.3...master
 
+## Common
+
+- Renderer: fix `enable_bilinear` option for original game textures ( https://github.com/julianxhokaxhiu/FFNx/pull/914 )
+
 # 1.24.3
 
 - Full commit list since last stable release: https://github.com/julianxhokaxhiu/FFNx/compare/1.24.2...1.24.3
--- a/src/ff8/uv_patch.cpp
+++ b/src/ff8/uv_patch.cpp
@@ -138,11 +138,24 @@ int pubintro_psxvram_buffer_init_sub_45B310()
 
 	// Divide by 255 instead of 256, and don't alter values of psx_floats1[0] and psx_floats1[255]
 	for (int i = 0; i < 256; ++i) {
-		ff8_externals.psx_floats1[i] = float(double(i) / 255.0);
+		ff8_externals.psx_floats1[i] = float(double(i) / (enable_bilinear ? 256.0 : 255.0));
 	}
 
 	for (int i = 0; i < 512; ++i) {
-		psx_floats512[i] = float(double(i) / 511.0);
+		psx_floats512[i] = float(double(i) / (enable_bilinear ? 512.0 : 511.0));
+	}
+
+	if (enable_bilinear)
+	{
+		ff8_externals.psx_floats1[0] = ff8_externals.psx_floats1[1];
+		ff8_externals.psx_floats1[255] = ff8_externals.psx_floats1[254];
+
+		psx_floats512[0] = psx_floats512[3];
+		psx_floats512[1] = psx_floats512[3];
+		psx_floats512[2] = psx_floats512[3];
+		psx_floats512[509] = psx_floats512[508];
+		psx_floats512[510] = psx_floats512[508];
+		psx_floats512[511] = psx_floats512[508];
 	}
 
 	return ret;
--- a/src/gl/special_case.cpp
+++ b/src/gl/special_case.cpp
@@ -41,28 +41,28 @@ uint32_t gl_special_case(uint32_t primitivetype, uint32_t vertextype, struct nve
 	uint32_t mode = getmode_cached()->driver_mode;
 	VOBJ(texture_set, texture_set, current_state.texture_set);
 	uint32_t defer = false, force_defer = false;
+	bool isExternalTexture = current_state.texture_set && VREF(texture_set, ogl.external);
 
-	// modpath textures rendered in 3D should always be filtered
-	if(vertextype != TLVERTEX && current_state.texture_set && VREF(texture_set, ogl.external)) current_state.texture_filter = true;
-
-	// modpath textures in menu should always be filtered
-	if((mode == MODE_MENU || mode == MODE_MAIN_MENU) && current_state.texture_set && VREF(texture_set, ogl.external)) current_state.texture_filter = true;
+	// Force disabling filter for ff7 minigames internal textures
+	if(!ff8 && (mode == MODE_SUBMARINE || mode == MODE_COASTER) && !isExternalTexture) current_state.texture_filter = false;
 
 	// some modpath textures have filtering forced on
-	if(current_state.texture_set && VREF(texture_set, ogl.gl_set->force_filter) && VREF(texture_set, ogl.external)) current_state.texture_filter = true;
-
-	// Texture filtering mostly does not work well in FF8
-	if(ff8) current_state.texture_filter = enable_bilinear && vertextype != TLVERTEX && current_state.texture_set;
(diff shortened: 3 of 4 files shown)


zaphod77:
So you fixed the edge lines?  Has this been tested at *all* resolutions?

My original pull blacklisted internal speed square hud, submarine hud, btl_win overlays, fonts, and the field hand from filtering in special case, because they have single pixel edge lines, especially with anisotropic off. condor, g-bike, snowboard, and chocobo racing are fine. it also blacklisted ff8 internals, to keep existing behavior with ff8. These issues are why internal textures were blacklisted from filtering in the first place, and why i had to test everything after the change, and put in a flag to make my version not apply to ff8.

Has this been fixed, and tested with anisotropic off?  

Also, that change to renderer removes the force that actually made it so ALL movies were unfiltered. to restore that behavior, the line that filters all movies in src\gl\gl.cpp needs to be inverted to unfilter all movies, because **no** movie ever has the external texture flag set.

The movies should actually be the same filtering state as the field graphics. which is unfiltered, according to the game, unless you forced it on in steam 2013. the 2026 releases force it ON, for parity with the upscaled and blurry movies.

myst6re:
For now I keep it simple, for external textures the behavior is unchanged. For internal I enable linear filtering only if the game ask to AND the `enable_bilinear ` option is ON.

I'm sure some rules will need to be added later to fine-tune where filtering is enabled or not.

For FF8 on my testing it enables filtering for 3D models, worldmap and battle. I had to disable some UV changes to prevent holes and glitches. Field background is not filtered

For FF7 I haven't tested the minigames. Battle, field and worldmap seems ok

zaphod77:
Test battle, field, and minigames with anisotropic off, and bilinear on, please. in non 640X480 (that part is important). if this glitches, then please add in the following special cases.

```
  // modpath textures in menu should always be filtered
  if((mode == MODE_MENU || mode == MODE_MAIN_MENU) && current_state.texture_set && VREF(texture_set, ogl.external)) current_state.texture_filter = true;

+  // submarine and speed square huds glitch if filtered and they are internal textures and resolution is not 640X480.
+  if (current_state.texture_set && !ff8 && !(VREF(texture_set, ogl.external)))
+  {
+    if ((mode == MODE_SUBMARINE) || (mode == MODE_COASTER))
+    {
+      current_state.texture_filter = false; // required to avoid thin lines at the edges of the HUD textures in these modes. 
+    }
+  }
```
```
	// avoid filtering window borders
	if(!_strnicmp(VREF(tex_header, file.pc_name), "menu/btl_win_c_", strlen("menu/btl_win_c_") - 1) && VREF(texture_set, palette_index) == 0) current_state.texture_filter = false;
+	// the following internal textures can glitch when filtered in ff7.
+	if (!(VREF(texture_set, ogl.external)))
+	{
+		if (!ff8 && (!_strnicmp(VREF(tex_header, file.pc_name), "menu/btl_win_", strlen("menu/btl_win_") - 1)))
+		{
+			current_state.texture_filter = false; // shows lines with anisotropic filtering off 
+		}
+		if (!ff8 && (!_strnicmp(VREF(tex_header, file.pc_name), "menu/usfont_", strlen("menu/usfont_") - 1)))
+		{
+			current_state.texture_filter = false; // shows lines with anisotropic filtering off
+		}
+		if (!ff8 && (!_strnicmp(VREF(tex_header, file.pc_name), "menu/jafont_", strlen("menu/jafont_") - 1)))
+		{
+			current_state.texture_filter = false; // just in case. 
+		}
+		if (!ff8 && (!_strnicmp(VREF(tex_header, file.pc_name), "flevel/hand", strlen("flevel/hand") - 1)))
+		{
+			current_state.texture_filter = false; // the famous finger also glitches. 
+		}
+	}

```
Also remember what i said about movies.  

```
	current_state.texture_filter = true;
```

needs to be false to restore previous behavior.

myst6re:
Okay I add the minigame check

For btl_win_, font_, hand, the game does not ask for filtering anyway

For movies, I don't know if I can disable filtering only for internal movies. I should keep the filtering for external movies, right?

zaphod77:
When I tried it, I saw those glitches when I did not. Did you test with anisotropic off? That was required to reveal the lines. 

There is no such thing as an external movie. They are ALL considered the same by the game and were all unfiltered because of the check in render, which overrode the check in gl. No movie has ever been filtered. The square movies were filtered in the upscale, which was confusing. Since movies are intended to integrate with field assets, they should be unfiltered as long as the game is not filtering the field.

myst6re:
> Did you test with anisotropic off?

Yes, I'm pretty sure my PR solves your issues

myst6re:
> There is no such thing as an external movie. They are ALL considered the same by the game and were all unfiltered because of the check in render, which overrode the check in gl. No movie has ever been filtered. The square movies were filtered in the upscale, which was confusing. Since movies are intended to integrate with field assets, they should be unfiltered as long as the game is not filtering the field.

I'm not sure it is my call here, @julianxhokaxhiu do you have any thoughts about this?

zaphod77:
Main menu with anisotropic off, and 1280x960, without the special cases.

image

lines in battle mode with text.

image

Lines above makou reactor 1 timer.

image


It's hard to screen shot the field cursors, because the line is only visible for a couple frames of the animation.

one of these flags is what fixes the text and other stuff.

BGFX_SAMPLER_MIN_POINT | BGFX_SAMPLER_MAG_POINT | BGFX_SAMPLER_MIP_POINT

even though the text isn't filtered, there's something wrong with it.

I can't seem to reproduce the issue with the hand, though.

strangely enough, ja_font is fine..

image

myst6re:
Is it with or without my PR?

zaphod77:
The only thing i haven't put in my fork for testing is the uv fix, which is ff8 specific, and my fork is pretty recent.  My card is nvidia, but i tested all the graphics apis, with identical results.  without forcing point sampling for min, mag, and mip, at 720p or higher, the glitches pop up.

The issue may indeed be nvidia specific, as i distinctly remember this happening with old resolution increasing patches.  Is your card AMD?

https://github.com/zaphod77/FFNx/compare/master...julianxhokaxhiu%3AFFNx%3Amaster

I have not yet committed the bilinear fix, because JP text support isn't in mainline yet, but i see nothing there that would affect this.

When the game doesn't request bilinear, my guess is that not all of those three render flags are set.  The correct action is to forcibly enable all three flags when the game requests unfiltered textures,  But i don't know how to detect if the game requested filtering or not.  hence me forcing unfiltered, even though they already are.

julianxhokaxhiu:
> > There is no such thing as an external movie. They are ALL considered the same by the game and were all unfiltered because of the check in render, which overrode the check in gl. No movie has ever been filtered. The square movies were filtered in the upscale, which was confusing. Since movies are intended to integrate with field assets, they should be unfiltered as long as the game is not filtering the field.
> 
> I'm not sure it is my call here, @julianxhokaxhiu do you have any thoughts about this?

AFAIK when I was working on that part I adapted the rendering code to respect the logics made then by Aali, and I decided to inherit it since a lot of mods were made already with that logic in place. Since now the number of mods has exploded, I would say let's keep the same behavior ensuring compatibility. I would never filter any movie in general such as not to introduce an unintended behavior in the game.

zaphod77:
Yeah, that was my feeling too. but the other code was trying to filter all
movies, and was dead code. but with the new pull request it's not, so it
needs fixing.

Unless ffnx will someday allow filtering the fields, then no point in
allowing filtering for movies.

zaphod77:
The jp font is fine because it's a higher resolution. I can't get the resolution high enough to show glitches.  but btl_win and usfont definitely need it on my graphics card.  I dunno what i was thinking with the flevel\hand bit.

julianxhokaxhiu:
Thanks a lot :)

PR #917, opened 2026-03-22
FF7: Add non-Steam2013 userdata paths

myst6re:
## Summary

The new rerelease 2026 version of Final Fantasy VII put the save files into AppData/Local.

### ACKs

- N/A I did test my code on FF8

### DIFF
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -135,6 +135,12 @@ uint32_t steam_stock_launcher = false;
 // global FF7 flag, check if is eStore edition
 uint32_t estore_edition = false;
 
+// global FF7 flag, check if is gog edition
+uint32_t gog_edition = false;
+
+// global FF7 flag, check if is windows store edition
+uint32_t windows_store_edition = false;
+
 // global FF7 flag, check if it is 2026 rerelease
 uint32_t ff7_2026_rerelease = false;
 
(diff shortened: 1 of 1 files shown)


julianxhokaxhiu:
What happened with this PR?

myst6re:
> What happened with this PR?

I wonder if the objective of ffnx is to natively find the save files where they officially are.

I saw your previous commit about supporting 2026's saves and I'm not sure what's the plan anymore

Also I don't know how to support the saves for the windows store version

julianxhokaxhiu:
Ah I see, while I think it would be a nice to have, I was more worried this would cause a set of issues which needs to be tested. I think we can safely start to introduce it on the 1.25 cycle, I just didn't want to introduce this in the middle of a bugfix release :) Feel free to open it again, happy to review and merge if it makes sense.

Also, don't worry if it doesn't support all scenarios for now, having the most common ones, like Steam or GOG, is already a huge step forward nevertheless. Thanks!

zaphod77:
the windows store version can't be supported this way.

the save files in it no longer have the extension, are in a very arcane path, and do a force cloud sync that will erase anything that's put there with ffnx.

julianxhokaxhiu:
I think this PR became kinda rotten compared to the current status of the code. I'd propose to close it if ok with you all since we're not aiming at supporting native save paths anyway.

PR #935, opened 2026-05-03
[ FF7 ] Ground beneath Ifrit: color inverted - v2

jakobdowdle:
## Summary

Fixes Ifrit summon ground hole rendering with two changes: a SetBlendMode trampoline hook so per-state blend modes actually reach the renderer, and a one-byte patch on the Ifrit summon's CIRCLE.RSD (the ground hole model) descriptor that flips it from additive to subtractive blending.

### Motivation

The PC port lost the original PSX subtractive blending on Ifrit's ground hole, so it shows up as white instead of black. This fixes issue #654 and brings the effect closer to the original PSX look.

### More Info

The descriptor for CIRCLE.RSD  in Ifrit's table hard-codes blend mode 1 (additive); PSX used mode 2 (subtractive). Two changes:

Data patch at 0x008bfee8 (0xb1 -> 0xb2), preserving flag bits. The address is referenced only by the Ifrit summon loader, so the change shouldn't bleed into other effects. Resolved dynamically through the run_ifrit_main call chain. Supports US/DE/FR/SP.
SetBlendMode trampoline hook that writes the PSX index into p_hundred -> blend_mode (+0x44) after the original runs. The original sets D3D factors but never this field (which is what common_setrenderstate reads) so the descriptor flip wouldn't reach BLEND_SUB without it.

### ACKs

### DIFF
--- a/src/externals_102_de.h
+++ b/src/externals_102_de.h
@@ -63,6 +63,7 @@ ff7_externals.matrix4x3_multiply =            0x66CC0B;
 ff7_externals.sub_6B26C0 =                    0x6B2690;
 ff7_externals.sub_6B2720 =                    0x6B26F0;
 ff7_externals.sub_673F5C =                    0x673F2C;
+ff7_externals.set_blend_mode =                0x694C50;
 ff7_externals.savemap =               (savemap *)0xF38A68;
 ff7_externals.menu_objects =          (menu_objects *)0xF39CF0;
 ff7_externals.magic_thread_start =            0x427928;
--- a/src/externals_102_fr.h
+++ b/src/externals_102_fr.h
@@ -63,6 +63,7 @@ ff7_externals.matrix4x3_multiply =            0x66CBDB;
 ff7_externals.sub_6B26C0 =                    0x6B2660;
 ff7_externals.sub_6B2720 =                    0x6B26C0;
 ff7_externals.sub_673F5C =                    0x673EFC;
+ff7_externals.set_blend_mode =                0x694C20;
 ff7_externals.savemap =               (savemap*)0xF39A78;
 ff7_externals.menu_objects =          (menu_objects*)0xF3AD00;
 ff7_externals.magic_thread_start =            0x427938;
--- a/src/externals_102_sp.h
+++ b/src/externals_102_sp.h
@@ -63,6 +63,7 @@ ff7_externals.matrix4x3_multiply =            0x66CBDB;
 ff7_externals.sub_6B26C0 =                    0x6B2660;
 ff7_externals.sub_6B2720 =                    0x6B26C0;
 ff7_externals.sub_673F5C =                    0x673EFC;
+ff7_externals.set_blend_mode =                0x694C20;
 ff7_externals.savemap =               (savemap *)0xF3A548;
 ff7_externals.menu_objects =          (menu_objects *)0xF3B7D0;
 ff7_externals.magic_thread_start =            0x427938;
--- a/src/externals_102_us.h
+++ b/src/externals_102_us.h
@@ -63,6 +63,7 @@ ff7_externals.matrix4x3_multiply =            0x66CC3B;
 ff7_externals.sub_6B26C0 =                    0x6B26C0;
 ff7_externals.sub_6B2720 =                    0x6B2720;
 ff7_externals.sub_673F5C =                    0x673F5C;
+ff7_externals.set_blend_mode =                0x694C80;
 ff7_externals.savemap =               (savemap *)0xDBFD38;
 ff7_externals.menu_objects =          (menu_objects *)0xDC0FC0;
 ff7_externals.magic_thread_start =            0x427928;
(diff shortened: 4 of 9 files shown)


julianxhokaxhiu:
Thanks for opening this PR, but again, I see two problems with this approach:
- You are trying to patch a data file in memory directly, which is outside of the scope of how we've done things here ( unless the fix makes sense from the code perspective, such as 2026 special fixes to not make the game crash ),
- You are adding an unrequired code patch when we govern the `set_blend_mode` that is part of the driver interface anyway, which lives in common.

In any case, I don't think patching the sub is the right way for it, but since we know finally where the bug is, can't we just document the fix in the file and let every mod deliver this one fixed? For eg, SAC/Echo-S/etc. And eventually, someone can build an independent mod that does only this.

What do you think?

jakobdowdle:
Hey I apologize, I was out on vacation right after opening this PR so I didn't get back to this until now!

> You are trying to patch a data file in memory directly, which is outside of the scope of how we've done things here ( unless the fix makes sense from the code perspective, such as 2026 special fixes to not make the game crash )

That makes sense. I'm not sure if there's a better way to do it, I tried a few other things first and this just ended up being the only thing I actually got working. Here's what I tried before:

- A replacement hook on the battle overlay dispatcher (`FUN_0063ac20` → `DAT_00cffad8`). As far as I could tell this was a dead end because the overlay polygon_sets never actually reach `common_draw_2D` in FFNx. The master list dispatch path (`field_EC` → `FUN_006a1a37`) doesn't actually do anything, so the hook had nothing to drive.
- Reimplementing that master-list dispatch myself, walking `game_obj+0x2E8` and pushing queued type-2 entries through `gl_draw_indexed_primitive`. It did dispatch ~1400× per battle, but it also double-drew ~1379 mode-1 effect polygon sets that `ff7gl_field_78` had already handled. The post battle fade also went white instead of black, so I reverted it.
- A byte patch on a different branch (`PUSH 0x01` → `PUSH 0x02` in `FUN_0065a0a2`). That branch turned out to handle RSD indices 0x94..0xAC, and none of those are Ifrit.

The descriptor byte at `0x008bfee8` was the narrowest thing I could find that only hits Ifrit, as far as I could tell. It's referenced exactly once in the entire `ff7_en.exe` (only by `FUN_00592720`, the Ifrit only summon loader), so I don't imagine it would bleed into anything else. But I do understand the precedent you guys have set here, and I'm fine following whatever you think is best.

> You are adding an unrequired code patch when we govern the `set_blend_mode` that is part of the driver interface anyway, which lives in common.

I might be misunderstanding this one, but here's what I was seeing: `common_setrenderstate` in `common.cpp` does own the final bgfx blend mapping. It reads its input from `hundred_data->blend_mode` (`p_hundred + 0x44`), and that field has to be a PSX index 0-4. The game's internal `SetBlendMode @ 0x694C80` writes the D3D src/dst factors at `+0x34/+0x38`, but for the descriptor-driven path used here, it doesn't put the PSX index into `+0x44`. So the trampoline is just there to capture the index argument and store it in `+0x44` after the original runs. I don't think it's duplicating `common_setrenderstate`, it's making sure the byte from the descriptor actually reaches the field `common_setrenderstate` reads.

That said, I'm happy to just apply the byte patch without the trampoline to see if the hole still goes black. That would at least confirm whether the trampoline is actually doing anything.

> can't we just document the fix in the file and let every mod deliver this one fixed? For eg, SAC/Echo-S/etc.

Totally open to that. The most important thing for me is just getting the discovery trace on the record somewhere (`FUN_00592720` → `CreateRenderBatch` → descriptor `0x008bfee8`). How would you want something like this documented? Happy to put it wherever fits best.

julianxhokaxhiu:
Thanks for coming back and clarifying the approach. I think we're focusing although again on the wrong place to patch.

If I look correctly at your PR you are trying to patch the `circle.rsd` data file already loaded in memory via FFNx, which however we usually don't do here unless it's strictly required to prevent game crashes on top of the vanilla game data ( see for eg. 2026 fixes ). I think the file you're trying to patch exists on this path in the game data folder

immagine

or this one maybe

immagine

This means that it could probably be replaced using `direct/` folder in FFNx and build a mod for it and add it to 7th catalog. The documentation part could be explaining how did you patch the file to fix it basically. But I don't think a game code patch is needed, because all you're doing here is fixing the blend mode for this 3D model file ( probably a bug during the conversion or missing game code from the PSX porting ).

What I suggest here is try not to focus on FFNx at all, but just on the file, try replacing it using the `direct/` mod path and see how it goes :)

Hope this helps, thanks!

zaphod77:
The issue here is that the PC data is flat out wrong, and should be corrected by ffnx if the file passes a checksum test.

KuraudoXS:
It wouldn’t be the first time the FFNx team has treated us to new fixes, and I believe this is a specific case/fix that deserves to be included in FFNx (as a vanilla PC bug). 
Just as an example, we can see that similar fixes relating to Summon effects have been included in the past. 
[“(FIX: A lot of missing effects in Summons (e.g. Kujata’s earthquake, or Odin’s slash effects)”](https://github.com/julianxhokaxhiu/FFNx/blob/master/Changelog.md#150) 

Please, TrueOdin, consider implementing this (even if it means tweaking something else in the code or take a different approach).

PR #948, opened 2026-07-09
FF7: Add multibyte_font mode for non-Japanese multibyte text

### LINKED ISSUE #39: [ FF7 ] Japanese support

faisalkindi:
Adds a `multibyte_font` config flag, separate from `ff7_japanese_edition`. It enables the extra font glyph space (jafont_2..6 via escape codes) and the multibyte width/draw system, for English-exe translations that need more glyphs than the base charset gives you — I built this against Arabic. It skips the JP-only menu scaling/layout, since that breaks native English menus.

Three engine bugs turned up while testing this — all general correctness bugs, not multibyte-specific, I just happened to hit them here:

- `field_submit_draw_text_640x480_jp` incremented a pointer instead of dereferencing it when counting newlines, so the field engine's line-row counter never moved. The window-close handshake waits on that counter, so multi-line messages could stall indefinitely.
- `field_text_box_window_opening_jp` could leave a window active with no owner assigned if it got there via the multibyte draw path — the owner check that lets a window grow then never passes, and the field script waiting on it hangs.
- Same function's auto-resize recomputed its grow target from the previous frame's output every frame. On multi-window scenes the target kept moving and the grow animation never finished.
- `draw_text_top_display_jp` (battle top message bar) was only hooked for `ff7_japanese_edition`. Multibyte text through this specific bar fell through to the vanilla single-byte draw and came out garbled. Moved it into the shared hook block.

Also added a live width-tuning system (`multibyte_widths.bin` / `multibyte_linestep.bin`, hot-reloaded off disk), so a translation project can tune per-character spacing without rebuilding FFNx.

Tested against a full Arabic localization of FF7, 713 field maps. The window-stall fix resolves what was a 100%-reproducible soft-lock. The top-display fix resolves garbled battle text. No regressions in `ff7_japanese_edition` — everything new is gated behind the flag or falls back to the existing JP path.

### DIFF
--- a/docs/ff7/multibyte_font.md
+++ b/docs/ff7/multibyte_font.md
@@ -53,7 +53,7 @@ The loader reads the 6 sheets by the same names the JP edition uses:
 
 Each sheet is a 16×16 grid of glyph cells (256 cells per sheet). Author them exactly like any
 other FFNx-replaceable menu texture — external hi-res textures are supported through the normal
-texture replacement path (see [External textures](external_textures.md)).
+texture replacement path (see [External textures](../mods/external_textures.md)).
 
 ## Per-character widths: `multibyte_widths.bin`
 
--- a/docs/ff7/multibyte_font.md
+++ b/docs/ff7/multibyte_font.md
@@ -59,7 +59,10 @@ texture replacement path (see [External textures](../mods/external_textures.md))
 
 Glyph advance widths come from a table you can override **at runtime, without recompiling**:
 
-- Path: `<basedir>/multibyte_widths.bin` (next to `FFNx.toml`).
+- Path: `data/kernel/multibyte_widths.bin`, resolved through the standard layers — the
+  `override_path` layer is checked first, then the release's data path: `data/lang-en/kernel/`
+  (or the active language) on Steam/GOG/Store/2026 releases, plain `data/kernel/` on the 1998
+  release.
 - Format: raw `6 * 256` bytes — one byte per glyph, sheets in order 1..6, code order `0x00-0xFF`
   within each sheet.
 - Byte packing: `(left_padding << 5) | width`, i.e. low 5 bits = advance width in font units
@@ -73,7 +76,7 @@ If the file is missing, built-in defaults are used.
 
 Optional. Controls the vertical line advance of field dialogue for multibyte text:
 
-- Path: `<basedir>/multibyte_linestep.bin`.
+- Path: `data/kernel/multibyte_linestep.bin` (same layered resolution as the widths file).
 - Format: a 2-byte little-endian value, line step in **quarter pixels** (e.g. `128` = 32.0 px,
   `98` = 24.5 px). A 1-byte file is also accepted as whole pixels. Accepted range 80–160
   quarter-px; out-of-range values are ignored.
@@ -87,7 +90,7 @@ default; adjust live until lines neither overlap nor float apart.
 Optional, read once at startup. Marks sheet-1 cells that contain icon art (item/weapon type
 icons, button prompts) rather than letters:
 
-- Path: `<basedir>/multibyte_iconmask.bin`.
+- Path: `data/kernel/multibyte_iconmask.bin` (same layered resolution as the widths file).
 - Format: raw 256 bytes, one per `jafont_1` code. Non-zero = icon cell.
 - Effect: marked cells are always drawn pure white instead of taking the current text color,
   so colored icon art keeps its true colors. This addresses the recolor conflict with icon
--- a/src/cfg.cpp
+++ b/src/cfg.cpp
@@ -297,7 +297,7 @@ void read_cfg()
 	ff7_fps_limiter = config["ff7_fps_limiter"].value_or(FPS_LIMITER_DEFAULT);
 	ff7_footsteps = config["ff7_footsteps"].value_or(false);
 	ff7_field_center = config["ff7_field_center"].value_or(true);
-	multibyte_font = config["multibyte_font"].value_or(false);
+	ff7_multibyte_font = config["ff7_multibyte_font"].value_or(false);
 	use_sdl_gamepad = config["use_sdl_gamepad"].value_or(false);
 	ff7_japanese_text = config["ff7_japanese_text"].value_or(false);
 	ff7_japanese_edition = config["ff7_japanese_edition"].value_or(false);
--- a/src/cfg.cpp
+++ b/src/cfg.cpp
@@ -297,6 +297,7 @@ void read_cfg()
 	ff7_fps_limiter = config["ff7_fps_limiter"].value_or(FPS_LIMITER_DEFAULT);
 	ff7_footsteps = config["ff7_footsteps"].value_or(false);
 	ff7_field_center = config["ff7_field_center"].value_or(true);
+	multibyte_font = config["multibyte_font"].value_or(false);
 	use_sdl_gamepad = config["use_sdl_gamepad"].value_or(false);
 	ff7_japanese_text = config["ff7_japanese_text"].value_or(false);
 	ff7_japanese_edition = config["ff7_japanese_edition"].value_or(false);
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -145,7 +145,7 @@ uint32_t ff7_japanese_edition = false;
 // multibyte font mode: enable the extra-font escape codes (FA-FE -> jafont_2..6) WITHOUT the
 // JP-only menu scaling/name-screen, so English-exe translations (e.g. Arabic) get the extra
 // glyph space while keeping native EN menu layout.
-uint32_t multibyte_font = false;
+uint32_t ff7_multibyte_font = false;
 
 // window dimensions requested by the game, normally 640x480
 uint32_t game_width;
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -142,6 +142,11 @@ uint32_t ff7_2026_rerelease = false;
 // global FF7 flag, check if is japanese edition ( detected as US )
 uint32_t ff7_japanese_edition = false;
 
+// multibyte font mode: enable the extra-font escape codes (FA-FE -> jafont_2..6) WITHOUT the
+// JP-only menu scaling/name-screen, so English-exe translations (e.g. Arabic) get the extra
+// glyph space while keeping native EN menu layout.
+uint32_t multibyte_font = false;
+
 // window dimensions requested by the game, normally 640x480
 uint32_t game_width;
 uint32_t game_height;
--- a/src/ff7/japanese_text.cpp
+++ b/src/ff7/japanese_text.cpp
@@ -18,8 +18,11 @@
 #include "../patch.h"
 #include <string.h>
 
+static void multibyte_load_widths();
+
 void engine_load_menu_graphics_objects_6C1468_jp(int a1)
 {
+  multibyte_load_widths();
   unsigned int v1; // eax
   unsigned int v2; // eax
   unsigned int v3; // ecx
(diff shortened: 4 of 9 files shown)


julianxhokaxhiu:
@CosmosXIII since this PR contains commits coming originally from your PR #737 I would ask for a review please

@faisalkindi thanks a lot for this PR, as soon as I manage to get the feedback from Cosmos I will most likely ask you to rebase so we merge them in the right order.

Thank you

zaphod77:
Yeah, my routine has many jp specific hacks in it.

Also, multibyte character support directly conflicts with many other very common hacks used for stuff like button prompts and field/battle/menu icons, because it forces the color data to actually recolor the multibyte font instead of being able to read different versions of it.  

Would this one also be able to apply a Chinese or Korean translation?  DO keep in mind that any translation requiring a multibyte character set will also require a multibyte name screen.

zaphod77:
Okay, my stuff is getting merged into main, so this will need a rebase after that.

zaphod77:
Okay my stuff is in, BUT...

Turns out there's is one silly special case in the JP text.

Hex 0xD9 is redirected to the heart in btl_win_d. This is not currently implemented in jp text support, but implementing this will break any other multibyte translation that thinks to use the spot.

julianxhokaxhiu:
Can you please rebase your PR on top of the latest master? So it will be easier to review it. Thank you

faisalkindi:
Rebased on latest master as requested — now a single commit. Resolved against the new heart-icon (#952/#953) and name-entry (#951) work: the shared multibyte hooks coexist with both, and the JP-only name-entry install stays gated behind `ff7_japanese_edition`.

@zaphod77 re: 0xDE — checked our Arabic charmap: 0xDE is unused in both the main map and the relocated-glyph plane, so implementing the heart redirect doesn't collide with this translation. Agreed it's worth documenting as a reserved slot for any future multibyte charset.

Re: Chinese/Korean — the flag adds the glyph-space plumbing and width system only; a multibyte name screen isn't included (Arabic doesn't need one), but the hook structure should serve as a base for it.

src/cfg.cpp
```diff
@@ -297,6 +297,7 @@ void read_cfg()
 	ff7_fps_limiter = config["ff7_fps_limiter"].value_or(FPS_LIMITER_DEFAULT);
 	ff7_footsteps = config["ff7_footsteps"].value_or(false);
 	ff7_field_center = config["ff7_field_center"].value_or(true);
+	multibyte_font = config["multibyte_font"].value_or(false);
```
julianxhokaxhiu: Please prefix the flag with `ff7_`

src/common.cpp
```diff
@@ -142,6 +142,11 @@ uint32_t ff7_2026_rerelease = false;
 // global FF7 flag, check if is japanese edition ( detected as US )
 uint32_t ff7_japanese_edition = false;
 
+// multibyte font mode: enable the extra-font escape codes (FA-FE -> jafont_2..6) WITHOUT the
+// JP-only menu scaling/name-screen, so English-exe translations (e.g. Arabic) get the extra
+// glyph space while keeping native EN menu layout.
+uint32_t multibyte_font = false;
```
julianxhokaxhiu: same here

src/ff7_opengl.cpp
```diff
@@ -377,17 +377,41 @@ void ff7_init_hooks(struct game_obj *_game_object)
...
-	if (ff7_japanese_edition)
+	// Shared multi-font glyph space (extra jafont_2..6 via FA-FE escape bytes): needed by BOTH the JP
+	// edition AND English-exe multibyte translations (e.g. Arabic). Only the font LOAD + per-char draw
+	// are shared; the remaining hooks below are JP-only layout that breaks native EN menus.
+	ffnx_info("multibyte hook install check: multibyte_font=%d ff7_japanese_edition=%d\n", multibyte_font, ff7_japanese_edition);
+	ffnx_info("mb-ext: load_menu=%X submit_char=%X sub_6F54A2=%X field_submit=%X field_boxes=%X menu_draw=%X battle_draw=%X mainmenu_draw=%X\n",
+		(uint32_t)ff7_externals.engine_load_menu_graphics_objects_6C1468, (uint32_t)ff7_externals.common_submit_draw_char_from_buffer_6F564E,
+		ff7_externals.sub_6F54A2, ff7_externals.field_submit_draw_text_640x480_6E706D,
+		(uint32_t)ff7_externals.field_draw_text_boxes_and_text_graphics_object_6ECA68, (uint32_t)ff7_externals.menu_draw_everything_6CC9D3,
+		(uint32_t)ff7_externals.battle_draw_menu_everything_6CEE84, (uint32_t)ff7_externals.main_menu_draw_everything_maybe_6C0B91);
```
julianxhokaxhiu: Please fix indentation and remove the `ffnx_info` calls, they will just be noise for the average player

src/ff7_opengl.cpp
```diff
@@ -396,7 +420,7 @@ void ff7_init_hooks(struct game_obj *_game_object)
 		patch_code_byte(0x632C4E + 0x3, 0xC);
 		patch_code_byte(0x632C4E + 0x4, 0xC);
 
-		replace_function(ff7_externals.sub_6F54A2, sub_6F54A2_jp);
+		// sub_6F54A2 hook lives in the shared (ff7_japanese_edition || multibyte_font) block above.
```
julianxhokaxhiu: Remove this line if already done above

julianxhokaxhiu:
Aside of the changes itself, can you please add a document in `docs` explaining how someone can make use of the multibyte feature? So we can open the door to other translators willing to make use of it. Thank you

faisalkindi:
Added `docs/mods/multibyte_font.md` (linked from the docs index) — covers enabling the flag, the FA–FE escape-byte encoding, jafont sheet authoring, the `multibyte_widths.bin` / `multibyte_linestep.bin` / `multibyte_iconmask.bin` formats with hot-reload behavior, the 0xDE reserved slot, and a quick-start checklist for new translations.

faisalkindi:
All four review points addressed in ad208f7:
- Flag renamed to `ff7_multibyte_font` everywhere (config key, variable, docs).
- Both `ffnx_info` debug calls removed.
- Indentation normalized in the hook block.
- Leftover comment at the old `sub_6F54A2` site removed.

docs/ff7/multibyte_font.md
julianxhokaxhiu, line 1: Can you move this file within a new `ff7` folder next to `ff8`? Thanks :)

julianxhokaxhiu:
Can you please also update the Changelog.md? In this case it is a separate feature so feel free to add it as a new line below the Japanese one. Thank you

faisalkindi:
Done — doc moved to `docs/ff7/multibyte_font.md` (new `ff7` folder next to `ff8`, linked from the index under an FF7 section), and Changelog.md updated with a new line below the Japanese text rendering entry.

julianxhokaxhiu (approved):

src/ff7/japanese_text.cpp
```diff
@@ -422,6 +425,75 @@ int charWidthData[6][256] =
...
+  static long long last_mtime = -1;
+  static DWORD last_check = 0;
+  if (!ff7_multibyte_font) return;
+  DWORD now = GetTickCount();
+  if (tried && (now - last_check) < 1000) return;
+  last_check = now;
+  char path[BASEDIR_LENGTH + 32];
+  // line step re-read every 1s tick — must NOT sit behind the widths mtime gate,
+  // or moving only the spacing slider never reaches it
+  _snprintf(path, sizeof(path), "%s/multibyte_linestep.bin", basedir);
```
julianxhokaxhiu: Sorry noticed this now, please update the path to request the file from the `data/kernel` path, use also the override layer helper code on top. Update also the doc file if you documented the paths used for this file.

Also please make sure to test it on top of Steam 2013/2026 releases that use `data/lang-en` and so on variants. These paths need to work across all the layers: 2026, 2013 and 1998.

Steam/GOG/Windows Store/eStore releases -> data/lang-*/kernel/ ( most likely you'll target `en` or `jp` )
 1998 -> data/kernel/

src/ff7/japanese_text.cpp
```diff
@@ -422,6 +425,75 @@ int charWidthData[6][256] =
...
+  FILE *lf = fopen(path, "rb");
+  if (lf)
+  {
+    unsigned char lb[2]; size_t ln = fread(lb, 1, 2, lf);
+    int q = (ln == 2) ? (lb[0] | (lb[1] << 8)) : (ln == 1 ? lb[0] * 4 : 0);  // 1-byte legacy = whole px
+    if (q >= 80 && q <= 160 && q != multibyte_field_linestep_q)
+    {
+      multibyte_field_linestep_q = q;
+      ffnx_info("ff7_multibyte_font: field line step = %d.%02d px\n", q / 4, (q % 4) * 25);
```
julianxhokaxhiu: This one is also most likely noise, feel free to remove it

src/ff7/japanese_text.cpp
```diff
@@ -422,6 +425,75 @@ int charWidthData[6][256] =
...
+  tried = true;
+  FILE *f = fopen(path, "rb");
+  if (!f) { if (first) ffnx_info("ff7_multibyte_font: no %s, using built-in widths\n", path); return; }
+  unsigned char buf[6 * 256];
+  if (fread(buf, 1, sizeof(buf), f) == sizeof(buf))
+  {
+    for (int i = 0; i < 6; i++)
+      for (int j = 0; j < 256; j++)
+        charWidthData[i][j] = buf[i * 256 + j];
+    ffnx_info("ff7_multibyte_font: widths %s\n", first ? "loaded" : "RELOADED (live)");
```
julianxhokaxhiu: This one is going to spam the logs, feel free to remove it

src/ff7/japanese_text.cpp
```diff
@@ -422,6 +425,75 @@ int charWidthData[6][256] =
...
+    unsigned char lb[2]; size_t ln = fread(lb, 1, 2, lf);
+    int q = (ln == 2) ? (lb[0] | (lb[1] << 8)) : (ln == 1 ? lb[0] * 4 : 0);  // 1-byte legacy = whole px
+    if (q >= 80 && q <= 160 && q != multibyte_field_linestep_q)
+    {
+      multibyte_field_linestep_q = q;
+      ffnx_info("ff7_multibyte_font: field line step = %d.%02d px\n", q / 4, (q % 4) * 25);
+    }
+    fclose(lf);
+  }
+  _snprintf(path, sizeof(path), "%s/multibyte_widths.bin", basedir);
```
julianxhokaxhiu: Same story as the other file

src/ff7/japanese_text.cpp
```diff
@@ -422,6 +425,75 @@ int charWidthData[6][256] =
...
+  {
+    for (int i = 0; i < 6; i++)
+      for (int j = 0; j < 256; j++)
+        charWidthData[i][j] = buf[i * 256 + j];
+    ffnx_info("ff7_multibyte_font: widths %s\n", first ? "loaded" : "RELOADED (live)");
+  }
+  else ffnx_error("ff7_multibyte_font: %s wrong size (need 1536 bytes)\n", path);
+  fclose(f);
+  if (!first) return;
+  _snprintf(path, sizeof(path), "%s/multibyte_iconmask.bin", basedir);
```
julianxhokaxhiu: Also here same story

julianxhokaxhiu (changes requested):
Found some other changes required before we can merge. See above. Thank you

faisalkindi:
Addressed in d19e285:
- Tuning files (`multibyte_widths.bin` / `multibyte_linestep.bin` / `multibyte_iconmask.bin`) are now requested as `data\kernel\<file>` through `redirect_path_with_override()` — override layer first, then `data/lang-*/kernel/` on Steam/GOG/Store/2026 and plain `data/kernel/` on 1998. Doc updated with the layered paths.
- All remaining `ffnx_info` calls in the loader removed (kept the single `ffnx_error` for a wrong-size widths file, since that's a real misconfiguration signal — happy to drop it too if you prefer).

One caveat to be transparent about: my current build environment is being rebuilt after a machine wipe, so I could not runtime-test the 2013/2026 path variants for this change today — the resolution logic follows the same `redirect_path_with_override` flow the FMV/file loaders use. I'll re-verify on Steam 2013 as soon as my toolchain is restored.

zaphod77:
the heart redirect is specific to the japanese edition. someone using an english exe wouldn't need to worry about it.

This one, however, is the way forward for 7th heaven compatibility.  So it would be useful to be able to port the square JP script to this support.

I understand why arabic doesn't need a new name screen.

That said certain languages need a severely extended name screen, and many need even MORE glyph space. to do this other blank spots in the us_font need to be used that don't conflct with anything else. I would use some of the space in 0xD4 to 0xDf to access additional glyph pages.

faisalkindi:
Agreed — 0xD4–0xDF as additional page escapes is a natural extension of the same mechanism; the loader/draw split in this PR should make that a small follow-up rather than a rework. Happy to review/test such a follow-up against the Arabic build once this lands.

zaphod77:
Also note that additional glyph pages could be used for stuff like mini avatars, field item icons, button prompts, and other stuff. This is a much better solution then hiding them in specific palettes

julianxhokaxhiu:
Thanks a lot for the changes. Code wise now looks good but I'd appreciate if possible a proof that this has been tested. A screenshot of the game running in Arabic would be enough, ideally on top of 1998 and either 2013/2026/GOG/etc. releases.

After that I'll be happy to approve and merge.

zaphod77:
One of the name screen, one in-game where a party member is talking, and a menu screen should be proof enough i think. :)

faisalkindi:
Runtime test done, as requested — on the **2026 Steam re-release** (FF7 1.02 US, `ff7_en.exe`), using the CI build of the PR head (`d19e285`, driver reports `1.24.3.90`), with `ff7_multibyte_font = true` and the Arabic translation (7th Heaven IRO shipping the `jafont` atlas + tuning files).

**Field dialogue (party member talking)**

field dialogue

**Main menu**

main menu

**Save menu**

save menu

Tuning-file resolution verified via `trace_files` with the three files placed **only** in `data/lang-en/kernel/`:

```
TRACE: Redirected: data\kernel\multibyte_widths.bin -> F:\...\workingdir\data\lang-en\kernel\multibyte_widths.bin
TRACE: Redirected: data\kernel\multibyte_linestep.bin -> F:\...\workingdir\data\lang-en\kernel\multibyte_linestep.bin
TRACE: Redirected: data\kernel\multibyte_iconmask.bin -> F:\...\workingdir\data\lang-en\kernel\multibyte_iconmask.bin
```

The override layer is attempted first (misses here, as expected — nothing installed in `override/`), then the `data/lang-en/kernel/` layer hits, and all three files load from there. So the layered flow from the review works on a lang-* release, and the widths/linestep/iconmask files reach the loader.

Two honest caveats:

- **Name screen intentionally not shown** — stock name entry stays as-is in this mode; that's documented as out of scope in `docs/ff7/multibyte_font.md`.
- I no longer have 1998/2013 installs in this environment, so the `data/kernel` fallback branch remains code-reviewed but not runtime-tested here; the lang-* variant called out in review is what's verified above. The Arabic build itself previously ran for weeks on the same hooks, so confidence is high.

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thanks a lot and well done for this massive achievement :)

zaphod77:
Very nice.

This leaves Korean and Chinese. Korean requires actual composition of characters to avoid bloating the space too far, and there's simply no way i can see around the fact that Chinese requires more than two bytes.

A korean phonetic naming screen will probably fit in one screen, but only if it allows composition.

faisalkindi:
Found a real collision during in-game testing of the Arabic build — fixed in `ab30802`.

**Symptom:** one Arabic letter form (medial qaf) never rendered — invisible in field text, wrong art in buffered text. Encoding, atlas, widths all verified correct byte-for-byte, so the glyph was being lost in the draw path.

**Root cause:** byte `0xD9` is hard-wired to the heart-icon redirect in the *shared* multibyte paths, not just the JP edition:
- `field_submit_draw_text_640x480_6E706D_jp`: the "not a button prompt or heart" guard excludes `0xD9` from the glyph block, so the byte draws nothing at all.
- `common_submit_draw_char_from_buffer_6F564E_jp`: `case 0xD9` always goes to the `menu_win_d` heart sprite.

So any multibyte translation that places a glyph on cell `0xD9` silently loses it — exactly the collision you predicted, @zaphod77.

**Fix (this commit):** the heart redirect now only applies when `ff7_multibyte_font` is off; in multibyte mode `0xD9` is an ordinary `jafont_1` cell in both paths (this matches your note that the redirect is JP-edition-specific). JP-edition behavior is unchanged. The doc is also corrected — it said the heart byte was `0xDE`, it's `0xD9`.

**Verified:** with the Arabic charmap's medial-qaf relocated to a free cell, the letter renders correctly in-game on this build (field + menu), and no other glyph regressed. Screenshot evidence in my earlier comment above is from the same runtime.

zaphod77:
I did say this was probably going to happen.  

Ah well. not my fault, but glad a fix was coded.

zaphod77:
There appears to be a regression in the battle square.  I had this fixed before.

I believe this fix is to blame.

field_submit_draw_text_640x480_jp incremented a pointer instead of dereferencing it when counting newlines, so the field engine's line-row counter never moved. The window-close handshake waits on that counter, so multi-line messages could stall indefinitely.

Please adjust the code so that battle square works again.

it's the dialog that shows the number of battle points

zaphod77:
All is good again.  the new resizer routine was to blame, maybe it works right for the Arabic translation, but it causes incorrect results with the japanese script.  I made the new auto resize calling behavior conditional to it not being the JP version.

PR #832, opened 2025-09-04
Substantial overhaul of NTSC-J mode and gamma functions

ChthonVII:
## Summary

(Note: This pull request builds on/replaces #816, and replaces #824, #825, and #826, which I will close shortly.)

- NTSC-J mode now simulates both "color correction" behavior and P22 gamut of a mid-90's Sony Trinitron PVM. (Formerly NTSC-J mode did a conversion from the NTSC-J specification, which was inaccurate because real televisions didn't conform to the spec.)
- LUTs computation reflects years of improvements to gamutthingy. Slightly more advanced gamut compression algorithm, VPRC; new Spiral CARISMA preprocessing; new source gamut "pruning" adjusts boundaries of source gamut to match the actual output range of color correction circuit (which is not entirely coextensive with the P22 phosphor gamut).
- BT1886 Appendix 1 EOTF function now used for NTSC-J movies (and 1998 PC intro movies).
- BT1886 Appendix 1 EOTF function now used for *everything* in NTSC-J mode (except movies with metadata that say otherwise, and Steam FF8 movies that are assumed to be BT709).
- BGFX's linearization now disabled. All linearization is done in shaders.
- In NTSC-J mode, linear RGB operations (e.g. advanced lighting) are done in "CRT linear RGB" (rather than linear sRGB), so they are fully correct.
- Somewhat surprisingly, lighting code *does* handle color values >1.0 with no major changes required. (What I long feared would be an insurmountable roadblock turned out to be no problem at all.)
- LUT code revamped again to just use texelFetch(), thus conclusively resolving all potential sampling problems forever.
- Removed redundant dithering, and moved dithering to gamma space so dither step size is proportional to quantization step size. (Also changed dither step size 255 -> 256.)
- Revised some matrices to use gamutthingy's output, with consistent number of decimal places.
- Update docs to reflect the above.
- Remove stubs for unused per-asset gamut override concept.
- Add some comments to shader code.
- Clean up some whitespace/indentation issues along the way.

### Motivation

The ultimate goal for NTSC-J mode is to closely approximates the "original intended colors" that developers saw when working on FF7/8, and that Japanese consumers originally saw playing them on a Playstation 1 hooked up to a Japanese CRT television.

[...]

PR #950, opened 2026-07-14
Common: Fix memory leak regression introduced in #832

### LINKED ISSUE #949: [ FF7 ] memleak while playing videos since last commit
The last commit seems to have introduced a memleak while playing videos. After a few seconds, the memory is full and the screen gets green:

Image

I think, it may have to do with the "av_image_alloc" in line 1135 of movie.cpp. If I add "av_freep(&sws_frame->data[0]);" in line 1147, the problem disappears.

ChthonVII:
Fix #949 as discussed therein.


julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you :)

PR #925, opened 2026-04-04
FF7: Improve native support for japanese text rendering

zaphod77:
## Summary

This is an attempt to get japanese text working fully in FF7.  This is based on the work of Cosmos, with many fixes to make stuff work better.  He did a lot of the hard work.  Included is a basic parser of the new button prompts, which allows modders to replace the button icons by modding btl_win, and full field text opcode support, including all color effects used by the Japanese script.  Field dialogs are pretty much perfect now.

There are a few issues, which I could use help with.

1) While all field dialog text is big, some menu text is small, and some is big, and this is not a L vs H difference.  What determines if a text is supposed to be small or big is what type of text it is, which i can't check inside of common_submit_draw_char_from_buffer_6F564E_jp
2) The naming menu screen needs to be rewritten to allow changing between kanji pages.
3) the prompts are still hardcoded, because i do not know which memory locations to peek in to read the key/button mappings.
4) There are some alignment issues, which are partially caused by number 1 above, and partially manual hacks done in the dotEmu driver which i dunno how to detect and apply yet.

### Motivation

To let japanese people upscale their graphics too while using the original JP script.

### ACKs

- [x] I have updated the [Changelog.md](https://github.com/julianxhokaxhiu/FFNx/blob/master/Changelog.md) file
- [x] I did test my code on FF7
- [ ] I did test my code on FF8

This only modifies japanese_text.cpp (and i think from Cosmo's pull request.  I have made it a draft, and placed comments liberally through my code so other people can understand what i'm doing and improve it.

PR #951, opened 2026-07-15
FF7: Restore the 3-mode name-entry screen for the japanese edition

### LINKED ISSUE #39: [ FF7 ] Japanese support

pphouse2789:
## Summary

Restores the 3-mode (hiragana / katakana / eisuu) name-entry screen for the japanese edition — step 2 of the plan @zaphod77 outlined in #925. A single commit with only my work, on top of current master.

Rather than re-implementing the screen, this rewires the machinery still present in ff7_ja.exe:

- full 9-row hiragana / katakana grids (verified jafont_1 glyph indices), plus a built-from-atlas eisuu (alphanumeric) grid — no eisuu master exists in the US data
- a proper 7-entry JP menu (hiragana / katakana / eisuu / space / delete / confirm / default), with the action jump table rewired so the first three items switch modes, and the menu column moved up so all 7 fit on screen
- the displayed grid lives in a private buffer (the static one at 0x921D70 gets its first 8 bytes clobbered before the screen appears, which shifted the initial layout)
- every name screen starts in hiragana instead of inheriting the previous character's last-used mode
- fixes the zero advance width of the XIII glyph (0xE6), which made the following message glyph overlap it

Known limitation: the ♥ glyph is left blank for now — as zaphod77 noted in #925, jafont_1 has an empty slot at hex D8 where the large heart should go, so that's a good future fix.

Credit for the JP foundation belongs to @CosmosXIII (#737) and @zaphod77 (#925).

### Screenshots

**Before** (bottom rows garbled, menu overflowing off-screen):

name_imput_ng

**After** (hiragana / katakana / eisuu):
fixed_hiragana
fixed_katakana
fixed_eisuu

### Motivation

The naming screen was the last major piece reported broken in #737 ("beside the renaming screen, the result is a perfectly functional japanese FF7"). This makes it behave like the original Japanese release.

### ACKs

- [ ] I did test my code on FF8 — N/A (FF7-specific)

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -10,7 +10,7 @@
 
 ## FF7
 
-- Core: Add native support for japanese text rendering ( https://github.com/julianxhokaxhiu/FFNx/pull/737 + https://github.com/julianxhokaxhiu/FFNx/pull/925 + https://github.com/julianxhokaxhiu/FFNx/pull/952 + https://github.com/julianxhokaxhiu/FFNx/pull/953) 
+- Core: Add native support for japanese text rendering ( https://github.com/julianxhokaxhiu/FFNx/pull/737 + https://github.com/julianxhokaxhiu/FFNx/pull/925 + https://github.com/julianxhokaxhiu/FFNx/pull/952 + https://github.com/julianxhokaxhiu/FFNx/pull/953 + https://github.com/julianxhokaxhiu/FFNx/pull/951) 
 
 # 1.24.3
 
--- a/src/ff7/japanese_text.cpp
+++ b/src/ff7/japanese_text.cpp
@@ -13,8 +13,10 @@
 //    GNU General Public License for more details.                          //
 /****************************************************************************/
 #include "../globals.h"
-#include "../log.h" 
+#include "../log.h"
 #include "../ff7.h"
+#include "../patch.h"
+#include <string.h>
 
 void engine_load_menu_graphics_objects_6C1468_jp(int a1)
 {
@@ -325,7 +327,11 @@ int charWidthData[6][256] =
         29, 14, 8, 23, 24, 21, 24, 24, 20, 19, 25, 22, 14, 16, 22, 18,
         27, 22, 26, 21, 27, 22, 21, 24, 22, 24, 31, 24, 23, 23, 14, 22,
         28, 27, 27, 29, 30, 12, 25, 22, 11, 0, 27, 23, 23, 23, 12, 22,
-        11, 23, 23, 0, 0, 0, 32, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        11, 23, 23, 0, 0, 0, 31, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        // ^ index 0xE6 (XIII glyph): was 32, but the width field is only 5 bits
+        //   (& 0x1F), so 32 truncated to width 0 -> zero advance -> the next
+        //   character overlapped it. 31 = max width -> advance matches the
+        //   drawn glyph width (no overlap).
         0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
 
     },{ // Jap - 1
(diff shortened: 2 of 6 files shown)


src/ff7/japanese_text.cpp
```diff
@@ -2520,3 +2526,195 @@ int sub_6F54A2_jp(byte *a1) // this function appears to affect aligning stuff to
...
+{
+  __asm
+  {
+    push 2
+    call jp_name_apply_mode
+    add  esp, 4
+    mov  eax, NAME_RESUME_ADDR
+    jmp  eax
+  }
+}
```
julianxhokaxhiu, line 2724: Would it be possible to move these patches into plain C? FFNx aims at allowing everyone to underatand and hack the engine, without inheriting assembly code. Thanks a lot

Changelog.md
```diff
@@ -11,6 +11,7 @@
 ## FF7
 
 - Core: Add native support for japanese text rendering ( https://github.com/julianxhokaxhiu/FFNx/pull/737 + https://github.com/julianxhokaxhiu/FFNx/pull/925 ) 
+- Core: Restore the 3-mode (hiragana / katakana / eisuu) name-entry screen for the japanese edition
```
julianxhokaxhiu: Please remove this line and add your PR in the line above.

julianxhokaxhiu (changes requested):
Overall looks good but there are minor changes required. Comments left on the respective areas :) Thank you

src/ff7/japanese_text.cpp
pphouse2789: Done — the inline-asm trampolines are gone. The three mode slots in the action jump table now just point at the engine's own continuation (0x71914C, i.e. the action is a no-op for the engine), and the mode switch is done in plain C from a per-frame hook that mirrors the engine's dispatch condition (decision press via the input check at 0x6F53F1, menu pane focused, cursor row <= 2 — details in the comments). Re-tested in game: mode switching, typing, and space/delete/confirm/default all behave as before.

Changelog.md
pphouse2789: Done.

src/ff7/japanese_text.cpp
```diff
@@ -2520,3 +2526,192 @@ int sub_6F54A2_jp(byte *a1) // this function appears to affect aligning stuff to
...
+// The engine still plays the decision SE itself (0x74580A, right before its
+// no-op dispatch), so the feedback is unchanged.
+static uint32_t jp_name_frame_hook = 0;
+static void jp_name_frame_719C08(int a1)
+{
+  unreplace_function(jp_name_frame_hook);            // restore the original 5 bytes
+  ((void(__cdecl*)(int))0x00719C08)(a1);             // run the original frame
+  rereplace_function(jp_name_frame_hook);            // re-apply our hook
+
+  if (*(uint32_t*)0x00921ED4 == 1)                   // menu pane focused
```
julianxhokaxhiu: Please move hardcoded addresses into relative ones, see the overall FFNx codebase how we do this one :) Thank you

src/ff7/japanese_text.cpp
```diff
@@ -2520,3 +2526,192 @@ int sub_6F54A2_jp(byte *a1) // this function appears to affect aligning stuff to
...
+//   decision button pressed this frame -> input check 0x6F53F1(0x200020)
+//   right (menu) pane focused          -> [0x00921ED4] == 1
+//   cursor on one of the 3 mode rows   -> [0x00DD453C + 1*0x38] <= 2
+// The engine still plays the decision SE itself (0x74580A, right before its
+// no-op dispatch), so the feedback is unchanged.
+static uint32_t jp_name_frame_hook = 0;
+static void jp_name_frame_719C08(int a1)
+{
+  unreplace_function(jp_name_frame_hook);            // restore the original 5 bytes
+  ((void(__cdecl*)(int))0x00719C08)(a1);             // run the original frame
```
julianxhokaxhiu: Same here

src/ff7/japanese_text.cpp
julianxhokaxhiu: Thanks a lot, in the overall function I see hardcoded addresses, I would appreciate if we could move them to relative ones as we did across the whole codebase

julianxhokaxhiu (changes requested):
I still see hardcoded addresses across the changes so please move them into relative ones, like we do in the overall codebase. Thank you

zaphod77:
So turns out that when it sees 0xD8, the game grabs the heart from btl_win_d, and that's why they left it blank.  At least in the JP version. I'm now working at implementing this feature. I still think it's stupid, and they should have just put the heart in, because this mixes drop shadows.

src/ff7/japanese_text.cpp
```diff
@@ -2520,3 +2526,202 @@ int sub_6F54A2_jp(byte *a1) // this function appears to affect aligning stuff to
...
+// starts in hiragana, instead of inheriting the previous character's
+// last-used mode (e.g. Barret's screen opening in eisuu after Cloud).
+static uint32_t jp_name_init_hook = 0;
+static void jp_name_init_71894B()
+{
+  unreplace_function(jp_name_init_hook);   // restore the original 5 bytes
+  ((void(__cdecl*)())ff7_externals.menu_sub_71894B)();
+  rereplace_function(jp_name_init_hook);   // re-apply our hook
+  jp_name_apply_mode(0);                   // force hiragana for the new screen
+}
```
julianxhokaxhiu, line 2737: Why on all the functions we have so many `unreplace_function` and `rereplace_function`? Can't you just save the original pointer and call the original function when needed, so to avoid spamming memory paging executions?

src/ff7/japanese_text.cpp
pphouse2789: Done — all raw addresses are gone from the code. Everything now goes through ff7_externals, derived relatively in ff7_data.h from the existing name_menu_sub_719C08 / menu_sub_718DBE chain (get_relative_call / get_absolute_value on the instruction operands), patch sites are anchor + offset, and even the original action targets (space/delete/confirm/default) are read from the engine's own jump table at runtime instead of being hardcoded.

src/ff7/japanese_text.cpp
pphouse2789: Good point — replace_function overwrites the function's entry bytes, so the original couldn't be called directly while hooked; the un/re-replace dance was working around that. Switched both hooks to replace_call_function on their (single) call sites instead: the original functions stay untouched and are now called directly through their ff7_externals pointers, so there's no unreplace/rereplace (and no per-frame memory-protection churn) left at all. Re-tested in game: everything behaves as before.

src/ff7/japanese_text.cpp
pphouse2789: Done — same as above, everything goes through ff7_externals now.

pphouse2789:
@zaphod77 Nice find — that explains the empty slot! I'll keep 0xD8 out of the eisuu grid until your btl_win_d heart rendering lands. For reference, once that's in, showing ♥ on the name screen should just be a matter of placing 0xD8 in one of the blank cells of the grid.

zaphod77:
Yes, getting the heart to show is most likely a matter of placing 0xD8 in the correct spot to the left of the arrow pointing right.

That said, i'm not sure if my code will run in the menus. i may need to place the hook for doing the heart elsewhere.

src/ff7/japanese_text.cpp
```diff
@@ -2520,3 +2526,206 @@ int sub_6F54A2_jp(byte *a1) // this function appears to affect aligning stuff to
...
+static void jp_name_frame_719C08(int a1)
+{
+  ((void(__cdecl*)(int))ff7_externals.name_menu_sub_719C08)(a1);
+
+  if (*ff7_externals.name_menu_selected_pane_921ED4 == 1)
+  {
+    // cursor row of the right (menu) pane; the cursor state is one struct
+    // per pane, 0x38 bytes each
+    uint32_t row = *(uint32_t*)(ff7_externals.name_menu_pane_cursor_rows_DD453C + 1 * 0x38);
+    if (row <= 2 && ff7_externals.menu_input_check_6F53F1(0x200020))
```
julianxhokaxhiu: What does `0x200020` mean in this case? Is there a way we can improve it using bit flags?

src/ff7/japanese_text.cpp
```diff
@@ -2520,3 +2526,206 @@ int sub_6F54A2_jp(byte *a1) // this function appears to affect aligning stuff to
...
+  // the init call sits inside name_menu_sub_719C08 (same offset ff7_data.h
+  // uses to derive menu_sub_71894B)
+  replace_call_function(ff7_externals.name_menu_sub_719C08 + 0x2A, jp_name_init_71894B);
+
+  // mode switching for the first three menu rows (see jp_name_frame_719C08);
+  // the per-frame call sits inside name_menu_sub_6CBD32 (same offset
+  // ff7_data.h uses to derive name_menu_sub_719C08)
+  replace_call_function(ff7_externals.name_menu_sub_6CBD32 + 0x7, jp_name_frame_719C08);
+
+  uint32_t handler = ff7_externals.menu_sub_718DBE;
```
julianxhokaxhiu: Memory is expensive, especially on a 32bit process. Please use directly the handler instead of saving it in another var.

src/ff7/japanese_text.cpp
```diff
@@ -2520,3 +2526,206 @@ int sub_6F54A2_jp(byte *a1) // this function appears to affect aligning stuff to
...
+  patch_code_dword(handler + 0xB33, (DWORD)(uintptr_t)&jp_name_menu_table[0][0]);
+
+  // rewire the 7-entry menu action jump table. The original US layout is
+  //   [0]=space [1]=delete [2]=confirm [3]=default [4]=cancel
+  // and the 7-item JP menu needs
+  //   [0..2]=hiragana/katakana/eisuu [3]=space [4]=delete [5]=confirm [6]=default
+  // The mode rows become a no-op for the engine (they jump straight to the
+  // dispatch continuation); the actual mode change is done in plain C by the
+  // per-frame hook (see jp_name_frame_719C08).
+  uint32_t table = ff7_externals.name_menu_action_jump_table_719B61;
```
julianxhokaxhiu: same here

src/ff7/japanese_text.cpp
```diff
@@ -2520,3 +2526,206 @@ int sub_6F54A2_jp(byte *a1) // this function appears to affect aligning stuff to
...
+  uint32_t action_space   = get_absolute_value(table, 0 * 4);
+  uint32_t action_delete  = get_absolute_value(table, 1 * 4);
+  uint32_t action_confirm = get_absolute_value(table, 2 * 4);
+  uint32_t action_default = get_absolute_value(table, 3 * 4);
```
julianxhokaxhiu: maybe store these as `ff7_externals` instead of declaring them here?

zaphod77:
Got the heart thing working. it was actually d9, not d8, so put that in to your eisuu table just before the arrow.

src/ff7/japanese_text.cpp
pphouse2789: It's the decision (OK) button: the menu input check ANDs its mask against a 32-bit button-state word, and the engine's own dispatch in menu_sub_718DBE tests bit 0x20 in both halves of that word (it pushes exactly 0x200020), so I pass the same mask. Now expressed as a named constant built from bit flags — jp_name_decision_mask = 0x0020 | (0x0020 << 16) — with a comment explaining it.

src/ff7/japanese_text.cpp
pphouse2789: Done — both locals are gone, the externals are used directly.

src/ff7/japanese_text.cpp
pphouse2789: Done — same as above

src/ff7/japanese_text.cpp
pphouse2789: Done — the four original action targets are now ff7_externals (name_menu_action_space_71905D / delete_71906C / confirm_719076 / default_719147), derived in ff7_data.h from the jump table like the rest. Also rebased on current master now that #952 landed, and folded #951 into the shared changelog line.

src/ff7/japanese_text.cpp
```diff
@@ -2539,3 +2545,205 @@ int sub_6F54A2_jp(byte *a1) // this function appears to affect aligning stuff to
...
+  else if (mode == 2) src = jp_name_grid_eisuu;
+  else                src = jp_name_grid_hiragana;
+  memcpy(jp_name_current_grid, src, 90);
+}
+
+// Decision (OK) button bit for the engine's menu-input check
+// (menu_input_check_6F53F1 ANDs its mask against a 32-bit button-state
+// word). The engine's own dispatch in menu_sub_718DBE tests the bit in
+// both halves of that word (it pushes exactly this mask), so we do the same.
+static const uint32_t jp_name_decision_mask = 0x0020 | (0x0020 << 16);
```
julianxhokaxhiu: You can move this to a `#define` so we save some heap :) Thanks

src/ff7/japanese_text.cpp
```diff
@@ -2539,3 +2545,205 @@ int sub_6F54A2_jp(byte *a1) // this function appears to affect aligning stuff to
...
+
+// Our own 90-byte displayed-grid buffer. The engine's grid-draw (0x7191E1,
+// 0x7198A0) and glyph-picker (getGlyph case 2, 0x718BF3) originally read the
+// grid at 0x921D70; we repoint those three reads at this buffer instead. The
+// static 0x921D70 has its first 8 bytes zeroed somewhere before the screen
+// appears (shifting the initial hiragana by 8 cells), so owning the buffer
+// makes the displayed grid immune to that clobber.
+static uint8_t jp_name_current_grid[90];
+
+static int jp_name_mode = 0; // 0=hiragana 1=katakana 2=eisuu
```
julianxhokaxhiu: What's the point of this variable? it seems to not be used anywhere in your patch. It just gets assigned but not be reused anywhere.

src/ff7/japanese_text.cpp
```diff
@@ -2539,3 +2545,205 @@ int sub_6F54A2_jp(byte *a1) // this function appears to affect aligning stuff to
...
+  patch_code_dword(ff7_externals.name_menu_action_jump_table_719B61 + 5 * 4, ff7_externals.name_menu_action_confirm_719076); // けってい
+  patch_code_dword(ff7_externals.name_menu_action_jump_table_719B61 + 6 * 4, ff7_externals.name_menu_action_default_719147); // デフォルト
+
+  // repoint the three reads of the displayed grid at our own buffer:
+  //   grid draw 1:   movzx dx,[ecx+eax+grid]  (disp32 @menu_sub_718DBE+0x428)
+  //   grid draw 2:   movzx cx,[eax+edx+grid]  (disp32 @menu_sub_718DBE+0xAE7)
+  //   glyph picker:  mov   al,[edx+ecx+grid]  (disp32 @menu_sub_718B9A+0x5C)
+  patch_code_dword(ff7_externals.menu_sub_718DBE + 0x428, (DWORD)(uintptr_t)&jp_name_current_grid[0]);
+  patch_code_dword(ff7_externals.menu_sub_718DBE + 0xAE7, (DWORD)(uintptr_t)&jp_name_current_grid[0]);
+  patch_code_dword(ff7_externals.menu_sub_718B9A + 0x5C, (DWORD)(uintptr_t)&jp_name_current_grid[0]);
```
julianxhokaxhiu, line 2782: The variable `jp_name_current_grid` will be read only once, as `name_input_jp_install` is called only once during game launch. What's the point of assigning a value to it if it's just read once?

src/ff7/japanese_text.cpp
pphouse2789: Done.

src/ff7/japanese_text.cpp
pphouse2789: Good catch — leftover from an earlier iteration, removed.

src/ff7/japanese_text.cpp
pphouse2789: It's read by the engine rather than by our code: the install only patches the buffer's address into the three engine instructions, and from then on the engine reads the buffer's contents every frame while drawing the grid / picking glyphs. jp_name_apply_mode rewrites its contents on every mode switch — that's how the mode switching works. I've extended the comment to make that clearer.

pphouse2789:
@zaphod77 Done — 0xD9 is now in the eisuu grid, right before the arrow. One note from testing on top of current master (with #952 in): on the name-entry screen, the grid cell and the typed-name preview stay blank.

zaphod77:
I was afraid of this.  This only seems to fix the heart for field mode.

I will need to investigate.  regardless, d9 is supposed to be there in the grid.

Since the field code bypasses common_submit_draw_char_from_buffer_6F564E_jp, the prompt and heart code also needs to be placed there as well.

zaphod77:
i THINK i have this solved,

Please test this commit

https://github.com/julianxhokaxhiu/FFNx/compare/master...zaphod77:FFNx:master

and tell me if it works.

src/ff7/japanese_text.cpp
julianxhokaxhiu: Thanks, now is much more clearer :) Appreciated

julianxhokaxhiu:
All right, from my own side the patch looks good to go. Let me know when you're ready as well so I can have a final pass, but code wise we're now in the right track. Thanks both!

zaphod77:
Okay i've confirmed your patch will work after i apply my second heart fix. :)

zaphod77:
I managed to test this pull (https://github.com/julianxhokaxhiu/FFNx/pull/953) with your name screen. the heart shows up as expected, in white.

image

julianxhokaxhiu:
Please rebease again your PR so we can get it in a mergeable state. Once you're aligned with @zaphod77 let me know and I'll be happy to have a final review.

pphouse2789:
Rebased on current master (with #953 in) — and confirmed on my side as well: the ♥ now shows up on the name-entry screen (grid cell and typed name) and works end to end. @zaphod77 thanks for the quick fix! @julianxhokaxhiu we're aligned — ready for the final review.

julianxhokaxhiu:
Please mark the PR as ready when you're ready for review. Thanks

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thanks a lot and well done :)

PR #952, opened 2026-07-16
FF7: Add heart support for Japanese text rendering

### LINKED ISSUE #39: [ FF7 ] Japanese support

zaphod77:
## Summary

The Japanese script expects the game engine to grab the heart from btl_win_d and show it when 0xD9 is found.  With this change, this event now happens.

### Motivation

This is first scene when chatting with Jesse.  All such flirty hearts are now restored.

### ACKs

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -11,6 +11,7 @@
 ## FF7
 
 - Core: Add native support for japanese text rendering ( https://github.com/julianxhokaxhiu/FFNx/pull/737 + https://github.com/julianxhokaxhiu/FFNx/pull/925 ) 
+- Japanese Text: Add support for 0xD9 to show the heart present in btl_win_d, which is used by the script, and works in Squares release.
 
 # 1.24.3
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -10,8 +10,7 @@
 
 ## FF7
 
-- Core: Add native support for japanese text rendering ( https://github.com/julianxhokaxhiu/FFNx/pull/737 + https://github.com/julianxhokaxhiu/FFNx/pull/925 ) 
-- Japanese Text: Add support for 0xD9 to show the heart present in btl_win_d, which is used by the script, and works in Squares release.
+- Core: Add native support for japanese text rendering ( https://github.com/julianxhokaxhiu/FFNx/pull/737 + https://github.com/julianxhokaxhiu/FFNx/pull/925 + https://github.com/julianxhokaxhiu/FFNx/pull/952) 
 
 # 1.24.3
 
--- a/src/ff7/japanese_text.cpp
+++ b/src/ff7/japanese_text.cpp
@@ -609,7 +609,7 @@ __int16 field_submit_draw_text_640x480_6E706D_jp(
             break;
           }
         default:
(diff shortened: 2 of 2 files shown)


zaphod77:
image
Example of hearts in FFNX.

Changelog.md
```diff
@@ -11,6 +11,7 @@
 ## FF7
 
 - Core: Add native support for japanese text rendering ( https://github.com/julianxhokaxhiu/FFNx/pull/737 + https://github.com/julianxhokaxhiu/FFNx/pull/925 ) 
+- Japanese Text: Add support for 0xD9 to show the heart present in btl_win_d, which is used by the script, and works in Squares release.
```
julianxhokaxhiu: Please remove this line and add this PR in the above

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thanks a lot :)

zaphod77:
You're welcome.  :)

I have yet to test this with the name screen, and have yet to test hearts with flash and rainbow.  Only true honey bee inn (cut content) does the latter, as far as I know, and i do not yet have the new name screen in to test.  I suspect the name screen will work when d9 is dropped into the spot, but flash and rainbow might not, because palette seven is used for the btl_win stuff.  The original palmer scene grabbed the heart from the normal font table, so coloring them worked.

I'm pondering the idea of peeking inside jafont_1, and only doing this substitution if there's a space really there, but i suspect the Japanese people don't generally care about True Honey Bee Inn, and that they see the content as cut for a reason, and thus color, rainbow, and flash potentially not working with these hearts is a non issue.

Bjburn2689:
As of the moment these hearts fail to be influenced by text color tags:
image
They should be under the effect of the multicolor tag here but are unaffected. Is there a way to fix this?

zaphod77:
Sigh. I KNEW someone would try to restore this cut content, and then complain about this....

The hearts are not from the same source.  They are fetched from btl_win, same as the controller icons.

The issue is the heart needs to be requested from .tex palette 7, and THEN recolored.  I have no idea how to do this.

Bjburn2689:
Okay I see the issue. I think it may be easier to append a heart to the font sheet than make that work. The PSX font sheet has it here, but they excluded it for some reason. I wonder if we couldn't add it back onto the texture and have ♥ point to that space rather than pull it from btl_win? 
image

zaphod77:
yes, that would have been the right way for square to have done it.

it seems they did it the other way just to mess with people trying to put the cut content back in!

But as the font sheet is square's property we can't include a modified version with ffnx. instead, we would need a program that we could run that pulls both files out of the lgp, and builds a replacement and drops it into direct or something.

Bjburn2689:
My ideal here would be to keep the current method by default but allow for using a modified version of the texture provided by a mod to do it the way I suggested above as a toggle a mod could set.

julianxhokaxhiu:
I would appreciate if instead of just another flag in the toml, we could find a way to do that natively driven via flevel opcodes or whatever.

Also @Bjburn2689 please open an issue to track this case, let's not spam a finished PR. Thank you :)

PR #953, opened 2026-07-16
FF7: Enable heart in all modes for Japanese text rendering

### LINKED ISSUE #39: [ FF7 ] Japanese support

zaphod77:
## Summary

My previous code only worked in field dialogues.  This code should work everywhere else. tested with the new name screen.

### Motivation

Allow Japanese people to put hearts in their character names with modded FF7.

### ACKs


zaphod77:
image
Look. heart on name screen, that goes into the character name. :)

julianxhokaxhiu:
Can you please update the Changelog.md as well?

zaphod77:
I would still like to have this conditionally active if a heart hasn't been placed in a modded jafont_1, any idea of how i might be able to detect this?  Actually doing it through the font will allow the heart to be colored with palette changes, which this probably does not.

julianxhokaxhiu:
Haven't gone down that path to be honest, but from the FFNx point of view you can just work on data you have, either from the game or from the files you're loading. Or game edition if they come with some predefined data onboard...any idea is welcome. So is this PR safe enough to be merged?

zaphod77:
I believe it is. i was able to test with the upcoming name screen, and it worked fine, as shown by my screenshot.

AS it stands, in japanese no one can use that empty character for anything after this patch. :)

after this and the naming screen then we have the last 10% that takes 90% of the time. working out how to resize and properly position every bit of text such that it aligns properly.    Checking color and z-value is not sufficient to determine what's supposed to be done with the text.  It seems what determines the scaling factor is the source of the text, which cannot be checked during the function, because it's already been copied to a buffer.

julianxhokaxhiu (approved):

julianxhokaxhiu:
Cheers!

PR #955, opened 2026-07-17
FF8: Allow loading more monster files up to 199

HobbitDur:
## Summary

Allow to load all the monster files c0mXXX.dat after 144 up to 199 (the files already there but they are blank and not used. So it works on vanilla, but allow modders to add new one)

### Motivation

To be able to add new monsters on a mod.
### ACKs

### DIFF
--- a/src/ff8/battle/monsters.cpp
+++ b/src/ff8/battle/monsters.cpp
@@ -19,6 +19,7 @@
 #include "../../ff8.h"
 #include "../../patch.h"
 #include "../../globals.h"
+#include "../../cfg.h"
 #include "../../log.h"
 
 #include <stdint.h>
@@ -62,6 +63,22 @@
 // (c0m000..c0m143 at indices 166..309, D0C000.DAT immediately at 310, 1117
 // entries total) is confirmed identical across all seven builds, so
 // FF8_BATTLE_FILES_ARRAY_LEN below is not version-dependent.
+//
+// KNOWN LIMITATION (not fixed by this patch - savemap corruption, not a
+// crash): the Scan/Libra "already scanned" bitfield SG_ENEMY_SCANNED_ONCE
+// (savemap global @ 0x1cfe964, IDA-verified) is only 5 DWORDs (20 bytes),
+// immediately followed in the packed savemap struct by SG_RENZOKUKEN_AUTO /
+// SG_RENZOKUKEN_INDICATOR / SG_ODIN_ANGEL_GILGA_FLAG / SG_TUTORIAL_INFO.
+// Both readers/writers of it - Damage_DispatchByAttackType's ATTACK_TYPE_SCAN
+// case (@0x4925a6) and hasEnemyAlreadyScanned (@0x493810) - index it as
+// SG_ENEMY_SCANNED_ONCE[(unsigned __int8)com_file_id / 32] with NO bounds
+// check, so it's only safe for com_file_id (com_id) 0..159. Every com_id this
+// patch adds (160..215, i.e. c0m144..c0m199) is already past that: Scanning
+// any of the new monsters (Scan spell / Libra / Enc-None chance-scan) writes
+// out of bounds into the Renzokuken/tutorial-flag savemap fields above,
+// silently corrupting them. Not addressed here; would need its own hook on
+// the Scan check/write (same shape as the BattleFile_CharacterLoad hook
+// below) to remap or bound com_id before it reaches SG_ENEMY_SCANNED_ONCE.
 // -------------------------------------------------------------------------
 
 // Length of the original BattleFilesArray (US 1.2), indices 0..1116.
@@ -79,6 +96,9 @@
 #define FF8_FIRST_NEW_FILE_INDEX (FF8_FIRST_NEW_COM_ID + 150) // 310
 #define FF8_LAST_NEW_FILE_INDEX  (FF8_LAST_NEW_COM_ID + 150)  // 365
 
+// Must stay static, not stack-local: ff8_battle_monsters_init() hands the exe
+// (via patch_code_dword) and ff8_externals.battle_filenames a raw pointer into
+// these arrays that's expected to remain valid for the rest of the process.
 static char *ff8_extended_battle_filenames[FF8_BATTLE_FILES_ARRAY_LEN + FF8_NEW_C0M_COUNT];
 static char ff8_extended_c0m_names[FF8_NEW_C0M_COUNT][12]; // "C0M199.DAT" + NUL = 11
 
@@ -103,13 +123,13 @@ void ff8_battle_monsters_init()
 {
 	if (!ff8_externals.battle_monster_file_load_call_site || !ff8_externals.battle_file_character_load)
 	{
-		ffnx_trace("Extra battle monsters (c0m144-c0m199): unsupported game version, skipping.\n");
+		if (trace_all) ffnx_trace("Extra battle monsters (c0m144-c0m199): unsupported game version, skipping.\n");
 		return;
 	}
 
 	if (ff8_externals.battle_filenames == nullptr)
 	{
-		ffnx_trace("Extra battle monsters: battle_filenames not resolved, skipping.\n");
+		if (trace_all) ffnx_trace("Extra battle monsters: battle_filenames not resolved, skipping.\n");
 		return;
 	}
 
--- a/src/ff8/battle/monsters.cpp
+++ b/src/ff8/battle/monsters.cpp
@@ -156,19 +156,20 @@ static_assert(FF8_SCANNED_ONCE_RELOCATE_SIZE >= 32, "must cover the full 0..255 
 #define FF8_SCENE_OUT_ENEMY_COM_VALUE_OFFSET 56
 #define FF8_SCENE_OUT_ENEMY_SLOTS            8
 
-// Both of these are handed to the exe (via patch_code_dword) and to
-// ff8_externals.battle_filenames as raw pointers, so they must stay valid for
-// the rest of the process: heap-allocated once, never freed, and static.
-static char **ff8_extended_battle_filenames = nullptr;
+// Backing storage for the new entries appended to battle_filenames; the table
+// keeps raw pointers into it for the rest of the process, so it must stay
+// static.
 static char ff8_extended_c0m_names[FF8_NEW_C0M_COUNT][12]; // "C0M199.DAT" + NUL = 11
 // Length of the original BattleFilesArray, counted at init (see below).
 static uint32_t ff8_battle_files_count = 0;
 
-// Counts the original BattleFilesArray instead of hardcoding its length. Every
-// entry is a pointer to a filename string inside the loaded module image, and
-// the dword immediately after the last entry is not (it is 0x101 on US 1.2),
-// so the run of in-image pointers ends exactly at the end of the array. Bounds
-// come from the module's own PE header, so this needs no per-version literal.
+// Counts the original BattleFilesArray. The exe never stores this table's
+// length anywhere - LoadBattleFile indexes it blindly - so there is no value
+// in memory to read back; it has to be measured. Every entry is a pointer to
+// a filename literal inside the exe's own image (bounds read from our own PE
+// header, the same way getProcessEntryPoint in utils.cpp does), and the dword
+// immediately after the last entry is not one (it is 0x101 on US 1.2), so the
+// run of in-image pointers ends exactly at the end of the array.
 static uint32_t ff8_count_battle_filenames(const char *const *filenames)
 {
 	HMODULE module = GetModuleHandleA(nullptr);
@@ -218,17 +219,25 @@ static int ff8_battle_monster_load_file(int fileIndex, void *dst)
 static bool ff8_get_battle_file_size_on_disk(const char *relative_path, uint32_t *size_out)
 {
 	char full_path[MAX_PATH];
-	WIN32_FILE_ATTRIBUTE_DATA attr;
 
 	snprintf(full_path, sizeof(full_path), "%s\\direct\\%s", ff8_externals.app_path, relative_path);
-	if (!GetFileAttributesExA(full_path, GetFileExInfoStandard, &attr))
+	FILE *file = fopen(full_path, "rb");
+	if (file == nullptr)
 	{
 		snprintf(full_path, sizeof(full_path), "%s\\%s", ff8_externals.app_path, relative_path);
-		if (!GetFileAttributesExA(full_path, GetFileExInfoStandard, &attr))
+		file = fopen(full_path, "rb");
+		if (file == nullptr)
 			return false;
 	}
 
-	*size_out = attr.nFileSizeLow;
+	fseek(file, 0, SEEK_END);
+	long file_size = ftell(file);
+	fclose(file);
+
+	if (file_size <= 0)
+		return false;
+
+	*size_out = (uint32_t)file_size;
 	return true;
 }
 
@@ -300,7 +309,7 @@ static void ff8_relocate_enemy_scanned_once()
 
 	if (from != *(uint32_t *)ff8_externals.battle_enemy_scanned_write_operand || from != expected)
 	{
-		ffnx_warning("Extra battle monsters: Scan scanned-once operands do not agree with the savemap layout (read 0x%08X, expected 0x%08X), skipping relocation - scanning c0m144-c0m199 may corrupt the savemap!\n", from, expected);
+		if (trace_all) ffnx_warning("Extra battle monsters: Scan scanned-once operands do not agree with the savemap layout (read 0x%08X, expected 0x%08X), skipping relocation - scanning c0m144-c0m199 may corrupt the savemap!\n", from, expected);
 		return;
 	}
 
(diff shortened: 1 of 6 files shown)


src/ff8/battle/monsters.cpp
```diff
@@ -0,0 +1,140 @@
+/****************************************************************************/
+//    Copyright (C) 2026 Julian Xhokaxhiu                                    //
```
julianxhokaxhiu: Feel free to add your own name as well to this list :) See other files how we do it. Thanks!

src/ff8/battle/monsters.cpp
```diff
@@ -0,0 +1,140 @@
...
+	if (!add_site)
+	{
+		ffnx_info("Extra battle monsters (c0m144-c0m199): unsupported game version, skipping.\n");
+		return;
+	}
+
+	char **orig = (char **)ff8_externals.battle_filenames;
+	if (orig == nullptr)
+	{
+		ffnx_warning("Extra battle monsters: battle_filenames not resolved, skipping.\n");
```
julianxhokaxhiu: This log line ( and the one above ) will most likely be shown to everyone, than to the ones modding. Shall we maybe do the opposite? Like show a log line only when the file is found and loaded?

src/ff8/battle/monsters.cpp
```diff
@@ -0,0 +1,140 @@
...
+	patch_code_dword(ff8_externals.battle_open_file + 0x11, (DWORD)(uintptr_t)ff8_extended_battle_filenames);
+
+	// 3) Replace "add eax, 96h" (5 bytes) with "call ff8_battle_monster_index_remap".
+	uint8_t call_patch[5];
+	call_patch[0] = 0xE8;
+	*(uint32_t *)&call_patch[1] = (uint32_t)&ff8_battle_monster_index_remap - (add_site + 5);
+	memcpy_code(add_site, call_patch, sizeof(call_patch));
+
+	ffnx_info("Extra battle monsters enabled: c0m%03d-c0m%03d usable via enemy_com_value %d-%d.\n",
+		FF8_FIRST_NEW_C0M, FF8_LAST_NEW_C0M, FF8_FIRST_NEW_COM_ID, FF8_LAST_NEW_COM_ID);
```
julianxhokaxhiu: Feel free to make this a single line, we don't need to cap at 80 chars per line like in the old days :)

src/ff8/battle/monsters.h
```diff
@@ -0,0 +1,21 @@
+/****************************************************************************/
+//    Copyright (C) 2026 Julian Xhokaxhiu                                    //
```
julianxhokaxhiu: Your name here as well please, ty

src/ff8_data.cpp
```diff
@@ -239,6 +239,29 @@ void ff8_find_externals()
...
+	else if (FF8_IT_VERSION)
+		ff8_externals.battle_monster_dat_loader_com_id_add_site = 0x506F2A;
+	else if (FF8_SP_VERSION)
+		ff8_externals.battle_monster_dat_loader_com_id_add_site = 0x506F5A;
+	else if (version == VERSION_FF8_12_JP)
+		ff8_externals.battle_monster_dat_loader_com_id_add_site = 0x50B00A;
+	else if (version == VERSION_FF8_12_JP_NV)
+		ff8_externals.battle_monster_dat_loader_com_id_add_site = 0x50B21A;
+	else
+		ff8_externals.battle_monster_dat_loader_com_id_add_site = 0;
```
julianxhokaxhiu: Can you please use relative addresses instead of hardcoding them?

julianxhokaxhiu (changes requested):
Minor things to be fixed here and there but the PR looks good :) Well done, feel free to rebase as well on top of the latest main because of Changelog.

src/ff8_data.cpp
HobbitDur: How can I do that ?

julianxhokaxhiu: You need to find which sub eventually uses it and then get the absolute value of the address. We have plenty of examples in the codebase :)

src/ff8/battle/monsters.cpp
HobbitDur: Good point — flipped it: the two skip-path logs are now `ffnx_trace` (hidden by default), and the "enabled" success line is the only one that stays at `info`. So it only shows up for the people actually using it.

src/ff8/battle/monsters.cpp
HobbitDur: Done, merged into a single line.

src/ff8_data.cpp
HobbitDur: Traced it: battle_monster_dat_loader is never called directly, but its caller BattleTask_DispatchComEntityLoad embeds it as a plain `mov eax, offset battle_monster_dat_loader` — so `get_absolute_value(dispatcher, 0x4B)` resolves it, and the `add eax, 96h` site sits at a fixed +0x23A from there. Verified identical across all 7 exes (EN/FR/DE/IT/SP/JP/JP_NV) in IDA.

BattleTask_DispatchComEntityLoad itself has no static xref anywhere in the exe either (only reached via function-pointer task dispatch), so its address is still a per-version literal — same fallback as sub_54A0D0 above — but that's now the only one, down from 7. Pushed in 50fd281.

src/ff8/battle/monsters.cpp
```diff
@@ -0,0 +1,141 @@
+/****************************************************************************/
+//    Copyright (C) 2026 Julian Xhokaxhiu, HobbitDur                        //
```
julianxhokaxhiu: Please use a separate line, see other files how we do it

src/ff8/battle/monsters.h
```diff
@@ -0,0 +1,21 @@
+/****************************************************************************/
+//    Copyright (C) 2026 Julian Xhokaxhiu, HobbitDur                        //
```
julianxhokaxhiu: Same here

src/ff8_data.cpp
```diff
@@ -239,6 +239,43 @@ void ff8_find_externals()
...
+	else if (version == VERSION_FF8_12_DE || version == VERSION_FF8_12_DE_NV)
+		battle_task_dispatch_com_entity_load = 0x506C50;
+	else if (FF8_IT_VERSION)
+		battle_task_dispatch_com_entity_load = 0x506C50;
+	else if (FF8_SP_VERSION)
+		battle_task_dispatch_com_entity_load = 0x506C80;
+	else if (version == VERSION_FF8_12_JP)
+		battle_task_dispatch_com_entity_load = 0x50AD30;
+	else if (version == VERSION_FF8_12_JP_NV)
+		battle_task_dispatch_com_entity_load = 0x50AF40;
```
julianxhokaxhiu: This block needs also to be resolved using relative/absolute addresses. Additionally please use `switch/case` to make it more readable and compact cases that use the same resolution. See other examples in the codebase where same blocks are used based on `version`

src/ff8/battle/monsters.cpp
HobbitDur: Fixed, own line per author now (matching ambient.cpp/.h style).

src/ff8/battle/monsters.h
HobbitDur: Same, fixed.

src/ff8_data.cpp
HobbitDur: To be discussed.

src/ff8/battle/monsters.cpp
```diff
@@ -0,0 +1,142 @@
+/****************************************************************************/
+//    Copyright (C) 2026 Julian Xhokaxhiu                                    //
```
julianxhokaxhiu:
```suggestion
//    Copyright (C) 2026 Julian Xhokaxhiu                                   //
```

src/ff8/battle/monsters.h
```diff
@@ -0,0 +1,22 @@
+/****************************************************************************/
+//    Copyright (C) 2026 Julian Xhokaxhiu                                    //
```
julianxhokaxhiu:
```suggestion
//    Copyright (C) 2026 Julian Xhokaxhiu                                   //
```

src/ff8_data.cpp
```diff
@@ -239,6 +239,58 @@ void ff8_find_externals()
...
+	case VERSION_FF8_12_SP_NV:
+		battle_task_dispatch_com_entity_load = 0x506C80;
+		break;
+	case VERSION_FF8_12_JP:
+		battle_task_dispatch_com_entity_load = 0x50AD30;
+		break;
+	case VERSION_FF8_12_JP_NV:
+		battle_task_dispatch_com_entity_load = 0x50AF40;
+		break;
+	}
```
julianxhokaxhiu: Ok switch case is there but addresses are still hardcoded. Please find them using the relative logic. Thanks

HobbitDur: To be discussed as I said, come to discord. I don't find how it could be a relative logic as there is none.

julianxhokaxhiu (dismissed):

src/ff8/battle/monsters.cpp
```diff
@@ -0,0 +1,147 @@
...
+		jb  normal
+		cmp eax, FF8_LAST_NEW_COM_ID    // 215
+		ja  normal
+		add eax, FF8_NEW_C0M_INDEX_BIAS // 957 -> appended c0m144..c0m199
+		ret
+	normal:
+		add eax, 150                    // original c0m000..c0m143 (and unused >= 216)
+		ret
+	}
+}
```
julianxhokaxhiu, line 189: Please translate this into pure C code using FFNx internals to patch memory. It will make the code more readable for whomever comes next. Thank you

HobbitDur: I am not sure it's a good idea, will do, but for some part assembly allow to fully understand the patch, and sometimes C version are not sure to be done correctly...

src/ff8/battle/monsters.cpp
```diff
@@ -0,0 +1,141 @@
...
+{
+	if (fileIndex >= FF8_FIRST_NEW_FILE_INDEX && fileIndex <= FF8_LAST_NEW_FILE_INDEX)
+		fileIndex = FF8_BATTLE_FILES_ARRAY_LEN + (fileIndex - FF8_FIRST_NEW_FILE_INDEX);
+
+	return ff8_battle_file_character_load(fileIndex, dst);
+}
+
+void ff8_battle_monsters_init()
+{
+	uint32_t call_site = ff8_externals.battle_monster_file_load_call_site;
```
julianxhokaxhiu: Please consider using directly `ff8_externals.battle_monster_file_load_call_site` instead of creating another variable that has the same value. In a 32bit space every bit counts. We can't afford wasting it.

src/ff8/battle/monsters.cpp
```diff
@@ -0,0 +1,141 @@
...
+	if (!call_site || !ff8_externals.battle_file_character_load)
+	{
+		ffnx_trace("Extra battle monsters (c0m144-c0m199): unsupported game version, skipping.\n");
+		return;
+	}
+
+	char **orig = (char **)ff8_externals.battle_filenames;
```
julianxhokaxhiu: Same here, use the value directly in the if comparison

HobbitDur:
Fully tested again, ready for review

src/ff8/battle/monsters.cpp
```diff
@@ -0,0 +1,376 @@
...
+		uintptr_t entry = (uintptr_t)filenames[count];
+
+		if (entry < image_start || entry >= image_end)
+			break;
+
+		++count;
+	}
+
+	return count;
+}
```
julianxhokaxhiu: What is this function supposed to do? I'm reading the description but feels quite confusing to me...if it's reading the original array size, why can't we just read the value in memory as we're already loaded in the game process? Why does it need to recalculate everything as "it is an external actor"?

src/ff8/battle/monsters.cpp
```diff
@@ -0,0 +1,376 @@
...
+	if (!GetFileAttributesExA(full_path, GetFileExInfoStandard, &attr))
+	{
+		snprintf(full_path, sizeof(full_path), "%s\\%s", ff8_externals.app_path, relative_path);
+		if (!GetFileAttributesExA(full_path, GetFileExInfoStandard, &attr))
+			return false;
+	}
+
+	*size_out = attr.nFileSizeLow;
+	return true;
+}
```
julianxhokaxhiu, line 125: I'd appreciate if you could use `ftell` like we do everywhere in the source as it will make it even more compatible later with 7th/J8, instead of advanced Windows APIs. Thank you

src/ff8/battle/monsters.cpp
```diff
@@ -0,0 +1,376 @@
...
+	// operand, since writing to one would corrupt unrelated code: the writer
+	// must reference the same address, and it must sit where the savemap
+	// layout says it does.
+	uint32_t from = *(uint32_t *)ff8_externals.battle_enemy_scanned_read_operand;
+	uint32_t expected = ff8_externals.field_vars_stack_1CFE9B8 + FF8_SG_ENEMY_SCANNED_ONCE_OFFSET;
+	uint32_t to = ff8_externals.field_vars_stack_1CFE9B8 + FF8_SCANNED_ONCE_RELOCATE_VAR;
+
+	if (from != *(uint32_t *)ff8_externals.battle_enemy_scanned_write_operand || from != expected)
+	{
+		ffnx_warning("Extra battle monsters: Scan scanned-once operands do not agree with the savemap layout (read 0x%08X, expected 0x%08X), skipping relocation - scanning c0m144-c0m199 may corrupt the savemap!\n", from, expected);
```
julianxhokaxhiu: Is this warning needed for normal players? if not, please shield it at least against `if (trace_all)` and check which other trace flags we have which could adapt to this case.

src/ff8/battle/monsters.cpp
```diff
@@ -0,0 +1,376 @@
...
+	if (ff8_externals.battle_filenames == nullptr)
+	{
+		if (trace_all) ffnx_trace("Extra battle monsters: battle_filenames not resolved, skipping.\n");
+		return;
+	}
+
+	ff8_battle_files_count = ff8_count_battle_filenames(ff8_externals.battle_filenames);
+	if (ff8_battle_files_count == 0 || ff8_battle_files_count >= FF8_BATTLE_FILES_ARRAY_MAX)
+	{
+		ffnx_warning("Extra battle monsters: unexpected battle file table length (%u), skipping.\n", ff8_battle_files_count);
```
julianxhokaxhiu: same here

src/ff8/battle/monsters.cpp
HobbitDur: It reads the number of element in the array, which is char* type ("C0M000.DAT", "D0C000.DAT",...) . So to avoid doing #define FF8_BATTLE_FILES_ARRAY_LEN 1117 I made this to be sure it reads the size and not just hardcode it.
How would you do it ?

src/ff8/battle/monsters.cpp
```diff
@@ -0,0 +1,376 @@
...
+	}
+
+	//    Repoint LoadBattleFile's array base: mov ebx, BattleFilesArray[eax*4].
+	//    battle_open_file + 0x11 is the absolute displacement of that instruction.
+	patch_code_dword(ff8_externals.battle_open_file + 0x11, (DWORD)(uintptr_t)ff8_extended_battle_filenames);
+
+	//    Also repoint FFNx's own cached base pointer so the hooks that index it
+	//    (ff8_battle_open_and_read_file in vram.cpp) resolve the appended
+	//    c0m144-199 indices instead of reading past the original array.
+	ff8_externals.battle_filenames = ff8_extended_battle_filenames;
```
julianxhokaxhiu: Is `ff8_externals.battle_filenames` memory managed by the game and free'd anywhere? If yes then we shall use this variable everywhere in the code instead of saving two pointers which use the same address.

In any case I'd prefer to see `ff8_extended_battle_filenames` being deprecated in favor of `ff8_externals.battle_filenames`

src/ff8/battle/monsters.cpp
```diff
@@ -0,0 +1,376 @@
...
+
+	// 2) Redirect the monster loader's battle-file load call to the C hook,
+	//    keeping the original target to forward to.
+	ff8_battle_load_file = (int (*)(int, void *))ff8_externals.battle_load_file_sub_508480;
```
julianxhokaxhiu: Define the type of `ff8_externals.battle_load_file_sub_508480` already to what you're casting so we avoid creating another variable in memory just to use this as a function pointer. The codebase is full of these examples for either FF7 or FF8.

src/ff8/battle/monsters.cpp
HobbitDur: Normally the ff8_relocate_enemy_scanned_once is called only if there is new monster. So vanilla player will never see it. And player with added monster will see it if there was a bug, so feel normal to show it to me.

src/ff8/battle/monsters.cpp
julianxhokaxhiu: If you know the type of the data the array holds you can do `sizeof(array)/sizeof(array[0])` assuming array is already a valid pointer to the memory that you know holds the array.

src/ff8/battle/monsters.cpp
julianxhokaxhiu: Ok if it's only once then fine by me :)

src/ff8/battle/monsters.cpp
HobbitDur: It's a char ** not really an array, it's decided at runtime.

HobbitDur: After discussion we agreed to use a hardcoded value.

src/ff8/battle/monsters.cpp
```diff
@@ -0,0 +1,239 @@
...
+// "direct" override folder first, then the plain extracted layout.
+static bool ff8_get_battle_file_size_on_disk(const char *relative_path, uint32_t *size_out)
+{
+	char full_path[MAX_PATH];
+
+	snprintf(full_path, sizeof(full_path), "%s\\direct\\%s", ff8_externals.app_path, relative_path);
+	FILE *file = fopen(full_path, "rb");
+	if (file == nullptr)
+	{
+		snprintf(full_path, sizeof(full_path), "%s\\%s", ff8_externals.app_path, relative_path);
```
julianxhokaxhiu: For the sake of simplicity all the paths we use internally use the Unix forward slash style. Feel free to check the codebase around to look at it, but please convert the paths to `/` instead of `\\`. Thank you

src/ff8/battle/monsters.cpp
```diff
@@ -0,0 +1,239 @@
...
+}
+
+// True if any encounter in scene.out uses a new monster (com_id >= 160) - the
+// Scan relocation below is only needed then. Read through sm_pc_read so file
+// overrides count, and sized from the real file so a modded scene.out works. If
+// the size can't be found, assume it's needed rather than skip a real fix.
+static bool ff8_scene_out_uses_new_monsters()
+{
+	uint32_t file_size;
+	if (!ff8_get_battle_file_size_on_disk("battle\\scene.out", &file_size))
```
julianxhokaxhiu: same here

src/ff8/battle/monsters.cpp
```diff
@@ -0,0 +1,239 @@
...
+// overrides count, and sized from the real file so a modded scene.out works. If
+// the size can't be found, assume it's needed rather than skip a real fix.
+static bool ff8_scene_out_uses_new_monsters()
+{
+	uint32_t file_size;
+	if (!ff8_get_battle_file_size_on_disk("battle/scene.out", &file_size))
+		return true;
+
+	uint8_t *scene_out = new uint8_t[file_size];
+	char scene_out_filename[] = "battle/scene.out";
```
julianxhokaxhiu: Minor optimization: define this at the beginning of the function and pass it to `ff8_get_battle_file_size_on_disk` so you define it once in the entire codebase.

julianxhokaxhiu (approved):

HobbitDur:
Thanks for the big review, help me learn for the next one :)

julianxhokaxhiu:
Well done and thanks a lot :) Congrats on your first PR!

PR #956, opened 2026-07-19
FF8: Add more magic

HobbitDur:
## Summary

This PR add more magic slot.

### Main idea
FF8 stores every spell and GF under a single byte id, and the exe distinguishes them with a bare lower-bound test — if (id >= 64) it's a GF. That works only because vanilla never ships an id above 79. The moment kernel.bin declares magic beyond that, those spells fall into GF code paths and index GF arrays out of bounds.

image

### Bigger kernel

The exe doesn't read kernel.bin's header for data sections — 31 section addresses are baked into the code at fixed offsets
So the load itself is intercepted: the game is handed an image that still looks vanilla, while the full file is kept aside.

image

Text sections are handled differently — those are resolved through the header at runtime, so their header offsets are simply pointed into the stashed full file, letting names and descriptions grow freely.

71 instructions across the exe still hold literal addresses pointing into the old 57-entry table. 

### GF condition

Four places make the `id >= 64` decision
Two are small standalone getters, so they're simply replaced wholesale with C functions

The other two sit deep inside large functions so they are managed directly in assembly like this:
image

### Fixed-size buffer
Three places size a buffer to the vanilla id range and index it unchecked:

- Menu magic re-ordering (full C rewritten)
- Junction validation (Full C rewritten)
- Drawn-once magic (Relocated in other memory unused)

### Kernel edition

The magic order need also to be taken into account in mmagic.bin. No tool yet but the file is pretty straightforward and detail can be found in the [FF8 wiki](https://hobbitdur.github.io/FF8ModdingWiki/technical-reference/menu/mmagic/#mmagicbin)
My tool [solomonRing ](https://github.com/HobbitDur/FF8UltimateEditor) already allow to edit the kernel.bin to add those new magic.

### Motivation
More magic in Cronos.

### ACKs

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -9,11 +9,14 @@
 - Core: Add support for SDL3 Gamepad API (`use_sdl_gamepad`) ( https://github.com/julianxhokaxhiu/FFNx/pull/915 )
 
 ## FF7
-
-- Core: Add native support for japanese text rendering ( https://github.com/julianxhokaxhiu/FFNx/pull/737 + https://github.com/julianxhokaxhiu/FFNx/pull/925 + https://github.com/julianxhokaxhiu/FFNx/pull/952 + https://github.com/julianxhokaxhiu/FFNx/pull/953 + https://github.com/julianxhokaxhiu/FFNx/pull/951)
+- Core: Add native support for japanese text rendering ( https://github.com/julianxhokaxhiu/FFNx/pull/737 + https://github.com/julianxhokaxhiu/FFNx/pull/925 + https://github.com/julianxhokaxhiu/FFNx/pull/952 + https://github.com/julianxhokaxhiu/FFNx/pull/953 + https://github.com/julianxhokaxhiu/FFNx/pull/951) 
+- Core: Add `ff7_multibyte_font` mode: multibyte text support for non-Japanese translations on the English executable ( https://github.com/julianxhokaxhiu/FFNx/pull/948 )
 
 ## FF8
-
+- Core: Unlock unused battle monster models c0m144-c0m199 (EN/FR/DE/IT/SP/JP), selectable via `enemy_com_value` 160-215 in scene.out
+- Direct mode: Optionally add the language prefix to the path for multi-language mods ( https://github.com/julianxhokaxhiu/FFNx/pull/898 )
+- Direct mode: Add the chunk feature for c0mXXX.dat files ( https://github.com/julianxhokaxhiu/FFNx/pull/898 )
+- Exe data: Ensure files loaded only once ( https://github.com/julianxhokaxhiu/FFNx/pull/898 )
 - External textures: Disable texture filtering in worldmap when filtering is enabled ( https://github.com/julianxhokaxhiu/FFNx/pull/954 )
 - Core: Support kernel.bin Magic sections beyond the vanilla 57 spells, unlocking ids 57-63 and 96-255 (64-95 reserved for GFs)
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -15,12 +15,14 @@
 
 ## FF8
 
-- Core: Unlock unused battle monster models c0m144-c0m199 (EN/FR/DE/IT/SP/JP), selectable via `enemy_com_value` 160-215 in scene.out
+- Core: Fix IT and DE versions crashing on launch ( https://github.com/julianxhokaxhiu/FFNx/pull/957 )
+- Core: Unlock unused battle monster models c0m144-c0m199 (EN/FR/DE/IT/SP/JP), selectable via `enemy_com_value` 160-215 in scene.out ( https://github.com/julianxhokaxhiu/FFNx/pull/955 )
 - Direct mode: Optionally add the language prefix to the path for multi-language mods ( https://github.com/julianxhokaxhiu/FFNx/pull/898 )
 - Direct mode: Add the chunk feature for c0mXXX.dat files ( https://github.com/julianxhokaxhiu/FFNx/pull/898 )
 - Exe data: Ensure files loaded only once ( https://github.com/julianxhokaxhiu/FFNx/pull/898 )
 - External textures: Disable texture filtering in worldmap when filtering is enabled ( https://github.com/julianxhokaxhiu/FFNx/pull/954 )
 - Core: Support kernel.bin Magic sections beyond the vanilla 57 spells, unlocking ids 57-63 and 96-255 (64-95 reserved for GFs) (https://github.com/julianxhokaxhiu/FFNx/pull/956)
+- Widescreen: Fix text dialogues in battles when using 16:9 ( https://github.com/julianxhokaxhiu/FFNx/pull/960 )
 
 # 1.24.3
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -10,11 +10,12 @@
 
 ## FF7
 
-- Core: Add native support for japanese text rendering ( https://github.com/julianxhokaxhiu/FFNx/pull/737 + https://github.com/julianxhokaxhiu/FFNx/pull/925 + https://github.com/julianxhokaxhiu/FFNx/pull/952 + https://github.com/julianxhokaxhiu/FFNx/pull/953 + https://github.com/julianxhokaxhiu/FFNx/pull/951) 
+- Core: Add native support for japanese text rendering ( https://github.com/julianxhokaxhiu/FFNx/pull/737 + https://github.com/julianxhokaxhiu/FFNx/pull/925 + https://github.com/julianxhokaxhiu/FFNx/pull/952 + https://github.com/julianxhokaxhiu/FFNx/pull/953 + https://github.com/julianxhokaxhiu/FFNx/pull/951)
 
 ## FF8
 
 - External textures: Disable texture filtering in worldmap when filtering is enabled ( https://github.com/julianxhokaxhiu/FFNx/pull/954 )
+- Core: Support kernel.bin Magic sections beyond the vanilla 57 spells, unlocking ids 57-63 and 96-255 (64-95 reserved for GFs)
 
 # 1.24.3
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -21,7 +21,7 @@
 - Direct mode: Add the chunk feature for c0mXXX.dat files ( https://github.com/julianxhokaxhiu/FFNx/pull/898 )
 - Exe data: Ensure files loaded only once ( https://github.com/julianxhokaxhiu/FFNx/pull/898 )
 - External textures: Disable texture filtering in worldmap when filtering is enabled ( https://github.com/julianxhokaxhiu/FFNx/pull/954 )
-- Core: Support kernel.bin Magic sections beyond the vanilla 57 spells, unlocking ids 57-63 and 96-255 (64-95 reserved for GFs) (https://github.com/julianxhokaxhiu/FFNx/pull/956)
+- Core: Support kernel.bin Magic sections beyond the vanilla 57 spells, unlocking ids 57-63 and 96-255 (64-95 reserved for GFs) ( https://github.com/julianxhokaxhiu/FFNx/pull/956 )
 - Widescreen: Fix text dialogues in battles when using 16:9 ( https://github.com/julianxhokaxhiu/FFNx/pull/960 )
 
 # 1.24.3
--- a/Changelog.md
+++ b/Changelog.md
@@ -9,16 +9,18 @@
 - Core: Add support for SDL3 Gamepad API (`use_sdl_gamepad`) ( https://github.com/julianxhokaxhiu/FFNx/pull/915 )
 
 ## FF7
+
 - Core: Add native support for japanese text rendering ( https://github.com/julianxhokaxhiu/FFNx/pull/737 + https://github.com/julianxhokaxhiu/FFNx/pull/925 + https://github.com/julianxhokaxhiu/FFNx/pull/952 + https://github.com/julianxhokaxhiu/FFNx/pull/953 + https://github.com/julianxhokaxhiu/FFNx/pull/951) 
 - Core: Add `ff7_multibyte_font` mode: multibyte text support for non-Japanese translations on the English executable ( https://github.com/julianxhokaxhiu/FFNx/pull/948 )
 
 ## FF8
+
 - Core: Unlock unused battle monster models c0m144-c0m199 (EN/FR/DE/IT/SP/JP), selectable via `enemy_com_value` 160-215 in scene.out
 - Direct mode: Optionally add the language prefix to the path for multi-language mods ( https://github.com/julianxhokaxhiu/FFNx/pull/898 )
 - Direct mode: Add the chunk feature for c0mXXX.dat files ( https://github.com/julianxhokaxhiu/FFNx/pull/898 )
 - Exe data: Ensure files loaded only once ( https://github.com/julianxhokaxhiu/FFNx/pull/898 )
 - External textures: Disable texture filtering in worldmap when filtering is enabled ( https://github.com/julianxhokaxhiu/FFNx/pull/954 )
-- Core: Support kernel.bin Magic sections beyond the vanilla 57 spells, unlocking ids 57-63 and 96-255 (64-95 reserved for GFs)
+- Core: Support kernel.bin Magic sections beyond the vanilla 57 spells, unlocking ids 57-63 and 96-255 (64-95 reserved for GFs) (https://github.com/julianxhokaxhiu/FFNx/pull/956)
 
 # 1.24.3
 
--- a/src/ff8.h
+++ b/src/ff8.h
@@ -1623,6 +1623,7 @@ struct ff8_externals
 	uint32_t battle_open_file_wrapper;
 	uint32_t battle_open_file;
 	char **battle_filenames;
+	uint32_t battle_monster_file_load_call_site; // BattleFile_CharacterLoad call inside battle_monster_dat_loader (+0x240), hooked to unlock c0m144-199 (see ff8_data.cpp / ff8/battle/monsters.cpp).
 	uint32_t battle_load_textures_sub_500900;
 	uint32_t loc_5005A0;
 	uint32_t battle_upload_texture_to_vram;
@@ -1669,6 +1670,12 @@ struct ff8_externals
 	DWORD* battle_current_actor_talking;
 	uint32_t sub_502380;
 	uint32_t sub_50A790;
+	uint32_t sub_502670; // embedded pointer read out of sub_502380 ("push offset ...")
+	uint32_t battle_entity_task_dispatch_sub_507080; // maps a battle entity id to its task function; called from sub_502670
+	uint32_t battle_monster_dat_loader; // sub_507120, the entity task for monsters; embedded pointer read out of the dispatcher
+	uint32_t battle_load_file_sub_508480; // generic "load battle file by index" wrapper, the monster loader's file-load call target (see monsters.cpp)
+	uint32_t battle_enemy_scanned_read_operand;  // disp32 of the ATTACK_TYPE_SCAN "already scanned?" test (see monsters.cpp)
+	uint32_t battle_enemy_scanned_write_operand; // disp32 of the "mark this com_id scanned" or (see monsters.cpp)
 	uint32_t sub_50A9A0;
 	uint32_t battle_read_effect_sub_50AF20;
 	DWORD* func_off_battle_effects_C81774;
--- a/src/ff8.h
+++ b/src/ff8.h
@@ -1721,12 +1721,16 @@ struct ff8_externals
 	uint32_t field_vars_stack_1CFE9B8;
 
 	// AddMoreMagic (src/ff8/kernel_magic.cpp): extended kernel.bin magic
-	// section support. Resolved in ff8_data.cpp - the data addresses
-	// relatively from nearby externals (no version branching), the code
-	// addresses from a per-version table; see that file for how each was
-	// derived. magic_sg_drawn_once_ext is NOT stored here - it is computed
-	// from field_vars_stack_1CFE9B8 (already resolved above) + 753, the
+	// section support. Everything is resolved relatively in ff8_data.cpp;
+	// see that file for how each address was derived.
+	// magic_sg_drawn_once_ext is NOT stored here - it is computed from
+	// field_vars_stack_1CFE9B8 (already resolved above) + 753, the
 	// field-script variable slot it relocates to.
+	uint32_t init_data_at_start_sub_470440;
+	uint32_t read_kernel_files_sub_47D2A0;
+	uint32_t set_all_monster_info_sub_48BA10;
+	uint32_t manage_monster_spell_visibility_sub_48C7A0;
+	uint32_t linked_menu_magic_sub_4F02F0;
 	uint32_t magic_k_magic;               // buffer + 540 (K_MAGIC data label)
 	uint32_t magic_load_file_to_buf;      // int LoadFileToBuffer(const char*, char*)
 	uint32_t magic_kernel_read_call;      // call LoadFileToBuffer(name, KERNEL_HEADER)
--- a/src/ff8.h
+++ b/src/ff8.h
@@ -1719,6 +1719,29 @@ struct ff8_externals
 	uint32_t fps_limiter;
 	double *time_volume_change_related_1A78BE0;
 	uint32_t field_vars_stack_1CFE9B8;
+
+	// AddMoreMagic (src/ff8/kernel_magic.cpp): extended kernel.bin magic
+	// section support. Resolved per-version in ff8_data.cpp; see that file
+	// for how each address was derived. magic_sg_drawn_once_ext is NOT
+	// stored here - it is computed from field_vars_stack_1CFE9B8 (already
+	// resolved above) + 753, the field-script variable slot it relocates to.
+	uint32_t magic_k_magic;               // buffer + 540 (K_MAGIC data label)
+	uint32_t magic_load_file_to_buf;      // int LoadFileToBuffer(const char*, char*)
+	uint32_t magic_kernel_read_call;      // call LoadFileToBuffer(name, KERNEL_HEADER)
+	uint32_t magic_site_name_getter;      // getMagicText:      (66)? cmp reg,40h / jcc
+	uint32_t magic_site_desc_getter;      // desc getter:       (66)? cmp reg,40h / jcc
+	uint32_t magic_site_spell_visibility; // draw-list vis:     (66)? cmp reg,40h / jcc
+	uint32_t magic_site_draw_execute;     // draw command:      66 cmp bx,40h / jcc -- includes the 0x66 prefix
+	uint32_t magic_fn_linked_stock;       // linkedStockFieldCharData(int char, int id)
+	uint32_t magic_fn_reorder_magic;      // menu_reorder_magic(int char, int preset)
+	uint32_t magic_fn_validate_magic;     // sub_4BE790(int char): per-char held-magic + junction validate
+	uint32_t magic_f_char_data;           // FF8FieldCharData[], stride 464
+	uint32_t magic_k_battle_command;      // FF8KernelBattleCommand[], stride 8
+	uint32_t magic_valid_junction;        // uint32[2] per char (8 chars): valid-junction bitfield
+	uint32_t magic_sg_chara_data;         // CharacterData[], stride 152, Magic @+16
+	uint32_t magic_magsort_buffer;        // magsortData magsortbuffer[N][7], stride 64/preset (direct array, not a pointer)
+	uint32_t magic_sg_drawn_once;         // savemap 64-bit drawn-once bitfield (ids 1-64)
+
 	uint32_t get_card_name;
 	uint32_t card_name_positions;
 	uint32_t drawpoint_messages;
--- a/src/ff8.h
+++ b/src/ff8.h
@@ -1721,10 +1721,12 @@ struct ff8_externals
 	uint32_t field_vars_stack_1CFE9B8;
 
 	// AddMoreMagic (src/ff8/kernel_magic.cpp): extended kernel.bin magic
-	// section support. Resolved per-version in ff8_data.cpp; see that file
-	// for how each address was derived. magic_sg_drawn_once_ext is NOT
-	// stored here - it is computed from field_vars_stack_1CFE9B8 (already
-	// resolved above) + 753, the field-script variable slot it relocates to.
+	// section support. Resolved in ff8_data.cpp - the data addresses
+	// relatively from nearby externals (no version branching), the code
+	// addresses from a per-version table; see that file for how each was
+	// derived. magic_sg_drawn_once_ext is NOT stored here - it is computed
+	// from field_vars_stack_1CFE9B8 (already resolved above) + 753, the
+	// field-script variable slot it relocates to.
 	uint32_t magic_k_magic;               // buffer + 540 (K_MAGIC data label)
 	uint32_t magic_load_file_to_buf;      // int LoadFileToBuffer(const char*, char*)
 	uint32_t magic_kernel_read_call;      // call LoadFileToBuffer(name, KERNEL_HEADER)
--- a/src/ff8.h
+++ b/src/ff8.h
@@ -1730,8 +1730,8 @@ struct ff8_externals
 	uint32_t magic_k_magic;               // buffer + 540 (K_MAGIC data label)
 	uint32_t magic_load_file_to_buf;      // int LoadFileToBuffer(const char*, char*)
 	uint32_t magic_kernel_read_call;      // call LoadFileToBuffer(name, KERNEL_HEADER)
-	uint32_t magic_site_name_getter;      // getMagicText:      (66)? cmp reg,40h / jcc
-	uint32_t magic_site_desc_getter;      // desc getter:       (66)? cmp reg,40h / jcc
+	uint32_t magic_fn_name_getter;        // getMagicText(int id), replaced wholesale in C
+	uint32_t magic_fn_desc_getter;        // magic description getter(int id), replaced wholesale in C
 	uint32_t magic_site_spell_visibility; // draw-list vis:     (66)? cmp reg,40h / jcc
 	uint32_t magic_site_draw_execute;     // draw command:      66 cmp bx,40h / jcc -- includes the 0x66 prefix
 	uint32_t magic_fn_linked_stock;       // linkedStockFieldCharData(int char, int id)
@@ -1743,6 +1743,7 @@ struct ff8_externals
 	uint32_t magic_sg_chara_data;         // CharacterData[], stride 152, Magic @+16
 	uint32_t magic_magsort_buffer;        // magsortData magsortbuffer[N][7], stride 64/preset (direct array, not a pointer)
 	uint32_t magic_sg_drawn_once;         // savemap 64-bit drawn-once bitfield (ids 1-64)
+	uint32_t magic_sg_gf_data;            // savemap GF records, stride 68, name at +0
 
 	uint32_t get_card_name;
 	uint32_t card_name_positions;
--- a/src/ff8.h
+++ b/src/ff8.h
@@ -1733,7 +1733,6 @@ struct ff8_externals
 	// magic_sg_drawn_once_ext is NOT stored here - it is computed from
 	// field_vars_stack_1CFE9B8 (already resolved above) + 753, the
 	// field-script variable slot it relocates to.
-	uint32_t init_data_at_start_sub_470440;
 	uint32_t read_kernel_files_sub_47D2A0;
 	uint32_t set_all_monster_info_sub_48BA10;
 	uint32_t manage_monster_spell_visibility_sub_48C7A0;
--- a/src/ff8/kernel_magic.cpp
+++ b/src/ff8/kernel_magic.cpp
@@ -26,89 +26,33 @@
 #include <string.h>
 
 // -------------------------------------------------------------------------
-// AddMoreMagic - extended kernel.bin magic section (EN/FR/DE/SP/IT/JP/JP_NV
-// retail 1.2 exes)
+// AddMoreMagic - lets kernel.bin hold more than the vanilla 57 magic spells.
 //
-// The exe ignores the kernel.bin header for data access: 31 data-section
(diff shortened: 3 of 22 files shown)


src/ff8/kernel_magic.cpp
```diff
@@ -0,0 +1,626 @@
...
+static uint8_t *ff8_trampoline_alloc(uint32_t size)
+{
+	static uint8_t *page = nullptr;
+	static uint32_t used = 0;
+	if (!page)
+		page = (uint8_t *)VirtualAlloc(nullptr, 4096, MEM_COMMIT | MEM_RESERVE, PAGE_EXECUTE_READWRITE);
+	uint8_t *stub = page + used;
+	used += size;
+	return stub;
+}
```
julianxhokaxhiu, line 147: We have already helpers doing exactly this, please remove this code and use existing memory replacement helpers.

src/ff8/kernel_magic.cpp
```diff
@@ -0,0 +1,626 @@
...
+	*(uint32_t *)p = (uint32_t)&ff8_is_gf_id - (uint32_t)(p + 4);
+	p += 4;
+	*p++ = 0x84; *p++ = 0xC0;                                 // test al,al
+	*p++ = 0x58 + reg;                                        // pop reg
+	*p++ = 0x0F; *p++ = 0x85;                                 // jnz rel32
+	*(uint32_t *)p = gf_target - (uint32_t)(p + 4);
+	p += 4;
+	*p++ = 0xC3;                                              // ret
+	return stub;
+}
```
julianxhokaxhiu, line 199: Please rewrite this in a more readable C way

src/ff8/kernel_magic.cpp
```diff
@@ -0,0 +1,626 @@
...
+
+	// call stub (5 bytes) + NOP fill for the remainder of the region.
+	uint8_t code[16];
+	code[0] = 0xE8;
+	*(uint32_t *)&code[1] = (uint32_t)stub - (site + 5);
+	for (uint32_t i = 5; i < patch_size; ++i) code[i] = 0x90;
+	memcpy_code(site, code, patch_size);
+
+	return true;
+}
```
julianxhokaxhiu, line 251: Also this sub can be rewritten better using our internal memory helpers

src/ff8/kernel_magic.cpp
```diff
@@ -0,0 +1,626 @@
...
+				}
+			}
+
+			patch_code_dword(addr, (DWORD)(to + (value - from)));
+			++rewritten;
+			addr += 3; // skip the rewritten dword
+		}
+	}
+
+	ffnx_trace("AddMoreMagic: %s: relocated %u displacement(s), skipped %u branch false-positive(s).\n",
```
julianxhokaxhiu: Consider using any `ffnx_` helper only IF needed and check if it is spamming logs when playing normally. If really needed, guard them using `if (trace_all)` at least.

src/ff8_data.cpp
```diff
@@ -978,6 +978,133 @@ void ff8_find_externals()
...
+		ff8_externals.magic_fn_reorder_magic = (version == VERSION_FF8_12_JP) ? 0x4F3FB0u : 0x4F41C0u;
+		ff8_externals.magic_fn_validate_magic = (version == VERSION_FF8_12_JP) ? 0x4C2B20u : 0x4C2D30u;
+		ff8_externals.magic_f_char_data = 0x2204F98u;
+		ff8_externals.magic_k_battle_command = 0x21F9EC4u;
+		ff8_externals.magic_valid_junction = 0x227D0ECu;
+		ff8_externals.magic_sg_chara_data = 0x2204080u;
+		ff8_externals.magic_magsort_buffer = 0x2231AF4u;
+		ff8_externals.magic_sg_drawn_once = 0x22048F4u;
+	}
+
```
julianxhokaxhiu, line 1050: This entire block has the same problem of the other PR, please use relative address resolution. And if needed, prefer `switch/case` syntax instead of `if/else`.

Changelog.md
```diff
@@ -9,7 +9,6 @@
 - Core: Add support for SDL3 Gamepad API (`use_sdl_gamepad`) ( https://github.com/julianxhokaxhiu/FFNx/pull/915 )
 
 ## FF7
-
```
myst6re: No need to remove this line break

Changelog.md
```diff
@@ -19,6 +18,7 @@
 - Direct mode: Add the chunk feature for c0mXXX.dat files ( https://github.com/julianxhokaxhiu/FFNx/pull/898 )
 - Exe data: Ensure files loaded only once ( https://github.com/julianxhokaxhiu/FFNx/pull/898 )
 - External textures: Disable texture filtering in worldmap when filtering is enabled ( https://github.com/julianxhokaxhiu/FFNx/pull/954 )
+- Core: Support kernel.bin Magic sections beyond the vanilla 57 spells, unlocking ids 57-63 and 96-255 (64-95 reserved for GFs)
```
myst6re: Add the url of the Pull request at the end of this line

src/ff8_data.cpp
```diff
@@ -989,6 +989,34 @@ void ff8_find_externals()
 
 	ff8_externals.field_vars_stack_1CFE9B8 = get_absolute_value(ff8_externals.opcode_pshm_w, 0x1E);
 
+	ff8_externals.init_data_at_start_sub_470440 = get_absolute_value(ff8_externals.credits_main_loop, 0xD2);
```
myst6re: Already declared line 850
```cpp
	ff8_externals.sub_470440 = get_absolute_value(ff8_externals.credits_main_loop, 0xD2);
```

Please check the warnings when launching the game

src/ff8_data.cpp
```diff
@@ -989,6 +989,34 @@ void ff8_find_externals()
...
+	ff8_externals.read_kernel_files_sub_47D2A0 = get_relative_call(get_relative_call(ff8_externals.init_data_at_start_sub_470440, 0x22), 0);
+	ff8_externals.magic_kernel_read_call = ff8_externals.read_kernel_files_sub_47D2A0 + 0x96; // call LoadFileToBuffer(name, KERNEL_HEADER)
+
+	ff8_externals.set_all_monster_info_sub_48BA10 = get_relative_call(ff8_externals.sub_47CCB0, 0x996);
+	ff8_externals.manage_monster_spell_visibility_sub_48C7A0 = get_relative_call(ff8_externals.set_all_monster_info_sub_48BA10, 0x1A7);
+	ff8_externals.magic_site_spell_visibility = ff8_externals.manage_monster_spell_visibility_sub_48C7A0 + 0x43; // cmp eax,40h / jge
+	ff8_externals.magic_fn_linked_stock = ff8_externals.manage_monster_spell_visibility_sub_48C7A0 + 0x340;
+	ff8_externals.magic_site_draw_execute = ff8_externals.manage_monster_spell_visibility_sub_48C7A0 + (FF8_US_VERSION ? 0xD9B : (JP_VERSION ? 0xDB6 : 0xD9C)); // 66 cmp bx,40h / jnb, inside computeCommandAction
+
+	ff8_externals.magic_fn_name_getter = get_relative_call(ff8_externals.battle_sub_48D200, FF8_US_VERSION ? 0x3C3 : (JP_VERSION ? 0x3DE : 0x3C4));
```
myst6re: I get crashes on the french version here

src/ff8_data.cpp
```diff
@@ -989,6 +989,34 @@ void ff8_find_externals()
...
+	ff8_externals.magic_fn_desc_getter = ff8_externals.magic_fn_name_getter + 0x50;
+
+	ff8_externals.magic_fn_validate_magic = get_relative_call(uint32_t(ff8_externals.menu_callbacks[1].func), 0xE5);
+	ff8_externals.linked_menu_magic_sub_4F02F0 = get_absolute_value(uint32_t(ff8_externals.menu_callbacks[3].func), 0x8);
+	ff8_externals.magic_fn_reorder_magic = get_relative_call(ff8_externals.linked_menu_magic_sub_4F02F0, 0x47BC);
+	ff8_externals.magic_load_file_to_buf = get_relative_call(ff8_externals.battle_open_file, 0x8E);
```
myst6re: I think that's ff8_externals.sm_pc_read

src/ff8/kernel_magic.cpp
```diff
@@ -0,0 +1,596 @@
...
+	uint8_t push_id_reg;   // 50+reg
+	uint8_t call_opcode;   // E8
+	int32_t call_offset;   // rel32 -> ff8_is_gf_id
+	uint8_t test_al_al[2]; // 84 C0
+	uint8_t pop_id_reg;    // 58+reg
+	uint8_t jnz_opcode[2]; // 0F 85
+	int32_t jnz_offset;    // rel32 -> the site's original GF branch target
+	uint8_t ret_opcode;    // C3
+};
+#pragma pack(pop)
```
julianxhokaxhiu, line 166: Please move this into `ff8.h`

src/ff8/kernel_magic.cpp
```diff
@@ -0,0 +1,596 @@
...
+{
+	const uint8_t *p = (const uint8_t *)site;
+	uint32_t pre = (p[0] == 0x66) ? 1 : 0;   // optional operand-size prefix
+
+	// cmp reg,imm8: (66) 83 /7 ib  (modrm F8=eax/ax, FB=ebx/bx)
+	// Just guarding if anything is unexpected in the byte we will replace
+	if (p[pre] != 0x83 || p[pre + 2] != 0x40)
+	{
+		ffnx_warning("AddMoreMagic: unexpected bytes at %s site 0x%X (%02X %02X %02X %02X), skipping patch!\n",
+			what, site, p[0], p[1], p[2], p[3]);
```
julianxhokaxhiu, line 214: Please write this as one line, no need to cut at 80 char width. Also, how many times will this be printed for normal users?

src/ff8/kernel_magic.cpp
```diff
@@ -0,0 +1,596 @@
...
+		return false;
+	}
+
+	// NOP the whole compare-and-branch, then turn its first 5 bytes into the
+	// call, so whatever follows the site stays exactly where it was.
+	memset_code(site, 0x90, patch_size);
+	replace_call_function(site, (void *)stub);
+
+	return true;
+}
```
julianxhokaxhiu, line 251: It feels to me this function is a mix of `patch_*` methods and `replace_call_function` methods. Let's not reinvent the wheel and use what is already provided, thank you

src/ff8/kernel_magic.cpp
```diff
@@ -0,0 +1,596 @@
...
+// The exe's own .text bounds, from its PE header (adapts to any build).
+static void get_code_section_bounds(uint32_t *start, uint32_t *end)
+{
+	HMODULE base = GetModuleHandleA(nullptr);
+	PIMAGE_DOS_HEADER dos = (PIMAGE_DOS_HEADER)base;
+	PIMAGE_NT_HEADERS nt = (PIMAGE_NT_HEADERS)((BYTE *)base + dos->e_lfanew);
+
+	*start = (uint32_t)base + nt->OptionalHeader.BaseOfCode;
+	*end = *start + nt->OptionalHeader.SizeOfCode;
+}
```
julianxhokaxhiu, line 266: I see your AI loves a lot to go around problems but we have already the address located when FFNx starts. Please reuse addresses instead of looking it up every time and building custom methods like this one.

src/ff8/kernel_magic.cpp
```diff
@@ -0,0 +1,596 @@
...
+			}
+
+			patch_code_dword(addr, (DWORD)(to + (value - from)));
+			++rewritten;
+			addr += 3; // skip the rewritten dword
+		}
+	}
+
+	if (trace_all) ffnx_trace("AddMoreMagic: %s: relocated %u displacement(s), skipped %u branch false-positive(s).\n",
+		what, rewritten, skipped_branch);
```
julianxhokaxhiu, line 303: Also here as a single line, thanks

src/ff8/kernel_magic.cpp
```diff
@@ -0,0 +1,596 @@
...
+// When such an id exists, move the field to free savemap space (field variable
+// 753, verified unused and inside the save CRC span) big enough for all 256
+// ids. Its DRAWN_ONCE_SITE_COUNT accessors are scattered and one has no anchor,
+// so a value-scan repoints them all. A stock game never gets here.
+static void relocate_drawn_once_bitfield()
+{
+	if (ff8_magic_count <= EXTENDED_MAGIC_FIRST)
+	{
+		if (trace_all) ffnx_trace("AddMoreMagic: no magic id >= %d, drawn-once left at vanilla 0x%08X.\n",
+			EXTENDED_MAGIC_FIRST, ff8_externals.magic_sg_drawn_once);
```
julianxhokaxhiu, line 414: same here

src/ff8/kernel_magic.cpp
```diff
@@ -0,0 +1,596 @@
...
+		return;
+	}
+
+	uint32_t sg_drawn_once_ext = ff8_externals.field_vars_stack_1CFE9B8 + 753;
+
+	uint32_t patched = relocate_scan(ff8_externals.magic_sg_drawn_once, sg_drawn_once_ext,
+		ff8_externals.magic_sg_drawn_once + 1, "drawn-once bitfield");
```
julianxhokaxhiu, line 421: and here

src/ff8/kernel_magic.cpp
```diff
@@ -0,0 +1,596 @@
...
+	if (patched != DRAWN_ONCE_SITE_COUNT)
+		ffnx_warning("AddMoreMagic: expected %d drawn-once sites, found %u - some drawn-once state may not persist correctly!\n", DRAWN_ONCE_SITE_COUNT, patched);
```
julianxhokaxhiu, line 423: How many times will this be printed for normal users?

src/ff8/kernel_magic.cpp
```diff
@@ -0,0 +1,596 @@
...
+	uint32_t k_magic_end = ff8_externals.magic_k_magic + VANILLA_MAGIC_COUNT * MAGIC_ENTRY_SIZE;
+	uint32_t rewritten = relocate_scan(ff8_externals.magic_k_magic, (uint32_t)&ff8_magic_table[0][0], k_magic_end, "K_MAGIC table");
+	if (rewritten != K_MAGIC_SITE_COUNT)
+		ffnx_warning("AddMoreMagic: expected %d K_MAGIC sites, rewrote %u - some magic reads may still use the vanilla table!\n", K_MAGIC_SITE_COUNT, rewritten);
+
+	// The two deep-in-function GF checks; the rest are replaced whole below.
+	DWORD stub_protect;
+	VirtualProtect(ff8_classify_stubs, sizeof(ff8_classify_stubs), PAGE_EXECUTE_READWRITE, &stub_protect);
+	install_id_trampoline(ff8_externals.magic_site_spell_visibility, REG_EAX, "draw-list visibility");
+	install_id_trampoline(ff8_externals.magic_site_draw_execute, REG_EBX, "draw execution");
```
julianxhokaxhiu, line 492: This stuff as said before needs to use already provided methods in the codebase

src/ff8/kernel_magic.cpp
```diff
@@ -0,0 +1,596 @@
...
+	replace_function(ff8_externals.magic_fn_name_getter, (void *)ff8_get_magic_name);
+	replace_function(ff8_externals.magic_fn_desc_getter, (void *)ff8_get_magic_description);
+	replace_function(ff8_externals.magic_fn_linked_stock, (void *)ff8_linked_stock_field_char_data);
+	replace_function(ff8_externals.magic_fn_reorder_magic, (void *)ff8_menu_reorder_magic);
+	replace_function(ff8_externals.magic_fn_validate_magic, (void *)ff8_char_validate_magic);
+
+	relocate_drawn_once_bitfield();
+
+	ffnx_info("AddMoreMagic: armed with %d magic entries (ids 57-63 free below GFs; extended magic %d-%d; ids 64-95 reserved for GFs; mmagic.bin must cover %d entries / %d bytes).\n",
+		ff8_magic_count, EXTENDED_MAGIC_FIRST, ff8_magic_count - 1, ff8_magic_count, ff8_magic_count * 4);
```
julianxhokaxhiu, line 503: Also here as one line

src/ff8/kernel_magic.cpp
```diff
@@ -0,0 +1,596 @@
...
+	bool valid = size > (int)sizeof(uint32_t) * (KERNEL_SECTION_COUNT + 1) && header[0] == KERNEL_SECTION_COUNT;
+	int entries = valid ? (int)((offsets[KERNEL_MAGIC_SECTION + 1] - offsets[KERNEL_MAGIC_SECTION]) / MAGIC_ENTRY_SIZE) : VANILLA_MAGIC_COUNT;
+
+	if (!valid || entries <= VANILLA_MAGIC_COUNT || entries > MAX_MAGIC_ID)
+	{
+		// Vanilla (or unexpected) kernel.bin: behave exactly like the
+		// original call. Nothing is armed, nothing else is patched.
+		if (size > 0) memcpy(dest, ff8_kernel_stash, size);
+		if (valid && entries != VANILLA_MAGIC_COUNT)
+			ffnx_warning("AddMoreMagic: kernel.bin has %d magic entries (max %d), ignoring extension.\n", entries, MAX_MAGIC_ID);
```
julianxhokaxhiu, line 527: How many times will this be printed?

src/ff8/kernel_magic.cpp
```diff
@@ -0,0 +1,596 @@
...
+	{
+		uint32_t src = offsets[i];
+		uint32_t dst = vanilla_data_offsets[i];
+		uint32_t copy_size = vanilla_data_offsets[i + 1] - dst;
+		uint32_t src_size = offsets[i + 1] - src;
+
+		out_header[1 + i] = dst;
+
+		if (i != KERNEL_MAGIC_SECTION && src_size != copy_size)
+			ffnx_warning("AddMoreMagic: kernel.bin data section %d has non-vanilla size %u (expected %u) - only the magic section may grow; game will likely misbehave!\n", i, src_size, copy_size);
```
julianxhokaxhiu, line 549: Also here, how many times?

src/ff8/kernel_magic.cpp
```diff
@@ -0,0 +1,596 @@
...
+	// freely. Also fill dest's vanilla-sized text area with real bytes - some
+	// code reads it by hardcoded offset, and leaving it uninitialised crashed
+	// the first menu open.
+	for (int i = KERNEL_FIRST_TEXT_SEC; i < KERNEL_SECTION_COUNT; ++i)
+		out_header[1 + i] = (uint32_t)(ff8_kernel_stash + offsets[i]) - (uint32_t)dest;
+
+	uint32_t text_dest_size = VANILLA_KERNEL_SIZE - vanilla_data_offsets[KERNEL_FIRST_TEXT_SEC];
+	uint32_t text_src_size = (uint32_t)size - offsets[KERNEL_FIRST_TEXT_SEC];
+	memcpy(dest + vanilla_data_offsets[KERNEL_FIRST_TEXT_SEC], ff8_kernel_stash + offsets[KERNEL_FIRST_TEXT_SEC],
+		text_src_size < text_dest_size ? text_src_size : text_dest_size);
```
julianxhokaxhiu, line 564: Also here all in one line please

src/ff8/kernel_magic.cpp
```diff
@@ -0,0 +1,596 @@
...
+		if (trace_all) ffnx_trace("AddMoreMagic: unsupported game version, extension disabled.\n");
+		return;
+	}
+
+	// Sanity: the call we replace must be an E8 to LoadFileToBuffer.
+	const uint8_t *call_site = (const uint8_t *)ff8_externals.magic_kernel_read_call;
+	uint32_t call_target = ff8_externals.magic_kernel_read_call + 5 + *(const int32_t *)(call_site + 1);
+	if (call_site[0] != 0xE8 || call_target != ff8_externals.magic_load_file_to_buf)
+	{
+		ffnx_warning("AddMoreMagic: kernel load call site mismatch (0x%02X -> 0x%X), extension disabled.\n", call_site[0], call_target);
```
julianxhokaxhiu, line 591: How many times will be printed?

src/ff8_data.cpp
```diff
@@ -989,6 +989,65 @@ void ff8_find_externals()
...
+	}
+	ff8_externals.magic_site_draw_execute = ff8_externals.manage_monster_spell_visibility_sub_48C7A0 + draw_execute_offset;
+	ff8_externals.magic_fn_name_getter = get_relative_call(ff8_externals.battle_sub_48D200, name_getter_offset);
+	ff8_externals.magic_fn_desc_getter = ff8_externals.magic_fn_name_getter + 0x50;
+
+	ff8_externals.magic_fn_validate_magic = get_relative_call(uint32_t(ff8_externals.menu_callbacks[1].func), 0xE5);
+	ff8_externals.linked_menu_magic_sub_4F02F0 = get_absolute_value(uint32_t(ff8_externals.menu_callbacks[3].func), 0x8);
+	ff8_externals.magic_fn_reorder_magic = get_relative_call(ff8_externals.linked_menu_magic_sub_4F02F0, 0x47BC);
+
+	ff8_externals.magic_k_battle_command = uint32_t(ff8_externals.unk_1CF3E48) + 228;  // kernel.bin data section 0
```
julianxhokaxhiu, line 1041: Please convert `228` to hex

src/ff8_data.cpp
```diff
@@ -989,6 +989,65 @@ void ff8_find_externals()
...
+	ff8_externals.magic_k_magic          = uint32_t(ff8_externals.unk_1CF3E48) + 540;  // kernel.bin data section 1
```
julianxhokaxhiu, line 1042: same here

Changelog.md
```diff
@@ -14,12 +14,14 @@
...
 
 ## FF8
+
 - Core: Fix IT and DE versions crashing on launch ( https://github.com/julianxhokaxhiu/FFNx/pull/957 )
 - Core: Unlock unused battle monster models c0m144-c0m199 (EN/FR/DE/IT/SP/JP), selectable via `enemy_com_value` 160-215 in scene.out ( https://github.com/julianxhokaxhiu/FFNx/pull/955 )
 - Direct mode: Optionally add the language prefix to the path for multi-language mods ( https://github.com/julianxhokaxhiu/FFNx/pull/898 )
 - Direct mode: Add the chunk feature for c0mXXX.dat files ( https://github.com/julianxhokaxhiu/FFNx/pull/898 )
 - Exe data: Ensure files loaded only once ( https://github.com/julianxhokaxhiu/FFNx/pull/898 )
 - External textures: Disable texture filtering in worldmap when filtering is enabled ( https://github.com/julianxhokaxhiu/FFNx/pull/954 )
+- Core: Support kernel.bin Magic sections beyond the vanilla 57 spells, unlocking ids 57-63 and 96-255 (64-95 reserved for GFs) ( https://github.com/julianxhokaxhiu/FFNx/pull/956 )
```
julianxhokaxhiu, line 24: Please move this in alphabetical order, so should be after line 19

src/ff8/kernel_magic.cpp
HobbitDur: I searched in the codebase, i didn't see any tranmpoline function. Only to rewrite full fonction but not "hack" a switch condition.

julianxhokaxhiu:
I'd appreciate if you could resume the activity on this one so we can merge it :) Thank you

HobbitDur:
I am still waiting for your remakrs

"This stuff as said before needs to use already provided methods in the codebase"
I don't find it anywhere, I don't see how to do it. @myst6re  maybe can confirm

src/ff8/kernel_magic.cpp
HobbitDur: If using a vanilla kernel, never.

src/ff8/kernel_magic.cpp
HobbitDur: Never on a normal kernel.bin

src/ff8/kernel_magic.cpp
HobbitDur: FFNx doesn't store the exe's code section bounds anywhere (the closest is getProcessEntryPoint(), which reads the same PE header for the entry point only; min_addr/max_addr in patch.cpp only exist under PATCH_COLLECT_DUPLICATES). I moved it next to getProcessEntryPoint() in utils.cpp and now read it once per arm instead of per scan.

src/ff8/kernel_magic.cpp
HobbitDur: Again, i do specific patches directly in some function, FFNx only propose full function replacement which is not possible for complex function. So yes I use replace function sometimes, and set data sometimes directly

src/ff8/kernel_magic.cpp
julianxhokaxhiu: I mean these methods https://github.com/julianxhokaxhiu/FFNx/blob/master/src/patch.cpp I don't want to have custom "trampoline injectors" only for this portion of code. All FFNx code must be coherent.

HobbitDur:
No more code patching for the GF checks

The exe has two places that read "id >= 64" as "this is a GF". That's fine on vanilla, but an extended magic id is not a GF. Both are now handled in C, with no hand-written assembly:

1. Draw list visibility — manageMonsterSpellVisibility (245 bytes, no calls) is rewritten in C and installed with replace_function, like the other magic functions this PR replaces. Only ids 64-79 count as GFs.

2. Stocking a drawn spell — this check sits in the middle of computeCommandAction, the battle command dispatcher (4599 bytes, 253 switch cases). Rewriting all of it for one comparison isn't reasonable, and rewriting only the draw case isn't either: its battle message is assembled differently in every language (different helpers, different order; JP even uses another number helper with different arguments), so a C version would mean one per build.

So the original function keeps doing the whole draw. FFNx hooks it, passes everything through untouched, and only for "stock a drawn spell with an extended id" does this:

pick a free id below 64 the caster doesn't hold and the monster doesn't offer;
point that id's entry in the FFNx-side magic table at the real spell (name, draw resistance, everything) and rename it in the monster's draw slot and the caster's inventory;
call the original function with it — the game draws, stocks, names and announces the spell exactly as it does a vanilla one, in every language;
rename everything the call wrote back to the real id, and restore the entry.

Nothing is disguised once the call returns; it lasts one synchronous call. The hook forwards every other command to the original the way ff8/vibration.cpp does (unreplace_function / rereplace_function).

All addresses are resolved in ff8_data.cpp and verified in IDA on the seven retail builds (EN, FR, DE, ES, IT, JP, JP_NV).

src/ff8/kernel_magic.cpp
```diff
@@ -0,0 +1,659 @@
...
+	}
+
+	return 0;
+}
+
+static int ff8_call_command_action(int attacker_slot, int command, int id, int variant, int target_slot, int target_mask, int linked)
+{
+	unreplace_function(ff8_compute_command_action_replaced);
+	int ret = ((compute_command_action_t)ff8_externals.battle_sub_48D200)(attacker_slot, command, id, variant, target_slot, target_mask, linked);
+	rereplace_function(ff8_compute_command_action_replaced);
```
julianxhokaxhiu: Please find a way to accomplish this without hammering memory constantly

src/ff8/kernel_magic.cpp
```diff
@@ -0,0 +1,659 @@
...
+}
+
+// ---- patch application (once, on first grown-kernel load) ---------------
+static void ff8_kernel_magic_arm()
+{
+	if (ff8_magic_armed) return;
+	ff8_magic_armed = true;
+
+	uint32_t code_start, code_end;
+	getProcessCodeSection(&code_start, &code_end);
```
julianxhokaxhiu: Why is this needed? You already have all the addresses defined in `ff8_externals` including base addresses.

src/utils.cpp
```diff
@@ -182,3 +182,15 @@ uint32_t getProcessEntryPoint()
...
+void getProcessCodeSection(uint32_t *start, uint32_t *end)
+{
+    HMODULE base = GetModuleHandleA(nullptr);
+
+    PIMAGE_DOS_HEADER dos = (PIMAGE_DOS_HEADER)base;
+    PIMAGE_NT_HEADERS nt  = (PIMAGE_NT_HEADERS)((BYTE*)base + dos->e_lfanew);
+
+    *start = (uint32_t)((BYTE*)base + nt->OptionalHeader.BaseOfCode);
+    *end = *start + nt->OptionalHeader.SizeOfCode;
+}
```
julianxhokaxhiu: This utility is not needed, we already calculate base addresses during launch, no need to do this everytime "on demand".

julianxhokaxhiu (changes requested):
Much better, looks pretty much aligned to what we usually do manually in FFNx. I love the approach also of "wrapping" the original call with extended logic on top, we use the same also for FF7 in various stages, so perfectly aligns with the rest of the code. Well done, minor changes required but we're definitely on the right track :)

src/ff8_data.cpp
```diff
@@ -1039,6 +1039,255 @@ void ff8_find_externals()
...
+		ff8_externals.battle_sub_48FE20 + damage_reads[3], // Battle_applyDamage (text part)
+		ff8_externals.battle_sub_48FE20 + damage_reads[4], // Battle_applyDamage (text part)
+		ff8_externals.battle_sub_48FE20 + damage_reads[5], // Battle_applyDamage (text part)
+		ff8_externals.battle_sub_48FE20 + damage_reads[6], // Battle_applyDamage (text part)
+		ff8_externals.battle_sub_48FE20 + damage_reads[7], // Battle_applyDamage (text part)
+		ff8_externals.battle_ai_opcode_sub_487DF0 + monster_ai_reads[0], // MonsterAI
+		ff8_externals.battle_ai_opcode_sub_487DF0 + monster_ai_reads[1], // MonsterAI
+		ff8_externals.linked_menu_magic_sub_4F02F0 + menu_magic_reads[0], // magic menu (text part)
+		ff8_externals.linked_menu_magic_sub_4F02F0 + menu_magic_reads[1], // magic menu (text part)
+	};
```
julianxhokaxhiu: Shall we move this into `ff8_data.cpp`? Not sure how often it will be read, or if it's only for this code needs. In case I'd put it maybe in the init function of `kernel_magic.cpp`, would make more sense. What do you think?

julianxhokaxhiu: I'd probably move the entire block as it's related t kernel magic anyway. Probably for the best.

julianxhokaxhiu (dismissed):

HobbitDur:
Ok now it's tested (just not with the last two commit which doesn't change a lot how it works, but still compiled). Should be good to merge.

src/ff8/kernel_magic.cpp
```diff
@@ -0,0 +1,960 @@
...
+
+// The dispatcher itself is replaced, so its first bytes are a jump back here.
+// Lift the replacement for the length of the call and put it straight back, the
+// way ff8/vibration.cpp does - the alternative, hooking call sites, has to know
+// all twelve of them and misses whichever one a future build adds.
+static int ff8_call_command_action(int attacker_slot, int command, int id, int variant, int target_slot, int target_mask, int linked)
+{
+	unreplace_function(ff8_compute_command_action_replaced);
+	int ret = ((compute_command_action_t)ff8_externals.battle_sub_48D200)(attacker_slot, command, id, variant, target_slot, target_mask, linked);
+	rereplace_function(ff8_compute_command_action_replaced);
```
julianxhokaxhiu: This got reintroduced by mistake I guess, please fix it again

src/ff8/kernel_magic.cpp
```diff
@@ -0,0 +1,960 @@
...
+	replace_function(magic_ext.manage_monster_spell_visibility_sub_48C7A0, (void *)ff8_manage_monster_spell_visibility);
+	replace_function(magic_ext.fn_name_getter, (void *)ff8_get_magic_name);
+	replace_function(magic_ext.fn_desc_getter, (void *)ff8_get_magic_description);
+	replace_function(magic_ext.fn_linked_stock, (void *)ff8_linked_stock_field_char_data);
+	replace_function(magic_ext.fn_reorder_magic, (void *)ff8_menu_reorder_magic);
+	replace_function(magic_ext.fn_validate_magic, (void *)ff8_char_validate_magic);
+
+	relocate_drawn_once_bitfield();
+
+	ffnx_info("AddMoreMagic: armed with %d magic entries (ids 57-63 free below GFs; extended magic %d-%d; ids 64-95 reserved for GFs; mmagic.bin must cover %d entries / %d bytes).\n", ff8_magic_count, EXTENDED_MAGIC_FIRST, ff8_magic_count - 1, ff8_magic_count, ff8_magic_count * 4);
```
julianxhokaxhiu: Is this info necessary to be printed? Can we put it behind the `trace_all` flag?

julianxhokaxhiu (changes requested):
Minor regressions, but looks good overall. One question: will this work with vanilla game files and did you test it across all languages? As I see this code being activated without any check ( for eg. if the kernel file is larger than usual )

src/ff8/kernel_magic.cpp
HobbitDur: We can, it was just to inform the user when using a mod with a bigger kernel that it is correctly loaded?

HobbitDur:
Yes i tested in vanilla, it works. The kernel being untouched in vanilla, it works fine. If the kernel is bigger, it will load the module. I didn't test in other languages, just checked with IDA.

src/ff8/kernel_magic.cpp
HobbitDur: Good catch, yes it was a temporary test and i didn't got back to the one expected, fixing it.

src/ff8/kernel_magic.cpp
julianxhokaxhiu: No one will ever notice it, minus the ones that tinker. Let's put it behind the flag :) Thanks

julianxhokaxhiu:
Ok I'd appreciate if we could test it at least in one non-US release, to ensure it doesn't make the game crash. If it does, let's shield the init function only to US releases for now. Thank you

HobbitDur:
Confirmed in french it worked both with vanilla and kernel modified.

julianxhokaxhiu (approved):

src/ff8.h
```diff
@@ -1775,6 +1801,7 @@ struct ff8_externals
 	uint32_t fps_limiter;
 	double *time_volume_change_related_1A78BE0;
 	uint32_t field_vars_stack_1CFE9B8;
+
```
myst6re: not needed

src/ff8/kernel_magic.cpp
```diff
@@ -0,0 +1,958 @@
...
+static char *ff8_kernel_stash = nullptr;     // full grown kernel.bin image
+static uint32_t ff8_drawn_once = 0;          // drawn-once bitfield, vanilla address or the relocated one
+
+// ---- magic-vs-GF classification -----------------------------------------
+// Only 64..79 are real GFs. The exe's "id >= 64" shortcut is what an extended
+// magic id trips over.
+static int __cdecl ff8_is_gf_id(int id)
+{
+	id &= 0xFFFF; // battle code passes a 16-bit id
+	return (id >= GF_FIRST_ID && id <= GF_LAST_ID) ? 1 : 0;
```
myst6re:
```suggestion
	return id >= GF_FIRST_ID && id <= GF_LAST_ID;
```

src/ff8/kernel_magic.cpp
```diff
@@ -0,0 +1,958 @@
...
+#define MAGIC_DESC_OFF          2
+#define K_GF_STRIDE             132
+#define K_GF_DESC_OFF           2
+// Savemap GF record stride; the GF's name sits at offset 0 of the record.
+#define GF_DATA_STRIDE          68
+
+// Vanilla data-section offsets (sections 0..31; index 31 = first text
+// section, used as the end bound of data section 30). Data section sizes are
+// language-independent, so this table is the same for every retail build.
+static const uint32_t vanilla_data_offsets[32] = {
```
myst6re: I don't understand this part. This data is already in the kernel.bin, why not just use the content of kernel.bin? What happens if someone modifies the kernel?

src/ff8/kernel_magic.cpp
```diff
@@ -0,0 +1,958 @@
...
+
+	// Rename it where the call looks it up, so a spell the caster already owns
+	// keeps stacking on its own slot.
+	int draw_slot = 0;
+	while (draw_slot < MONSTER_DRAW_SLOT_COUNT && monster[MONSTER_DRAW_SLOT_SIZE * draw_slot] != (uint8_t)spell_id)
+		++draw_slot;
+	if (draw_slot < MONSTER_DRAW_SLOT_COUNT)
+		monster[MONSTER_DRAW_SLOT_SIZE * draw_slot] = (uint8_t)stand_in;
+
+	for (int i = 0; i < BATTLE_MAGIC_SLOTS; ++i)
```
myst6re: Please use brackets for readability

src/ff8/kernel_magic.cpp
```diff
@@ -0,0 +1,958 @@
...
+		if (inventory[i].id == (uint8_t)spell_id)
+			inventory[i].id = (uint8_t)stand_in;
+
+	int ret = ff8_call_command_action(attacker_slot, command, stand_in, variant, target_slot, target_mask, linked);
+
+	// Put the real spell back everywhere the call left the stand-in.
+	for (int i = 0; i < BATTLE_MAGIC_SLOTS; ++i)
```
myst6re: same

src/ff8/kernel_magic.cpp
HobbitDur: Updated the name and explanation, and discussed directly in private.

myst6re (approved):

src/ff8.h
```diff
@@ -1094,15 +1094,41 @@ struct ff8_field_state_other {
...
 	uint8_t stat_multiplier;
 	uint8_t unk3[23];
 };
 
+// The record is indexed by stride, and the named head was split out of a single
+// unk1[370] - so both the size and where the split lands have to hold.
+static_assert(sizeof(ff8_char_computed_stats) == 464);
+static_assert(offsetof(ff8_char_computed_stats, magic) == 130);
+static_assert(offsetof(ff8_char_computed_stats, curr_hp) == 370);
+static_assert(sizeof(ff8_field_magic_slot) == 5);
```
julianxhokaxhiu: Can you please remove these asserts?

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thanks a lot, well done :)

PR #959, opened 2026-07-22
FF7: Gate 0xD9 heart redirect to the Japanese edition

faisalkindi:
## Summary

Follow-up to #948 (found via runtime testing of the Arabic translation after that PR merged).

Byte `0xD9` is hard-wired to the heart-icon redirect in the **shared** multibyte text paths, not just the Japanese edition:

- `field_submit_draw_text_640x480_6E706D_jp`: the "not a button prompt or heart" guard excludes `0xD9` from the glyph block, so the byte draws nothing.
- `common_submit_draw_char_from_buffer_6F564E_jp`: `case 0xD9` always goes to the `menu_win_d` heart sprite.

Any translation using `ff7_multibyte_font` that places a glyph on cell `0xD9` silently loses it (found with Arabic medial qaf — invisible in field text). As @zaphod77 noted on #948, the heart redirect is a JP-edition feature.

## Fix

The redirect now only applies when `ff7_multibyte_font` is off; in multibyte mode `0xD9` is an ordinary `jafont_1` glyph cell in both paths. JP-edition behavior unchanged. The doc is also corrected: the heart byte is `0xD9` (it said `0xDE`).

## Verification

- Arabic medial qaf (which encodes to `0xD9` in that charmap) renders correctly in-game with this change's semantics — confirmed on the 2026 Steam re-release (field + menu).
- JP-edition text path untouched (heart code still executes when `ff7_multibyte_font` is off).

### DIFF
--- a/docs/ff7/multibyte_font.md
+++ b/docs/ff7/multibyte_font.md
@@ -34,9 +34,9 @@ at which codes — FFNx does not impose any particular character set. Practical 
 - Control codes of the game's text format (`0xE0`+ range in kernel/field text: names, colors,
   new-line, new-page, variables...) keep their engine meaning. Don't place glyphs on bytes your
   target text sections use as control codes.
-- `0xDE` in battle window D is redirected to a heart icon by the Japanese text support
-  (a vanilla JP quirk). Treat `0xDE` as reserved if your translation may run with JP text
-  features enabled.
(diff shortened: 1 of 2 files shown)


julianxhokaxhiu (approved):

julianxhokaxhiu:
Thanks a lot :)

PR #961, opened 2026-07-22
FF8: AI able to target two slot

HobbitDur:
## Summary

FF8 AI use opcode, one we called "target" (0x04). Target defines what will be the next target for next attack. 
In vanilla you can either target all, or target one, following the below table of possible values:

| Value | Meaning | How it resolves |
|------|---------|-----------------|
| **0-10** | Specific **character** by identity | com-scan (matches character id) |
| **11-15** | Spare character ids | com-scan → no match → no target |
| **16-159** | Specific **enemy** by com id — `c0m000`-`c0m143` (value = c0m# + 16) | com-scan |
| **160-199** | `c0m144`-`c0m183` | com-scan → dead in vanilla; **live only with the c0m144-199 unlock** |
| **200** | `SELF` | the acting monster |
| **201** | `RANDOM ENEMY` | one random character (side 0-2) |
| **202** | `RANDOM ALLY` | one random monster (side 3-7) |
| **203** | `LAST ATTACKER` | whoever last hit the actor |
| **204** | `ALL ENEMIES` | all characters (0+1+2), several-flag set |
| **205** | `ALL ALLIES` | all monsters (3-7), several-flag set |
| **206** | `EVERYONE` | all 8 slots, several-flag set |
| **207** | `RANDOM NONSELF ALLY` | a random monster that isn't the actor |
| **208** | `RANDOM ENEMY EACH HIT` | re-rolls a character per hit |
| **209** | `NEW ALLY` | the slot of a monster just entering combat |
| **210-215** | `c0m194`-`c0m199` | com-scan → dead in vanilla; live with the unlock |
| **216-219** | `c0m200`-`c0m203` (never exist) | com-scan → **always dead** |
| **220** | `varA MASK` | the slot bitmask held in AI variable A |
| **221** | `varB MASK` | the slot bitmask held in AI variable B |
| **222** | `varC MASK` | the slot bitmask held in AI variable C |
| **223** | `varD MASK` | the slot bitmask held in AI variable D |
| **224** | `varE MASK` | the slot bitmask held in AI variable E |
| **225** | `varF MASK` | the slot bitmask held in AI variable F |
| **226** | `varG MASK` | the slot bitmask held in AI variable G |
| **227** | `varH MASK` | the slot bitmask held in AI variable H |

This leaves a big gap of values between 228 and 255. This PR aims to use it in order to have the following target possibilities:

| Val | Slots | Val | Slots | Val | Slots |
|----|-------|----|-------|----|-------|
| 228 | 0+1 | 235 | 1+3 | 242 | 2+6 |
| 229 | 0+2 | 236 | 1+4 | 243 | 3+4 |
| 230 | 0+3 | 237 | 1+5 | 244 | 3+5 |
| 231 | 0+4 | 238 | 1+6 | 245 | 3+6 |
| 232 | 0+5 | 239 | 2+3 | 246 | 4+5 |
| 233 | 0+6 | 240 | 2+4 | 247 | 4+6 |
| 234 | 1+2 | 241 | 2+5 | 248 | 5+6 |

Slot are the different chara/enemies, so slot 0 is the first chara till slot 2, then slot 3 is first monster till slot 6.

In order to do this, we intercept the default switch case from this:

```
switch ( op_param_target_basic )
{
  case 200u: ... target_info_mask = 1 << p_monster_slot_id; continue;   // SELF
  ...
  case 220u ... 227u: ... target_info_mask = 1 << buffer_v7; continue;  // varA-H
  default:                                    // ← values 0-199 and 228-255 land here
    v122 = 0;
    p_com_file_id = &BATTLE_SLOT_DATA[0].com_file_id;
    break;
}
```

to this:

```
default:
    // ── injected by the thunk ──
    uint32_t pm = ff8_ai_pair_mask_for(op_param_target_basic);  // 228-248 → pair mask, else 0
    if (pm) { target_info_mask = pm; continue; }                // pair: done, next opcode
    // ── original vanilla default (unchanged for 0-199, 249-255) ──
    v122 = 0;
    p_com_file_id = &BATTLE_SLOT_DATA[0].com_file_id;
    break;
```

I didn't find a way to do it in C so i went for assembly, feel free to point how I could do that

One way to implement this better would be to manage bitmask in varA to handle two bit set (atm it handle only 1). But this is way more complex to do that just adding values.
Just putting it there if anytime in the futur we need more values for other reason, this is one possible gain.

### Motivation

Gives more option to modders whnen battling, especially XATM machine new animation made in cronos.

### ACKs

### DIFF
--- a/src/ff8/battle/ai.cpp
+++ b/src/ff8/battle/ai.cpp
@@ -1,0 +1,124 @@
+/****************************************************************************/
+//    Copyright (C) 2026 Julian Xhokaxhiu                                   //
+//    Copyright (C) 2026 HobbitDur                                          //
+//                                                                          //
+//    This file is part of FFNx                                             //
+//                                                                          //
+//    FFNx is free software: you can redistribute it and/or modify          //
+//    it under the terms of the GNU General Public License as published by  //
+//    the Free Software Foundation, either version 3 of the License         //
+//                                                                          //
+//    FFNx is distributed in the hope that it will be useful,               //
+//    but WITHOUT ANY WARRANTY; without even the implied warranty of        //
+//    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         //
+//    GNU General Public License for more details.                          //
+/****************************************************************************/
+
+// Monster-AI "target" opcode (0x04): make the unused target values 228-248
+// select a fixed pair of battle slots, so one action hits two targets.
+// 228 = slots 0+1, 229 = 0+2, ... 248 = 5+6 (every pair a<b over slots 0-6).
+// FF8 has 7 battle slots: 0-2 party, 3-6 enemies (4 max) - slot 7 never exists.
+// Vanilla resolves these to "no target found", so the values are free to reuse.
+//
+// The opcode has no call boundary, so we repoint its range-check `ja` at a tiny
+// thunk: it asks ff8_ai_pair_mask_for() for the pair mask, writes it to the
+// interpreter's target-mask local and resumes the loop; other values fall
+// through to the original code. All game addresses come from ff8_externals.
+
+#include "ai.h"
+
+#include "../../ff8.h"
+#include "../../patch.h"
+#include "../../log.h"
+
+#include <stdint.h>
+
+#define FF8_AI_PAIR_FIRST_VALUE 228
+#define FF8_AI_PAIR_LAST_VALUE  248
+#define FF8_AI_SLOT_COUNT       7  // FF8 battle entity slots: 0-2 party, 3-6 enemies
+#define FF8_AI_PAIR_COUNT       (FF8_AI_PAIR_LAST_VALUE - FF8_AI_PAIR_FIRST_VALUE + 1) // 21
+#define FF8_AI_TARGET_MASK_LOCAL_OFFSET 0x78 // [esp+0x78] at the range-check ja
+
+static_assert(FF8_AI_SLOT_COUNT * (FF8_AI_SLOT_COUNT - 1) / 2 == FF8_AI_PAIR_COUNT,
+	"value range must match the number of slot pairs C(7,2)=21");
+
+// value-228 -> slot bitmask, built from the a<b pair enumeration.
+static uint32_t ff8_ai_pair_mask_table[FF8_AI_PAIR_COUNT];
+
+static void ff8_ai_build_pair_mask_table()
+{
+	int index = 0;
+	for (int a = 0; a < FF8_AI_SLOT_COUNT; ++a)
+		for (int b = a + 1; b < FF8_AI_SLOT_COUNT; ++b)
+			ff8_ai_pair_mask_table[index++] = (1u << a) | (1u << b);
+}
+
+// Pair mask for a value in 228-248, or 0 for anything else (handled as vanilla).
+extern "C" uint32_t __cdecl ff8_ai_pair_mask_for(uint32_t target_value)
+{
+	if (target_value < FF8_AI_PAIR_FIRST_VALUE || target_value > FF8_AI_PAIR_LAST_VALUE)
+		return 0;
+
+	return ff8_ai_pair_mask_table[target_value - FF8_AI_PAIR_FIRST_VALUE];
+}
+
+// Thunk continuations, set from ff8_externals so no game address is hardcoded.
+static uint32_t ff8_ai_loop_head_addr = 0;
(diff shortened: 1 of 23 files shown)


julianxhokaxhiu:
I'll have a look at the code part tomorrow, but in the meantime could you add some documentation in `docs/ff8` about this layer? The PR seems to be quite detailed and would be nice to have this knowledge somewhere in the repo for anyone willing to use it later without having to dig across PRs. Thank you :)

src/ff8/battle/ai.cpp
```diff
@@ -0,0 +1,124 @@
...
+static void ff8_ai_build_pair_mask_table()
+{
+	int index = 0;
+	for (int a = 0; a < FF8_AI_SLOT_COUNT; ++a)
+		for (int b = a + 1; b < FF8_AI_SLOT_COUNT; ++b)
+			ff8_ai_pair_mask_table[index++] = (1u << a) | (1u << b);
+}
+
+// Pair mask for a value in 228-248, or 0 for anything else (handled as vanilla).
+extern "C" uint32_t __cdecl ff8_ai_pair_mask_for(uint32_t target_value)
```
julianxhokaxhiu, line 57: Why the `extern` syntax?

src/ff8/battle/ai.cpp
```diff
@@ -0,0 +1,124 @@
...
+		test  eax, eax
+		jz    do_default
+		add   esp, 4
+		mov   dword ptr [esp + FF8_AI_TARGET_MASK_LOCAL_OFFSET], eax
+		jmp   dword ptr [ff8_ai_loop_head_addr]
+	do_default:
+		pop   edx                          // restore dl for the original scan
+		jmp   dword ptr [ff8_ai_default_scan_addr]
+	}
+}
```
julianxhokaxhiu, line 95: Please rewrite this in C

HobbitDur: I have no idea how to do it, as it's really wacky and tight to the assembly code. If you can help me show how to do it I am all for it.

julianxhokaxhiu: Ok I verified these days that if you use GPT5.6 Sol it should be able to understand the game context and rewrite this either as C++ or a byte patch. Let it know this information and it will rewrite this block better which would fit the way we do FFNx :) Thank you

HobbitDur: So you prefer a byte patch than a clear asm instruction ? Byte patch is just hell to understand what it means :o

julianxhokaxhiu: Not saying that I prefer it but when you ask the LLM to think that way magically it unlocks logics that before it thought it wouldn't be possible. Hammer it until it finds a way to do it via native C code :)

myst6re: There is a `patch_code_dword` to replace a relative address to `ff8_ai_target_pair_thunk()`. Nothing strange at all. But this is a switch/case patch. We cannot get the value of the register "dl" without rewrite everything in `ff8_externals.battle_ai_opcode_sub_487DF0`. "dl" is passed to `ff8_ai_pair_mask_for` and then the context is either restored or the result is passed to "[esp+0x78]".
Again, "dl" and "[esp+0x78]" are not stored in a variable we can get, and we are patching inside a switch/case then a loop, then another switch case.

Another patch will be highly complex. My approach in this kind of situation is either rewrite everything, or override a part of the code to make a call, passing all the needed parameters, etc. For that we need enough space in unused/replaceable portion of the code.

julianxhokaxhiu: I'd prefer to have this rewritten in C then. AI is extremely good at looking at IDA via MCP for eg. and then re-implementing it. I'm not fan of "reading registries" and patching whatever. If we want to have full and custom control we do it like we always did: rewrite the sub and add only the portion of code we're expanding.

HobbitDur: I understand your point of view, but for huge function with thousands of lines this is not really feasible or testable fully. So risky when a precise patch make sure we understand what we are adoing, which seems to be a big part of what you expect from a PR.

julianxhokaxhiu (approved):

HobbitDur:
Full tested, good to merge.

julianxhokaxhiu (approved):
LGTM, I'll wait for Myst to approve as well

myst6re (approved):

julianxhokaxhiu:
Thanks a lot and well done :)

PR #963, opened 2026-07-26
Common: change SDL3 Subsystems to prevent potential CMake build conflicts during maintenance/version updating

AL2009man:
## Summary

While SDL3 itself is ABI Stable as hell, there can be external factors that are beyond SDL's control, such as upgrading or downgrading Windows SDK versions.

I personally faced this issue from a different community patch when they upgraded their Windows SDK that broke SDL completely. While those issues have been addressed by SDL maintainers (after I reported the bug myself), it'll be a while for those issues to get rolled out into the next monthly bug fix release candidate, and FFNx is no exception. Thus: this PR is made as a simple maintenance change backported from a different project (although it'll apply to CMakeLists only).

### Motivation

To prevent any potential issues whenever the maintainer upgrades SDL versions, vcpkg, or the Windows SDK: a portion of SDL subsystems **that isn't about Game Controllers** (HIDAPI, JOYSTICK, SENSOR, HAPTICS) has been disabled. We're using SDL's Gamepad API, so it make sense for the rest of the subsystems to be disabled. 

of course, if anyone plans to upgrade FFNx's audio/windowing/gpu to leverage SDL APIs, they can always head over to `CMakeLists.txt` and remove one of the subsystems at anytime.

### ACKs

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -95,6 +95,18 @@ find_package(CMakeRC CONFIG REQUIRED)
 find_package(lfreist-hwinfo CONFIG REQUIRED)
 find_package(cryptopp CONFIG REQUIRED)
 find_package(tomlplusplus CONFIG REQUIRED)
+
+# SDL3 subsystem configurations:
+set(SDL_AUDIO        OFF CACHE BOOL "" FORCE)
+set(SDL_VIDEO        OFF CACHE BOOL "" FORCE)
+set(SDL_GPU          OFF CACHE BOOL "" FORCE)
+set(SDL_RENDER       OFF CACHE BOOL "" FORCE)
+set(SDL_CAMERA       OFF CACHE BOOL "" FORCE)
+set(SDL_POWER        OFF CACHE BOOL "" FORCE)
+set(SDL_DIALOG       OFF CACHE BOOL "" FORCE)
+set(SDL_TRAY         OFF CACHE BOOL "" FORCE)
+set(SDL_NOTIFICATION OFF CACHE BOOL "" FORCE)
+
 find_package(SDL3 CONFIG REQUIRED)
 
 set(RELEASE_NAME "FFNx")


julianxhokaxhiu (approved):

julianxhokaxhiu:
Despite the patch seems to be correct from the CMake POV, I wonder if those flags actually do anything or is just placebo on your own side. I did try to search for them in either the https://github.com/microsoft/vcpkg/tree/master/ports/sdl3 port file ( but even if they do something, they're set the wrong way if they were related to it ) as well as in the SDL3 official docu ( https://wiki.libsdl.org/SDL3/README-cmake ) and I found neither.

How did you came up with those flags?

AL2009man:
> Despite the patch seems to be correct from the CMake POV, I wonder if those flags actually do anything or is just placebo on your own side.

it should do something about it, but strangely: the build notes doesn't showcase SDL Subsystem list in my end, so i'll guess it's a vkpkg thing (unlike others projects that does it traditionally). I did try to add some sort of confirmation to determine that it's isn't placebo thing.

here's a quick example from a different github repository that doesn't use vkpkg, but I disabled the selected main submodule

image

I'll have to look into vkpkg and see what I can *try* to do. (edit: i'm still looking into it, but depending on the conclusion, this PR *might* have to be closed and will be kept as it is.)


> How did you came up with those flags?

I honestly don't remember how I exactly did it back then (was like two-three months ago), but I basically went through deepwiki and search engines inbetween VS Code's very own back when I was troubleshooting a bug (with assistance from someone who also troubleshoot the bug), but basically https://github.com/libsdl-org/SDL/blob/main/CMakeLists.txt#L248 and customized the existing CMakeList. ultimately, 

[but, I did found someone doing the same setup I eventually did](https://github.com/libsdl-org/SDL/issues/14918). 

...but it's a lot easier to enable/disable subsystems through CMake GUI

AL2009man:
After reviewing and investigating for a bit, i'm going to close this PR for the time being til there's a better way. 

I think what we currently have is fine. If there's any issue that could spawned after upgrading a slightly older Windows SDK version, just make sure it's on the latest version or modify it to avoid build errors...and file a bug report to SDL directly whenever possible (they're pretty quick on bug fixes). I wanna make sure you guys don't face the same issues I've had with Windows SDK breaking configurations!

(i am not familiar with vkpkg, more familiar with CMake's build stuffs)

PR #964, opened 2026-07-27
FF7: Fix battle square crash regression caused by #948

### LINKED ISSUE #962: [ FF7 ] Regression with battle square in JP support.
**Describe the bug**
When the dialog printing the current battle points prints, it cannot be advanced past.  I had encountered this bug while making my patch and had fixed it, but it now seems broken again.

**To Reproduce**

1. Go to Battle Square
2. Get into a series of battels.
3. Stop battling and collect your battle points
4. The game refuses to advance past the dialog showing them. 

**Expected behavior**
The game should continue.

**Additional context**
The dialog did not have a size specified, but had a variable length variable in it. I put in code to get past this, and give the variable a fixed width of 8 characters so the auto resizer could handle this dialog.  The code is still present, but the bug seems to have returned, due to some other change.

zaphod77:
… as it's required to avoid a crash after battle square.

## Summary

Without calling the resizer every frame and rewriting the values, the battle square dialog fails to expand to it's full height, and this causes the dialog to wedge.  I had to restore the old behavior from my original patch, and i do it by running it every frame if it's the japanese edition.  

### Motivation

Fixing the regression introduced by the english ver multibyte support.

### ACKs

### DIFF
--- a/src/ff7/japanese_text.cpp
+++ b/src/ff7/japanese_text.cpp
@@ -2578,7 +2578,7 @@ void field_text_box_window_opening_6317A9_jp(short WINDOW_ID)
   // stalling the open. Instead, compute the target once here while the window is still small,
   // then hold width/height fixed so the animation converges normally, matching vanilla behavior.
   // Field files already ship with correctly sized windows, so this only fixes the animation target.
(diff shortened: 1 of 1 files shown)


julianxhokaxhiu (approved):

julianxhokaxhiu:
Thanks a lot :)

PR #970, opened 2026-08-19
FF7: Fix heart icon for japanese edition again

zaphod77:
## Summary

Changes the faulty logic so that prompts and hearts only replace empty characters. It turns out i can peek at the character widths to determine if a character is actually there or not.  The means we no longer need to check if we are using the japanese script directly to make these decisions.  Sorry for the state of the commit history, but i clicked the wrong button on the website, and i can only discard commits when upstream is ahead.  Once it's ahead again i can do this again.

### Motivation

To fix silly bugs that people have spotted.

### ACKs

### DIFF
--- a/src/ff7/japanese_text.cpp
+++ b/src/ff7/japanese_text.cpp
@@ -320,7 +320,7 @@ int charWidthData[6][256] =
         30, 30, 28, 31, 30, 30, 29, 29, 30, 30, 29, 30, 31, 30, 29, 27,
         30, 29, 29, 29, 31, 30, 28, 23, 30, 30, 30, 31, 29, 31, 30, 30,
         31, 30, 30, 31, 31, 29, 21, 28, 29, 30, 30, 27, 31, 30, 30, 29,
-        29, 30, 30, 22, 22, 22, 22, 23, 22, 22, 22, 22, 22, 51, 62, 24,
+        29, 30, 30, 22, 22, 22, 22, 23, 22, 22, 22, 22, 22, 51, 62, 16,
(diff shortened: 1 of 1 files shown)


zaphod77:
Please test this fully to make sure no hearts or prompts appear where they shouldn't, and they always appear where they should.

Note that `ff7_japanese_edition` is still required, because other behavior does still depend on that flag.

src/ff7/japanese_text.cpp
```diff
@@ -702,8 +714,20 @@ __int16 field_submit_draw_text_640x480_6E706D_jp(
...
+          if (charWidthData[0][0xD9u] == 0) // if width is zero for that
+            heartAtD9 = true;                            // then that slot is a heart
+          // check if it could be a prompt.
+          if (charWidthData[curPage][*buffer_text] == 0) // msut have zero spacing
           {
+            // check range
+            if (*buffer_text > 0xF5u || *buffer_text < 0xFAu)
+              isPrompt = true;
+          }
+          if ((!possibleOpcode) || ((!isPrompt) && (*buffer_text != 0xd9u || !heartAtD9))) // check for prompts if on first page, and for heart if on first page and actual japanese
```
julianxhokaxhiu:
```suggestion
          if (!possibleOpcode || (!isPrompt && (*buffer_text != 0xd9u || !heartAtD9))) // check for prompts if on first page, and for heart if on first page and actual japanese
```

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you

PR #986, opened 2026-09-17
Core: Fix crash when logging a Windows error message

HobbitDur:
## Summary

`GetErrorMessage()` returns a `std::wstring`, but both of its call sites passed the object itself through the varargs of a printf-style log call. `isFileSigned()` additionally printed a `const char*` path with `%ls`. Both are undefined behaviour, and in practice the process dies while formatting the message.

### Why it matters

`isFileSigned()` takes that path whenever `WinVerifyTrust` does not return `ERROR_SUCCESS` — which includes a file that is simply missing or unsigned:

```c
if (status != ERROR_SUCCESS) ffnx_error("Unable to verify '%ls': %ls", dllPath, GetErrorMessage(status));
```

On the Steam release `load_library()` checks `FFNx_steam_api.dll` through this function before the hash fallback, so a missing or mismatched `FFNx_steam_api.dll` crashes the game during startup. The log ends on a bare `ERROR:` with no message, because the crash happens inside the formatting of that very line:

```
[00000000] TRACE: Detected Steam 2013 edition.
[00000000] TRACE: Metadata: Initializing manager.
[00000000] TRACE: First chance exception 0xe0434352 at 0x75F0AED4
[00000000] ERROR:
```

The intended behaviour is to report which file could not be verified, and then let the existing "Invalid %s detected" message box explain the problem. With the fix, that is what happens.

The crash handler had the same bug in its `MiniDumpWriteDump` failure path, where it fires while already handling a crash — so a failed minidump write turned into a second crash instead of a log line.

### Change

Two lines: `%s` for the narrow path, and `.c_str()` on the `std::wstring` at both sites.

### Motivation
Allow more error info

### ACKs

🤖 Generated with [Claude Code](https://claude.com/claude-code)


julianxhokaxhiu (approved):

julianxhokaxhiu:
I understand the bug and I'm fine merging it. However I don't get how converting it to a `c_str()` will fix the root cause. `%ls` is exactly meant to be used for Wide chars so the conversion will happen either by `printf` or before by us.

Are we sure this is a real bug or just hallucination from Claude?

//EDIT: Also please update the description to use the template we ask to human users.

myst6re:
> I understand the bug and I'm fine merging it. However I don't get how converting it to a `c_str()` will fix the root cause. `%ls` is exactly meant to be used for Wide chars so the conversion will happen either by `printf` or before by us.
> 
> Are we sure this is a real bug or just hallucination from Claude?
> 
> //EDIT: Also please update the description to use the template we ask to human users.

Yes %ls means wide, but I don't think printf can ever support std::string, so you'll always need to convert to a c string

julianxhokaxhiu:
Ok I noticed now the method returns a `std::wstring` which is the root cause of the problem. It's not ready for a classic `printf` even though internally is a `wchar_t`. We could potentially use its internal pointer with the existing formatting, but I guess the PR bypasses all that headache and goes for the `c_str` pathway.

Allright, good to go then. Please make sure the template is updated and I'll be happy to merge. Cheers

HobbitDur:
YEs sorry about the template, i got into the bug in the middle of something so asked claude to do the PR as it was really simple and didn't think about the template, i'll try to use it next time. I must admit I have the code "tested" in the sense that i can launch the game but didn't check again if I am missing the info again (so can't repeat how exactly the error message appear).

julianxhokaxhiu (approved):

julianxhokaxhiu:
Thank you :)

